PR #18, opened 2014-08-26
Implement FromStr for Url

wycats:
This was supported in the old `url` package. We used it to import env-var based configuration into a variety of types.


TyOverby:
Other libraries (like an argument parser library) could depend on an item having FromStr.  I think it's a nice usability feature.

PR #37, opened 2014-10-03
Modifications to use new slicing_syntax

dkhenry:


Hoverbear:
I don't think this is a problem anymore, please re-open this issue if so.

PR #38, opened 2014-10-12
Added a method to serialize a Url to a path with a query string

gtolle:
Intended for use by HTTP clients. The Hyper client currently calls serialize_path() when constructing its GET requests, but this discards the query string. I'm suggesting this new rust-url method so that Hyper can use it.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/38)


pyfisch:
Is this PR still relevant? Should I create a new PR with updated code for this function?

SimonSapin:
With #176 this will be `&url[Position::BeforePath..Postition::AfterQuery]`.

PR #70, opened 2015-01-03
Fix for upstream changes

sfackler:


src/format.rs
```diff
@@ -72,14 +72,14 @@ pub struct UrlNoFragmentFormatter<'a> {
...
+        try!(formatter.write_str(&*self.url.scheme));
+        try!(formatter.write_str(":"));
         try!(self.url.scheme_data.fmt(formatter));
         match self.url.query {
             None => (),
             Some(ref query) => {
-                try!(formatter.write(b"?"));
-                try!(formatter.write(query.as_bytes()));
+                try!(formatter.write_str("?"));
+                try!(formatter.write_str(&**query));
```
SimonSapin: Is `&**query` preferable to `query.as_slice()`? (Same for `&*self.url.scheme` v.s. `self.url.scheme.as_slice()`)

sfackler: Not sure. It's what I've been using for my own stuff. I don't feel too strongly either way, though.

SimonSapin: Ok. Please change them to `.as_slice()`, then.

sfackler:
Updated

SimonSapin:
Thanks!

PR #71, opened 2015-01-06
Update serialize

retep998:

PR #72, opened 2015-01-06
serialize traits use associated types

kstep:


kstep:
Oops, duplicate of #71. */me writing: "I will always check already existing PRs for my changes" 1000 times*

PR #76, opened 2015-01-23
update rust & fix tests

suhr:
Subj.


SimonSapin:
Thanks! I’ve published v0.2.17 with your first commit. Tests are failing, but I expect this will be resolved when [the `Result::unwrap` change is reverted](https://github.com/rust-lang/rust/pull/21558).

PR #77, opened 2015-01-24
Fix tests for nightly

rgs1:
Signed-off-by: Raul Gutierrez S <rgs@itevenworks.net>


SimonSapin:
Thanks, but I’d rather not do this as it looks like [the `Result::unwrap` change is going to be reverted](https://github.com/rust-lang/rust/pull/21558).

SimonSapin:
It was reverted. Thanks anyway!

PR #78, opened 2015-02-03
Update

retep998:


Hoverbear:
This is for the `iter()` changes regarding moves and the `FromStr` changes.

badboy:
About the `iter()` changes, today I committed this on my own fork to make it work again: https://github.com/badboy/rust-url/commit/12d7e614babf42b151375e6eba422c605a2c76f7

SimonSapin:
@hoverbear Thanks for volunteering to co-maintain! You should have access to publish on crates.io now. To do so:

* Run `git pull` if a PR was merged with the green button.
* Edit `Cargo.toml` to increment the version number. (Micro component for language upgrade)
* Run `cargo publish`. If it fails, fix things until it succeeds.
* Run `git commit -a -m v0.2.17` (with the version number copy/pasted form the output of `cargo publish`)
* Run `git tag v0.2.17` (with the version number copy/pasted form the output of `cargo publish`)
* Run `git push && git push --tags`

Hoverbear:
@SimonSapin Thanks. When is it appropriate to publish a new crate version?

SimonSapin:
@Hoverbear Whenever there is a change since the last publication that might be useful to someone, which is definitely the case for fixes like this for breaking language changes. If in doubt, publish. Publishing is cheap. I would have Travis-CI do it on every push to master that passes tests, except that deciding which component of the version number to increment is hard to automate.

(I also gave access to @seanmonstar who published for this PR (thanks!) so you don’t need to do it right now.)

PR #79, opened 2015-02-06
upstream rust: std::path moved to std::old_path

mikedilger:
Just changed all references to std::path to std::old_path to keep it compiling.
Further work will be needed to use the new std::path

PR #80, opened 2015-02-07
Revert "upstream rust: std::path moved to std::old_path"

SimonSapin:
Reverts servo/rust-url#79


Hoverbear:
@SimonSapin Why did you revert this?

SimonSapin:
It and b8c456ee02301cf492346884177752ed42d25d92 (which I had already published but forgot to push) were redundant, and I prefer the latter which uses `use std::old_path as path;` so that the rest of the code doesn’t need to be changed.

Issue #81, opened 2015-02-23
`url::form_urlencoded::serialize` is moving iterated key-value pairs


Byron:
Currently [serialize()](http://servo.github.io/rust-url/url/form_urlencoded/fn.serialize.html) is difficult to use for all users who don't want to drain vectors of string slices, but prefer keep iteration strictly read only.
Maybe, in the servo project it's uncommon to do that, but at least I am making my case and believe read-only operation should be the default.

### Example
```rust
url::form_urlencoded::serialize([("client_id", "bogus"),
                                 ("scope", "A B C"].iter().map(|&(a,b)| (a, b)));
```
To my mind, the `.map()` call should not be required.

Maybe my code is the exception though, and not the rule, but when implementing protocols one will usually have a pre-determined set of key-value pairs.

SimonSapin:
This part of the API has not gotten a lot of use or thought yet. What specific input type do you think would work better here?

Byron:
I think if the signature would be changed to ...
```Rust
// note the `&` in front of the tuple
// disclaimer: don't know if it actually compiles
pub fn serialize<'a, I>(pairs: I) -> String where I: Iterator<Item = &'a (&'a str, &'a str)> {
```
... then we should be good for my case.
While we are talking about it, it might be worth removing [`serialized_owned`](http://servo.github.io/rust-url/url/form_urlencoded/fn.serialize_owned.html) from the public API as well, just because it's unintuitive that it all of the sudden is that specific. Just making it generic might be worth doing alternatively.

SimonSapin:
This signature works for you specific use case, but is less flexible: the outer `&` means you have to have existing tuples of `&str` somewhere, to reference them.

I’m starting to thing an iterator is not so great for this. What do you think of a method you call multiple times instead?

```rust
let mut s = String::new();
{
    let mut query = url::from_urlencoded::Serializer::new(&mut s);
    query.add_pair("client_id", "bogus");
    query.add_pair("scope", "A B C");
}
```

Byron:
The builder patterns seems like 'too much' in this case. It would require the serializer to keep its own, dynamically allocated state, which is less efficient than it should be. It's my believe that an API should not enforce unnecessary copies or allocations. Ideally, it's made so that it works out of the box for 80% of the use-cases, but can be adapted with a little effort (like `.iter().map(|&(a,b)| (a, b))` in my case .. which is of course part of the 80% of the code ;)).

Here are my thoughts as code:

```Rust
[...]

PR #82, opened 2015-02-24
`form_urlencoded::serialize(...)` handles owned as well as borrowed pairs

Byron:
This is the result of a long path full of (my own) trial and error on how to allow `serialize()` not to be hardcoded to owned tuples, and increase usability.

I have added some tests which prove it now works both ways, without altering the existing API in any way (i.e. this is not a breaking change).

Even though discussed in #81, I didn't removed `serialize_owned()`, as it's possibly breaking and I would want to leave that to the superior judgement of the maintainers.

### DIFF
--- a/src/form_urlencoded.rs
+++ b/src/form_urlencoded.rs
@@ -13,6 +13,7 @@
 //! Converts between a string (such as an URL’s query string)
 //! and a sequence of (name, value) pairs.
 
+use std::borrow::Borrow;
 use std::ascii::AsciiExt;
 use encoding::EncodingOverride;
 use percent_encoding::{percent_encode_to, percent_decode, FORM_URLENCODED_ENCODE_SET};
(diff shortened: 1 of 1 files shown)


SimonSapin:
> the superior judgement of the maintainers.

These are strong words, when the maintainers often feel like this :)


Regardless, I’m a bit reluctant to add complexity to this signature, when it seems the only reason borrowed tuples are around in the first place is that [`[T; N]` doesn’t implement `IntoIterator`](http://internals.rust-lang.org/t/implementing-intoiterator-for-t-n/1648). See also https://github.com/servo/rust-url/issues/81#issuecomment-75754029

Byron:
Well, at least I suspect the maintainers have less 'no idea of what they are doing' than me :).

If until Rust beta, [T; N] should implement IntoIterator (and I am sceptic this will happen), my change indeed shouldn't be necessary.
One could wait with the merge until then.

Personally I think the complexity in the signature is just what Rust is, and should be something we all get more and more accustomed to. Also I prefer adding complexity to the implementation to keep it light and simple on the caller's end. *The needs of the many outweigh the needs of the few* :) !

#81 comment will be addressed there.

PR #83, opened 2015-02-24
Public urlutils and check type in set_scheme

valenting:
Make urlutils public and disallow changing the scheme type to an incompatible type

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,7 +147,7 @@ use encoding::EncodingOverride;
 mod encoding;
 mod host;
 mod parser;
-mod urlutils;
+pub mod urlutils;
 pub mod percent_encoding;
 pub mod form_urlencoded;
 pub mod punycode;
@@ -434,12 +434,19 @@ pub enum SchemeType {
     FileLike,
 }
 
-
 impl SchemeType {
     pub fn default_port(&self) -> Option<u16> {
         match self {
             &SchemeType::Relative(default_port) => Some(default_port),
             _ => None,
+        }
+    }
+    pub fn same_as(&self, other: SchemeType) -> bool {
+        match (self, other) {
+            (&SchemeType::NonRelative, SchemeType::NonRelative) => true,
+            (&SchemeType::Relative(_), SchemeType::Relative(_)) => true,
+            (&SchemeType::FileLike,    SchemeType::FileLike) => true,
+            _ => false
         }
     }
 }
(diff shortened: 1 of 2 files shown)


SimonSapin:
I’m not sure about disallowing the changes. In the test case below, no two of the browsers tested are interoperable but they all do change the scheme:

http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=3428
```html
<a href="mailto:a@b.net">
<script>
var a = document.body.firstChild; 
console.log(a.href);
a.protocol = "http";
console.log(a.href)
</script>
```

In Chormium:
```
log: mailto:a@b.net
log: http://a@b.net/
```

In old Opera (Presto):
```
log: mailto:a@b.net
log: http:///a@b.net
```

In Firefox:
```
log: mailto:a@b.net
log: http://software.hixie.ch/utilities/js/live-dom-viewer/a@b.net
```

valenting:
One of the ssues we've had in Firefox is setters that make the URL go into an invalid state.
For example:
```
#[test]
fn change_protocol() {
    let mut url = Url::parse("file:///[:::_1]/abc.html").unwrap(); // valid file path, but invalid hostname
    {
        let mut wrapper = UrlUtilsWrapper { url: &mut url, parser: &UrlParser::new()};
        wrapper.set_scheme("http");
    }
    let result = Url::parse(&url.serialize()).unwrap().serialize();
}
```
There are probably hundreds of these examples. Sometimes harmless, sometimes a big security risk. IMO the only way of safely changing the scheme is enforcing they the same type.

PS. there are things we don't allow in Firefox. Like this:
```
var url = new URL("http://example.com");
url.protocol = "data"; // this does nothing
```

SimonSapin:
I’ve filed https://www.w3.org/Bugs/Public/show_bug.cgi?id=28091 on the spec about this.

SimonSapin:
Ok, in the spec, a "relative flag" is maintained separately from the scheme. So in this test case, you’d end up with an URL in the `http` scheme but that is still a non-relative URL. What do you think?

valenting:
We could also reparse the url after setting the scheme, and roll it back if that fails. 
I think that would work.

SimonSapin:
Ideally we’d do what the spec says, and if that’s not good we should convince Anne to change the spec.

SimonSapin:
Alright, let's take this for now. I filed #85 about this issue, if we want to revsisit later.

PR #84, opened 2015-02-25
Bump rustc-serialize and use std::path

alexcrichton:
This commit bumps the dependency on rustc-serialize to 0.3 and allows
`Url::from_file_path` to work with `std::path::Path` as well (transitively
working for `PathBuf` too).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,5 +21,5 @@ version = "0.2"
 optional = true
 
 [dependencies]
-rustc-serialize = "0.2"
+rustc-serialize = "0.3"
 matches = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,7 +119,7 @@ assert!(css_url.serialize() == "http://servo.github.io/rust-url/main.css".to_str
 */
 
 
-#![feature(core, std_misc, collections, old_path)]
+#![feature(core, std_misc, collections, old_path, path, os)]
 
 extern crate "rustc-serialize" as rustc_serialize;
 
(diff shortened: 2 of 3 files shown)


alexcrichton:
Er sorry, forgot to fill out some other portions, adding tests and such now.

alexcrichton:
Ok! Filled out tests and verified that it's working on both unix and windows

SimonSapin:
Is it still useful to have both `Url::form_file_path` and `Url::form_directory_path`, in the new `std::path `, or does `Path` now record enough information to tell them apart? (This can be a follow up.)

It's unfortunate that we now can't manipulate/test paths for platforms other than the one we build for (compared to using e.g. `std::old_path::windows:: Path` directly). But I guess the complexity of multiple types and traits is not worth it.

alexcrichton:
Currently yes I believe that the distinction is available. Very little normalization happens by default so `a` is a distinct path from `a/` (which I think is what you want, right?).

For example: http://is.gd/l1SS8I

SimonSapin:
Right, this is what I meant. But I don’t know if we can rely on things like `std::env::current_dir` (when it’s ported to new path), or other user code that constructs a path to a directory, to include that `CurDir` component.

alexcrichton:
Ah yes that I would not rely on (I doubt `current_dir` returns paths with trailing slashes)

SimonSapin:
Ok, so we’ll keep the separate `from_*_path` methods.

Thanks for the PR!

PR #88, opened 2015-03-05
Reduce the number of unstable APIs used

alexcrichton:
I was able to get rid of the `collections` feature entirely and it looks like one of the `core` features, `Str`, isn't necessary.


SimonSapin:
Thanks!

PR #89, opened 2015-03-17
Fix windows build

alexcrichton:


SimonSapin:
Oops. The downside of not having both `std::path::unix` and `std::path::windows` is that I’d now need a Windows machine to test Windows paths…

PR #90, opened 2015-03-17
Adapt to new std::os::$platform modules

adrianheine:
In rust-lang/rust@1f5f76adc36082ec514a6cfa4ad2d7e6655a5253, OS-specific modules
in std:: were re-ordered. This change adapts to the new module layout.


SimonSapin:
Thanks!

PR #93, opened 2015-04-09
urlutils.rs - an empty password should be saved as None, not empty string

valenting:
Right now, if I set a username and password, then I clear them, the url serialization would be http://:@example.com
Easily solved just by setting the password to None when the input is empty string.


SimonSapin:
This matches the spec at https://url.spec.whatwg.org/#set-the-password . Thanks! Let me know if you need this on crates.io.

PR #94, opened 2015-04-22
Remove [#deprecated] attribute

liamfraser:
Fixes error: stability attributes may not be used outside of the standard library in latest nightly (rustc 1.0.0-nightly (2baf34825 2015-04-21) (built 2015-04-22))


SimonSapin:
Thanks, but I prefer removing the deprecated re-exports entirely. I’ve done this in 54b97f14bd24eacf2e9eaa089d9ee5be7fd2aa1c and published v0.2.30 on crates.io.

PR #97, opened 2015-04-24
Fix spelling for 'percent'

frewsxcv:


SimonSapin:
That’s automatic word completion for you… Thanks!

PR #99, opened 2015-04-24
General cleanup and modernization

frewsxcv:
Read through some of the code, cleaned up a few things along the way


SimonSapin:
I’m not of fan of `as &str`, but meh. Thanks!

PR #100, opened 2015-05-01
Percent-encode tilde in application/x-www-form-urlencoded

nox:


SimonSapin:
Why this change? It is against spec: https://url.spec.whatwg.org/#percent-encoded-bytes

Manishearth:
I guess we're mixing up https://url.spec.whatwg.org/#concept-urlencoded-byte-serializer and https://url.spec.whatwg.org/#percent-encoded-bytes ?

SimonSapin:
Looks like I completely misread the patch, sorry! Yes indeed, `FORM_URLENCODED_ENCODE_SET` is only used for the `application/x-www-form-urlencoded` serializer.

SimonSapin:
Let me know if you need this on crates.io and/or a backport to an earlier Rust version for Servo.

Manishearth:
Hopefully the Servo rustup will land soon.

nox:
If not a backport (that would be cool), at least I would want someone to explain to me how the fuck do I override rust-url with my own while still enabling the feature "query_encoding". =)

SimonSapin:
Ok, I published version 0.2.32 on crates.io.

To override, create a `.cargo/config` file (e.g. at the root of the servo repository, it’s is `.gitignore`d) that contains `paths = ["../path/to/rust-url"]`.

To enable `query_encoding`, have this in the `Cargo.toml` file of the crate where you want to use it:

```toml
[dependencies.url]
version = "0.2.32"
features = ["query_encoding"]
```

(And remove the previous entry for rust-url, something like `url = "0.2"` in the `[dependencies]` section.)

The two really should be orthogonal. If enabling features doesn’t work with overrides, that may be a bug in Cargo.

nox:
I reported it upstream: https://github.com/rust-lang/cargo/issues/1571.

SimonSapin:
`.cargo/config` is not the place where you specify features (`Cargo.toml` of the dependent is). It only specifies where to find source code.

Have you tried what I explained above?

PR #101, opened 2015-05-04
fix build on Windows

jaxx:
fix for the compile error "\src\lib.rs:997:5: 1005:6 error: the trait `core::marker::Sized` is not implemented for the type `[collections::string::String]` [E0277]" on Windows


SimonSapin:
Thanks!

PR #104, opened 2015-05-08
Percent encode curly braces

valenting:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1152455#c6 most UAs encode curly braces in the path of the URL.
I think rust-url should do the same.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -13,7 +13,7 @@ use std::fmt::{self, Formatter};
 use super::{UrlParser, Url, SchemeData, RelativeSchemeData, Host, SchemeType};
 use percent_encoding::{
     utf8_percent_encode_to, percent_encode,
-    SIMPLE_ENCODE_SET, DEFAULT_ENCODE_SET, USERINFO_ENCODE_SET, QUERY_ENCODE_SET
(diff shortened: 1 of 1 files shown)


SimonSapin:
The spec says it must not: https://url.spec.whatwg.org/#percent-encoded-bytes . If the spec is wrong, an issue should be filed on it and discussed before we unilaterally start deviating. (See "Participate" near the top of the spec, and feel free to @-mention me in spec issues.)

valenting:
https://github.com/whatwg/url/issues/16

valenting:
The spec has been changed: https://github.com/whatwg/url/issues/16

SimonSapin:
The spec added `{` and `}` to the "default encode set". I’d prefer doing that as well rather than some other change, even if the black-box behaviour happens to be identical (which I’m not convinced it is).

valenting:
Right, that sounds fair. I'm closing this PR. Let me know if you want me to submit another, or if it's faster to change it yourself.
Thanks!

SimonSapin:
Done in 79a9ef12c19cdc3d8a22daea1b528e25840d48ea

PR #107, opened 2015-05-09
implement FromStr for Url

blaenk:
This enables the use of the `parse()` method on `&str`

Fixes #18.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,7 @@ extern crate matches;
 
 use std::fmt::{self, Formatter};
 use std::hash;
+use std::str;
 use std::path::{Path, PathBuf};
 
 pub use host::{Host, Ipv6Address};
@@ -239,6 +240,13 @@ impl hash::Hash for Url {
     }
 }
 
+impl str::FromStr for Url {
+    type Err = ParseError;
+
+    fn from_str(url: &str) -> ParseResult<Url> {
+        Url::parse(url)
+    }
+}
 
 /// A set of optional parameters for URL parsing.
 pub struct UrlParser<'a> {
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -280,3 +280,8 @@ fn new_directory_paths() {
                                       "bar".to_string(), "".to_string()][..]));
     }
 }
+
+#[test]
+fn from_str() {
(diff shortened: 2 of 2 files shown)


seanmonstar:
Just warning, this has been closed as wontfix a few times now.

blaenk:
Thanks for the heads up, I guess I should've searched, but I didn't think it would be particularly controversial. I've contributed the same change to the `rust-lang/regex` and `rust-lang/glob` packages and they were merged.

blaenk:
From @SimonSapin in #18:

> I’ve resisted doing this so far because I really don’t like FromStr. It feels like "too much magic" to me: when reading code that uses it, I sometimes have no idea what the expected type is because the compiler infers it.

Back then it was typically used with `from_str`, but nowadays you have the `parse` method which is a common interface for parsing all kinds of types, and is covered pretty [early on in the rust book](http://doc.rust-lang.org/book/guessing-game.html). Yeah the type is still inferred, but I don't think it's all that confusing when you have something like `"thing".parse::<Type>()` or `let ty: Type = "thing".parse().unwrap()`. Another difference from the time that was written is that `parse()` now yields a `Result`, so you don't lose error information when the parsing fails.

[`rust-lang/regex`](https://github.com/rust-lang/regex/pull/70) and [`rust-lang/glob`](https://github.com/rust-lang/glob/pull/39) also have impls for this trait, and I've seen it in many other packages as well, such as the [toml](http://alexcrichton.com/toml-rs/toml/enum.Value.html) package. It's a nice, uniform way of parsing strings, especially for types like these that map very readily to existing methods.

SimonSapin:
I’m still not a huge fan of `FromStr`, but I suppose it doesn’t hurt much to implement it.

Thanks for the patch!

SimonSapin:
Published as v0.2.43 on crates.io.

blaenk:
Sweet thanks!

On Sunday, May 10, 2015, Simon Sapin <notifications@github.com> wrote:

> Published as v0.2.43 on crates.io.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/servo/rust-url/pull/107#issuecomment-100611886>.
>


-- 
- Jorge Israel Peña

PR #108, opened 2015-05-11
Prevent parsing brackets from causing panic

frewsxcv:
The code:

```
Url::parse("http://[]")
```

The panic:

```
thread '<main>' panicked at 'index out of bounds: the len is 0 but the index is 0',
/Users/coreyf/.cargo/registry/src/github.com-1ecc6299db9ec823/url-0.2.34/src/host.rs:104
```

in addition to:

The code:

```
Url::parse("http://[:]")
```

The panic:

```
/Users/coreyf/.cargo/registry/src/github.com-1ecc6299db9ec823/url-0.2.34/src/host.rs:104
thread 'tests::url_parsing' panicked at 'index out of bounds: the len is
1 but the index is 1', src/host.rs:110
```

Was found using https://github.com/kmcallister/afl.rs 👍

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -101,6 +101,11 @@ impl Ipv6Address {
         let mut piece_pointer = 0;
         let mut compress_pointer = None;
         let mut i = 0;
+
+        if len < 2 {
+            return Err(ParseError::InvalidIpv6Address)
+        }
+
         if input[0] == b':' {
             if input[1] != b':' {
                 return Err(ParseError::InvalidIpv6Address)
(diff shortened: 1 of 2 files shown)


frewsxcv:
Dont merge this yet, theres another bug in here

SimonSapin:
Nice! But I think a similar issue exists for `Url::parse("http://[:]")`. Could you change `if len == 0` to `if len < 2`? Reading the spec, I think any input of length 1 is also invalid.

frewsxcv:
@SimonSapin Yeah, I was just about to add that in :)

frewsxcv:
Should be reviewable; tests pass

SimonSapin:
Thanks

PR #111, opened 2015-05-17
Implemented the origin method (fixes #54)

### LINKED ISSUE #54: Method to extract origin from a URL
Currently, there is no method to extract the origin of a URL as String
* eg: from https://example.com:490/samplefolder/samplefile.txt, extract https://example.com:490/

It would be better if there is a method which does this, otherwise one has to use url.scheme, url.domain() and url.port() to construct the origin.

KiChjang:
Fixes #54, but still need some clarafication as to how to generate a GUID for schemes such as blob, file and others.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/111)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,5 +21,6 @@ version = "0.2"
 optional = true
 
 [dependencies]
+uuid = "0.1.17"
 rustc-serialize = "0.3"
 matches = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,6 +119,7 @@ assert!(css_url.serialize() == "http://servo.github.io/rust-url/main.css".to_str
 */
 
 extern crate rustc_serialize;
+extern crate uuid;
 
 #[macro_use]
 extern crate matches;
@@ -135,6 +136,8 @@ use percent_encoding::{percent_encode, lossy_utf8_percent_decode, DEFAULT_ENCODE
 
 use format::{PathFormatter, UserInfoFormatter, UrlNoFragmentFormatter};
 use encoding::EncodingOverride;
+
+use uuid::Uuid;
 
 mod encoding;
 mod host;
@@ -184,7 +187,7 @@ pub struct Url {
 #[derive(PartialEq, Eq, Clone, Debug)]
 pub enum Origin {
     /// A globally unique identifier
-    UID(usize),
+    UID(String),
 
     /// Consists of the URL's scheme, host and port
     Tuple(String, Host, u16)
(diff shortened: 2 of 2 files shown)


Manishearth:
Are we sure that the port will always be the default one? With a quick look at the code that seems to be the case.

Otherwise lgtm.

Generating guids probably can be done with https://crates.io/crates/uuid

KiChjang:
@Manishearth Where do you see that port always return the default one?

Manishearth:
As in the port included in a URL that didn't specifically say one.

KiChjang:
Good catch, there's a method called port_or_default

SimonSapin:
> Generating guids probably can be done with https://crates.io/crates/uuid.

That’s unclear to me. I’ve filed https://github.com/whatwg/url/issues/21 on the spec.

SimonSapin:
I think that "globally unique identifiers" for the purpose of URL origins are **not** UUIDs, but are the same as "opaque identifier" as defined in HTML: https://html.spec.whatwg.org/multipage/browsers.html#origin

Manishearth:
I don't know; Blob URIs use GUIDs I think.

SimonSapin:
@Manishearth [citation needed] :)

Manishearth:
https://w3c.github.io/FileAPI/#unicodeSerializationOfBlobURL

SimonSapin:
http://krijnhoetmer.nl/irc-logs/whatwg/20150522#l-267

KiChjang:
Hmm... how exactly do we implement opaque identifiers?

SimonSapin:
Unclear to me. I’m trying to figure this out:

https://github.com/whatwg/url/issues/21
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28675

Manishearth:
> Internal values, with no serialisation, for which the only meaningful operation is testing for equality.

Would a wrapped 256-bit hash (or something) do?

SimonSapin:
A hash of what data?

Manishearth:
Oh, sorry, I got confused there.

We could just hash random values but we're back to GUIDs.

Really, we can just use uuid or similar to create a number and then wrap the number in an opaque (except for PartialEq/Eq) tuple  struct.

KiChjang:
What exactly is this opaque tuple struct? How does it look like? Does my current implementation suffice?

Manishearth:
`#[derive(Debug, PartialEq, Eq, Copy)] pub struct OpaqueOrigin(u64);`

Inner value is private, thus, opaque.

-----Original Message-----

SimonSapin:
This still seems very wrong to me, but apparently generating a new random identifier every time the method is called is indeed intentional: https://github.com/whatwg/url/issues/56

@bors-servo r+

PR #118, opened 2015-07-10
Add optional serialization with `serde`

pcwalton:
r? @SimonSapin

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,12 +14,19 @@ license = "MIT/Apache-2.0"
 [features]
 
 query_encoding = ["encoding"]
+serde_serialization = ["serde"]
 
 [dependencies.encoding]
 
 version = "0.2"
 optional = true
 
+[dependencies.serde]
+
+version = "*"
+optional = true
+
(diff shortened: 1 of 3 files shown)


SimonSapin:
Please make a few changes:

* Use a `_` in the feature name, to be consistent with `query_encoding`. (Or change `query_encoding` to use `-`, I’d be surprised if anyone uses it.)
* Add `#[cfg(…)]` attributes on both `impl`s.
* Add this doc-comments like [this one](https://github.com/servo/rust-url/blob/67fbc8db2e5226f1a7c6f9dc38307d28b25402a8/src/lib.rs#L281) to both `impls`s.
* Add a line to `Makefile` so that Travis tests both with and without this feature. (Travis runs `make test`.)
* Change the `cargo doc` line in `Makefile` so that all features are enabled. According to `cargo help doc`, the value for `--features` is space-separated.

pcwalton:
@SimonSapin Done. r?

SimonSapin:
v0.2.36 is on crates.io

PR #119, opened 2015-07-29
IDNA support

valenting:
I wrote a UTS 46 implementation based on http://intertwingly.net/projects/pegurl/idna.js
It needs a fast path for ascii codepoints, but I wanted to get some feedback on it first.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/119)


make_idna_table.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2013-2014 Valentin Gosu.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+
+# Run as: python make_idna_table.py idna_table.txt > src/idna_table.rs
```
SimonSapin: Is `idna_table.txt` provided by Unicode? Where can it be found?

valenting: http://www.unicode.org/Public/idna/latest/IdnaMappingTable.txt

src/idna.rs
```diff
@@ -0,0 +1,74 @@
+use idna_mapping::*;
+use std::char;
+
+fn idna_mapped(mapping: &'static [u32]) -> Result<String, &'static str> {
+  let mut ret = "".to_string();
```
SimonSapin: Please use four space indents, for consistency with other files.

src/host.rs
```diff
@@ -55,10 +57,20 @@ impl Host {
...
+
+            let mut domain = match idna::domain_to_ascii(&domain) {
+                Ok(s) => s,
+                Err(_) => return Err(ParseError::InvalidDomainCharacter)
+            };
+
             if !domain.is_ascii() {
-                Err(ParseError::NonAsciiDomainsNotSupportedYet)
-            } else if domain.find(&[
+                domain = match punycode::encode_str(&domain) {
```
SimonSapin: Converting to punycode should be part of `domain_to_ascii` per http://www.unicode.org/reports/tr46/#ToASCII , and should happen on individual dot-separated labels, not the entire domain.

src/host.rs
```diff
@@ -55,10 +57,20 @@ impl Host {
...
+                    Some(dom) => "xn--".to_string() + &dom,
```
SimonSapin: I think that adding the `xn--` prefix should also be part of `domain_to_ascii`, but UTS#46 has a bug: https://github.com/whatwg/url/issues/53

SimonSapin:
Could you also integrate the tests form http://www.unicode.org/Public/idna/latest/IdnaTest.txt ?

----

I’m considering rewriting the Python scripts to Rust and running them as [Cargo build scripts](http://doc.crates.io/build-script.html) to avoid having generated files in the repository. However in Gecko it’s not as easy if you’re not using Cargo. It would be a separate `.rs` file to be compiled for the host (does Gecko do cross-compilation?), run with an `OUT_DIR` environment variable, and have that run be a dependency of building the url crate, which should be done with the same environment variable.

How painful is it to do in Gecko? If it would make your like more difficult I can stick to Python and generated files in the repository.

valenting:
It would be a bit harder to have the Gecko build system run the rust files at build time, but at the moment that's not a blocker. We can include the the generated rust-generated files just as we do with the python generated ones at the moment.

make_idna_table.py
SimonSapin: Please add this URL in a code comment.

SimonSapin:
See https://github.com/servo/rust-url/issues/152

valenting:
I've implemented the BIDI checks, and now all of the tests pass (except the contextJ tests, which client software isn't required to pass).
For combining_mark I call into unicode_normalization::char::canonical_combining_class - which I fear might not map correctly to General_Category=Mark. The unicode-normalization table generator has this data, but doesn't use it:
https://github.com/unicode-rs/unicode-normalization/blob/master/scripts/unicode.py#L92

SimonSapin:
Other than a few nits I think this is ready to land. Thanks again!

SimonSapin:
Not sure what’s happening with homu, but the Travis build for the PR push is green.

PR #123, opened 2015-08-09
Add heapsize support.

notriddle:
Part of servo/heapsize#5.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/123)


SimonSapin:
Since we’re in the crate defining the types now, could `#[derive(HeapSizeOf)]` be used instead? Using the plugin requires unstable Rust, but that’s OK since the dependency is optional. (So it’d really be `#[cfg_attr(feature = "heap_size", derive(HeapSizeOf))]`.)

notriddle:
Sure.

PR #130, opened 2015-09-16
Adds docs for form_urlencoded::parse

conradkleinespel:
Hello,

I found myself wondering if this function would actually decode the values, as opposed to just splitting on the `=` sign and making some validity checks. I would be nice to have docs just say that the names/values are decoded.

Best regards,
Conrad

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/130)


SimonSapin:
That’s a good point, thanks!

@bors-servo r+

Issue #114, opened 2015-05-30
Implement Ord/PartialOrd?


alexcrichton:
I notice that currently these traits aren't implemented for `Url`, and I was curious if this was done on purpose? If not, would it be possible to add them?

SimonSapin:
I don’t think it makes sense to consider an URL intrinsically "greater" or "lesser" than another. But what’s the use case?

alexcrichton:
The need for this is derived from Cargo, which primarily uses it for deterministic output in a few locations. There are spots where we sort an array of `PackageId` (which transitively contains a Url at some point) to ensure that the output is consistent among stages. Put another way, the order itself isn't super important so long as it's consistent.

I personally view core traits like this as being useful to define on core types like `Url`, even when they don't always make 100% sense (e.g. `PartialEq` for `RefCell`)

SimonSapin:
Ok, that sounds like a real use case and I don’t have better way to solve it. I still don’t think any ordering is meaningful for URLs, but adding a few `#[derive(PartialOrd)]` is not harmful either.

I don’t get your second point, though. What qualifies as a core trait? Should `Url` implement `Add`, for example? What qualifies as a core type? What makes `Url` more "core" than, say, `cssparser::RGBA`? As to `RefCell`, since it’s a container it doesn’t need to have a meaningful equality by itself, it just defers to its contents.

alexcrichton:
Ah I just mean "core" in the sense that it's often a building block for many other types. For example there are probably going to be a lot of structs which store a `Url`, some of which may want to have some sort of ordering eventually.

seanmonstar:
As for whether Ordering makes sense for Url: it seems natural that it would
be the same as Ord for strings: likely alphabetically.

SimonSapin:
Ordering URLs by the string ordering of their serialization seems reasonable, but it’s not the same ordering as what `#[derive(PartialOrd)]` would get us with the current internal representation. Still, we should probably stick with `derive`, to avoid performance issues like #113.

PR #133, opened 2015-11-02
Make port: None equal to port: default_port

### LINKED ISSUE #132: `port: None` should be equal to `port: default_port`
The iron framework constructs an URL with `RelativeSchemeData { port: 42, default_port: 42, ... }` (see iron/iron#400). The url crate should treat this as equal to `RelativeSchemeData { port: None, default_port: 42 }`.

At least this behavior would make sense for HTTP URLs.

untitaker:
Fix #132

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/133)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -137,6 +137,7 @@ use std::fmt::{self, Formatter};
 use std::str;
 use std::path::{Path, PathBuf};
 use std::borrow::Borrow;
+use std::hash::{Hash, Hasher};
 
 #[cfg(feature="serde_serialization")]
 use std::str::FromStr;
@@ -226,7 +227,7 @@ pub enum SchemeData {
 }
 
 /// Components for URLs in a *relative* scheme such as HTTP.
-#[derive(PartialEq, Eq, Clone, Debug, Hash, PartialOrd, Ord)]
+#[derive(Clone, Debug, PartialOrd, Ord)]
 #[cfg_attr(feature="heap_size", derive(HeapSizeOf))]
 pub struct RelativeSchemeData {
     /// The username of the URL, as a possibly empty, percent-encoded string.
@@ -266,6 +267,33 @@ pub struct RelativeSchemeData {
     pub path: Vec<String>,
 }
 
+impl RelativeSchemeData {
+    fn get_identity_key(&self) -> (&String, &Option<String>, &Host, Option<u16>, Option<u16>, &Vec<String>) {
+        (
+            &self.username,
+            &self.password,
+            &self.host,
+            self.port.or(self.default_port),
+            self.default_port,
+            &self.path
+        )
+    }
+}
+
+
+impl PartialEq for RelativeSchemeData {
+    fn eq(&self, other: &RelativeSchemeData) -> bool {
+        self.get_identity_key() == other.get_identity_key()
+    }
+}
+
+impl Eq for RelativeSchemeData {}
+
+impl Hash for RelativeSchemeData {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.get_identity_key().hash(state)
+    }
+}
 
 impl str::FromStr for Url {
     type Err = ParseError;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,6 +138,7 @@ use std::str;
 use std::path::{Path, PathBuf};
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
+use std::cmp::Ordering;
 
 #[cfg(feature="serde_serialization")]
 use std::str::FromStr;
@@ -227,7 +228,7 @@ pub enum SchemeData {
 }
 
 /// Components for URLs in a *relative* scheme such as HTTP.
-#[derive(Clone, Debug, PartialOrd, Ord)]
+#[derive(Clone, Debug)]
 #[cfg_attr(feature="heap_size", derive(HeapSizeOf))]
 pub struct RelativeSchemeData {
     /// The username of the URL, as a possibly empty, percent-encoded string.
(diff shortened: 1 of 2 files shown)


untitaker:
cc @SimonSapin

untitaker:
Will add tests if you agree with the concept.

src/lib.rs
```diff
@@ -226,7 +226,7 @@ pub enum SchemeData {
 }
 
 /// Components for URLs in a *relative* scheme such as HTTP.
-#[derive(PartialEq, Eq, Clone, Debug, Hash, PartialOrd, Ord)]
+#[derive(Clone, Debug, Hash, PartialOrd, Ord)]
```
eefriedman: If you're not deriving PartialEq, you can't derive Hash, PartialOrd, or Ord.

untitaker: Should rustc catch that?

Manishearth: It should, yes.

Hash is fine, it's the Ord traits which are problematic.

eefriedman: Probably... please file a bug (https://github.com/rust-lang/rust/issues/new).

eefriedman: Err, how is Hash not a problem.  From https://doc.rust-lang.org/nightly/std/hash/trait.Hash.html: "if two keys are equal, their hashes should also be equal".

untitaker: I see. So these two behaviors -- comparing an URL with explicit port vs non-explicit and total order -- are mutually exclusive? Which one is more desired?

Manishearth: @eefriedman Before that: "If you are also implementing Eq, there is an additional property that is important:". So there is an invariant to keep in mind when implementing both Eq and Hash. However, Eq and Hash do not need each other, you can just implement one if you want.

untitaker: Perhaps we can explicitly impl Hash and derive Eq, Ord from that?

Manishearth: I guess you should add a manual Hash impl which hashes a default_port instead of None when it encounters None.

untitaker: Can rustc derive a correct Eq impl from a custom Hash impl?

Manishearth: Nope. It doesn't, and it's not possible to either, since there isn't enough information on what choices to make. In general an unsolvable problem.

untitaker: Why not? Return `hash(self) == hash(other)` from `eq` method.

Anyway, thanks for the quick review, I'll get to work.

Manishearth: Hashes may have collisions. Equal keys imply equal hashes, not the other way around.

untitaker:
I've updated the PR with custom Hash and Eq impls.

Should I do the same for Ord? Do we even need to order URLs?

untitaker:
Any update on this? Is there something left to do?

Manishearth:
LGTM, but I'll wait for Simon to review it

SimonSapin:
Yes, Ord and PartialOrd should be consistent with Eq and PartialEq. Please change them to use the identity key as well, and add those tests you mentioned :)

I also think that ordering URLs doesn't make sense, but Cargo wants to sort things that contain them: https://github.com/servo/rust-url/issues/114

untitaker:
* Added some tests for equality.

* I'm not sure how ordering should look like. Do we want the same exact ordering behavior as before, i.e. do we need extensive tests? Or is breakage fine as long as it doesn't happen too often? It's "obvious" that this new impl is deterministic. cc @alexcrichton 

* The hashing tests are ugly, `check_eq` was supposed to be a macro but then I somehow hit a ICE

src/lib.rs
untitaker: BTW I've filed the `derive(Hash)` issue against rustc: https://github.com/rust-lang/rust/issues/29758

src/lib.rs
```diff
@@ -266,6 +268,45 @@ pub struct RelativeSchemeData {
...
+            &self.host,
+            self.port.or(self.default_port),
+            self.default_port,
+            &self.path
+        )
+    }
+}
+
+
+impl PartialEq for RelativeSchemeData {
```
untitaker: I'd be interested into factoring out all these boilerplate impls into a new crate. Would that be something useful, and would it be worth adding as dependency?

I haven't decided on an API yet, but perhaps a macro that impls PartialEq, Hash, etc given a type and a function like `get_identity_key`.

SimonSapin:
@bors-servo r+

Looks good, thanks!

It’s not important that the ordering is the same as before. Only that it’s consistent (with the usual properties: total, anti-symmetric, transitive) within one execution of a program. The current impls are fine. I don’t think ordering tests are important here, the impls are very straightforward.

Yes, the impls are boilerplate-ish, but they’re only one line each. I’m skeptical that it’s useful to have a crate just for this, at least until there’s more than one thing that would use it.

untitaker:
>within one execution of a program

I don't exactly know where Cargo uses them, but I thought that it'd produce massive diffs for e.g. `.lock`-files when upgrading cargo and then running `cargo update`. I guess that's not a real issue though.

SimonSapin:
@alexcrichton, how do you feel about the ordering of URLs changing?

alexcrichton:
Does this radically change how URLs are ordered, or is it basically in just some niche cases the ordering gets "more correct" ?

SimonSapin:
URLs that do not specify a port number now sort as if they specified the scheme’s default port number. If none of the URLs in a given set specify it (which is the common case), I *think* the relative order is is unchanged.

alexcrichton:
Eh that seems fine. The ordering of URLs is already a little ambiguous anyway I think, so tweaking it a bit should be fine.

PR #134, opened 2015-11-12
Added VALUE_CHARS EncodeSet

mikedilger:
This character set is based off of rfc5987 section 3.2
https://tools.ietf.org/html/rfc5987#section-3.2
for the ABNF `value-chars` which is used by certain HTTP
header field values such as Content-Disposition `filename*`

Not sure about the constant name 'VALUE_CHARS'

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/134)


SimonSapin:
This is HTTP-specific, right? How about `HTTP_VALUE`?

To use it from other crates you’ll also want to add an `EncodeSet` constant in `percent_encoding.rs`. Please include the URL to the definition in a doc-comment there.

mikedilger:
Ok, I think I've addressed your concerns in this last commit.

SimonSapin:
Looks good.

@bors-servo r+

Issue #116, opened 2015-06-13
Invalid IPv4 addresses are not rejected


nox:
https://github.com/w3c/web-platform-tests/blob/11f3aee19205a2ae97efa1cbeb211a2395192ea8/url/urltestdata.txt#L325

SimonSapin:
I believe this test does not match the current spec, but in this case I think the spec should be changed. Spec issue: https://www.w3.org/Bugs/Public/show_bug.cgi?id=26431

frewsxcv:
Spec has been resolved

nox:
This was closed, but we never added `http://192.168.0.257` to the tests.

PR #135, opened 2015-11-14
Add IPv4 variant to Host and remove custom Ipv6 address type

pyfisch:
The Whatwg URL Standard was changed to describe Ipv4 address
parsing. Before this change there was no difference made
between domains and Ipv4 addresses. This change also removes
the custom Ipv6 type and uses the types provided by std for
ip addresses.

https://url.spec.whatwg.org/#syntax-host-ipv4

Note `heap_size` feature is broken because the HeapSizeOf trait is not yet implemented for the ip addresses.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/135)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -7,37 +7,23 @@
 // except according to those terms.
 
 use std::ascii::AsciiExt;
-use std::cmp;
 use std::fmt::{self, Formatter};
+use std::net::{Ipv4Addr, Ipv6Addr};
 use parser::{ParseResult, ParseError};
-use percent_encoding::{from_hex, percent_decode};
+use percent_encoding::{percent_decode};
 
 
 /// The host name of an URL.
 #[derive(PartialEq, Eq, Clone, Debug, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature="heap_size", derive(HeapSizeOf))]
 pub enum Host {
-    /// A (DNS) domain name or an IPv4 address.
-    ///
-    /// FIXME: IPv4 probably should be a separate variant.
-    /// See https://www.w3.org/Bugs/Public/show_bug.cgi?id=26431
+    /// A (DNS) domain name.
     Domain(String),
-
-    /// An IPv6 address, represented inside `[...]` square brackets
-    /// so that `:` colon characters in the address are not ambiguous
-    /// with the port number delimiter.
-    Ipv6(Ipv6Address),
+    /// An IPv4 address.
+    V4(Ipv4Addr),
+    /// An IPv6 address.
+    V6(Ipv6Addr),
 }
-
-
-/// A 128 bit IPv6 address
-#[derive(Clone, Eq, PartialEq, Copy, Debug, Hash, PartialOrd, Ord)]
-pub struct Ipv6Address {
-    pub pieces: [u16; 8]
-}
-#[cfg(feature="heap_size")]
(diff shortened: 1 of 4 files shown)


pyfisch:
https://github.com/pyfisch/heapsize/tree/4bdd4e2524e506d9dd23115dff6a9a2249fe6cc4 implements `HeapSizeOf` for ip addresses.

pyfisch:
By the way this solves https://github.com/servo/rust-url/issues/116

untitaker:
@pyfisch I think you should file a PR against heapsize regarding that commit.

Manishearth:
Blocked on https://github.com/servo/heapsize/pull/17 getting r+ since crates and the repo weren't in sync.

Ms2ger:
Please bump the heapsize dep to `0.1.3`.

pyfisch:
done

Ms2ger:
And squash the commits, please

pyfisch:
done

Ms2ger:
Thanks!

src/host.rs
```diff
@@ -51,22 +37,30 @@ impl Host {
...
-            let decoded = percent_decode(input.as_bytes());
-            let domain = String::from_utf8_lossy(&decoded);
-            // TODO: Remove this check and use IDNA "domain to ASCII"
-            if !domain.is_ascii() {
-                Err(ParseError::NonAsciiDomainsNotSupportedYet)
-            } else if domain.find(&[
-                '\0', '\t', '\n', '\r', ' ', '#', '%', '/', ':', '?', '@', '[', '\\', ']'
-            ][..]).is_some() {
-                Err(ParseError::InvalidDomainCharacter)
+            if let Ok(addr) = input.parse() {
```
SimonSapin: @pyfisch @Ms2ger have you verified that `Ipv4Addr::parse` and `Ipv6::parse` match https://url.spec.whatwg.org/#host-parsing in every edge case? For example: leading zeros, hex/octal components in v4, …

SimonSapin:
I’ve reverted this change. It looks like there was a misunderstanding in the review, sorry. At least the IPv4 parsing of the standard library differs from https://url.spec.whatwg.org/#concept-ipv4-parser . (Note that the "parsing" and "serialization" sections of the spec are more relevant to implementation than the "syntax" sections, which only give a high level overview as opposed to exact algorithms.)

I don’t know if the current parsing or serialization behavior of std for *IPv6* happens to match the current spec, but I don’t think we should rely on it staying that way. (I don’t know if it should. Web compatibility with all its weirdness is an overriding concern for browsers, but not as much for a programming language standard library.)

@pyfisch, could you please open a new PR? But keep the existing `fmt` and `parse` methods for IPv6, and write IPv4 parsing per https://url.spec.whatwg.org/#concept-ipv4-parser . Or let us know if you’d rather someone else does this work.

SimonSapin:
(Edge cases to look for in IPv4 include: leading zeroes, hexadecimal or octal components, less than 4 components.)

pyfisch:
I will create a new pull request which will use custom parse methods as they are described in the urlspec. The canonical formatting is unambiguous so I will try to use the methods provided by std but add many tests for edge cases.

SimonSapin:
Please also add "Fixes #116" in the new PR or commit message. Thanks!

PR #137, opened 2015-11-20
Implement 'url!(..)' macro

### LINKED ISSUE #136: Compile-time url parsing 'url!' macro
Inspired by [sfackler/rust-postgres-macros](https://github.com/sfackler/rust-postgres-macros) which is a compile time SQL parser

There are ton of instances in servo, where we do something like `Url::parse("http://mozilla.com").unwrap()`. It'd be great if we could do that validation at compile time with a macro (e.g. `url!("http://mozilla.com")`).

I have no idea how feasible this is, just throwing around the idea.

frewsxcv:
Fixes #136

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/137)

### DIFF
--- a/plugin/Cargo.toml
+++ b/plugin/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "url_plugin"
+version = "0.1.0"
+authors = ["Corey Farwell <coreyf@rwell.org>"]
+
+[lib]
+name = "url_plugin"
+plugin = true
+
+[dependencies.url]
+path = "../"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
 /Cargo.lock
 /.cargo/config
+/plugin/target
+/plugin/Cargo.lock
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -1,0 +1,149 @@
+#![feature(plugin_registrar, quote, rustc_private)]
+#![crate_type="dylib"]
+
+extern crate rustc;
+extern crate syntax;
+extern crate url;
+
+use rustc::plugin::Registry;
+use std::error::Error;
+use syntax::ast::{TokenTree, ExprLit, LitStr, Expr};
+use syntax::codemap::Span;
+use syntax::ext::base::{ExtCtxt, MacResult, MacEager, DummyResult};
+use syntax::ext::build::AstBuilder;
+use syntax::fold::Folder;
+use syntax::parse;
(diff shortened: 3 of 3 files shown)


frewsxcv:
I preemptively opened this because I have a couple questions:

1. How should tests be structured/organized for compiler plugins?
2. Should the `url` crate depend on `url_plugin`? Wouldn't that create a cyclic dependency?

plugin/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "url_plugin"
+version = "0.1.0"
+authors = ["Corey Farwell <coreyf@rwell.org>"]
+
+[lib]
+name = "url_plugin"
+plugin = true
+
+[dependencies.url]
+path = "../"
```
Manishearth: I think for publishing to work you need to have a version dep here too. Not sure

SimonSapin:
To answer questions:

1. I don’t think many tests are needed here, one to check that the plugin doesn’t panic and returns something sensible should be enough. It’d have to be in a separate crate so you can use the plugin. Cargo will compile a separate crate (run a separate `rustc --test` invocation) if you have a file named e.g. `plugin/tests/smoke_test.rs`, so you don’t need another `Cargo.toml` file.
2. No it shouldn’t, why? Yes it would.

The code looks ok but (sorry I didn’t mention this earlier) I wonder if this plugin should be here rather than in Servo. We can move it back if someone asks, but in the meantime I don’t know if anyone else would use it. Meanwhile, this code uses unstable compiler internals (as plugins do) and rust-url is sort of expected to work on Rust Nightly, while Rust-in-Servo is updated less frequently.

To put it more bluntly, moving it to Servo is likely less future work for me :]

Manishearth:
>  I wonder if this plugin should be here rather than in Servo

We don't need to ship this crate as part of rust-url. It can be published separately and used only by Servo.

SimonSapin:
I’m suggesting having this in the servo/servo repository rather than this one, where the Rust version used on CI is updated less frequently. This is not about publishing on crates.io.

frewsxcv:
Suggestions on where it should live in Servo?

SimonSapin:
The `plugins` crate?

frewsxcv:
https://github.com/servo/servo/issues/8622

DenisGorbachev:
The `url!` macro for compile-time validation is now available via [`url-macro`](https://crates.io/crates/url-macro) crate.

Disclaimer: I'm the author.

PR #138, opened 2015-11-20
Change host to use ip address types provided by std.

### LINKED ISSUE #116: Invalid IPv4 addresses are not rejected

pyfisch:
Removes the custom IPv6Addr type and replaces it with the std one.
Parses IPv4Addrs to the std type using the parser described in the
url.spec.whatwg.org handling all edge cases. Add tests.

fixes #116

This is a breaking change. Version bumped to v0.5.0.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/138)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "0.4.0"
+version = "0.5.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
--- a/src/host.rs
+++ b/src/host.rs
@@ -9,6 +9,7 @@
 use std::ascii::AsciiExt;
 use std::cmp;
 use std::fmt::{self, Formatter};
+use std::net::{Ipv4Addr, Ipv6Addr};
 use parser::{ParseResult, ParseError};
 use percent_encoding::{from_hex, percent_decode};
 
@@ -17,26 +18,15 @@ use percent_encoding::{from_hex, percent_decode};
 #[derive(PartialEq, Eq, Clone, Debug, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature="heap_size", derive(HeapSizeOf))]
 pub enum Host {
-    /// A (DNS) domain name or an IPv4 address.
-    ///
-    /// FIXME: IPv4 probably should be a separate variant.
-    /// See https://www.w3.org/Bugs/Public/show_bug.cgi?id=26431
+    /// A (DNS) domain name.
     Domain(String),
-
+    /// A IPv4 address, represented by four sequences of up to three ASCII digits.
+    Ipv4(Ipv4Addr),
     /// An IPv6 address, represented inside `[...]` square brackets
     /// so that `:` colon characters in the address are not ambiguous
     /// with the port number delimiter.
-    Ipv6(Ipv6Address),
+    Ipv6(Ipv6Addr),
 }
-
-
-/// A 128 bit IPv6 address
-#[derive(Clone, Eq, PartialEq, Copy, Debug, Hash, PartialOrd, Ord)]
-pub struct Ipv6Address {
-    pub pieces: [u16; 8]
-}
-#[cfg(feature="heap_size")]
-known_heap_size!(0, Ipv6Address);
 
 
 impl Host {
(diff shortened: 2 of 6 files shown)


SimonSapin:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

_[src/host.rs, line 88 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_t7KGfJ62PGH55w6p:-K3_t7KGfJ62PGH55w6q:1123238933) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L88)):_
Nit: The `input.len()` check here is not necessary, `starts_with` already does it.

---

_[src/host.rs, line 89 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_tL1JEEY0ctkmLGRI:-K3_tL1JEEY0ctkmLGRJ:483181460) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L89)):_
Nit: rather than keep integers indices around, you could have `input = &input[2..]` here. (If you also change `input: &str` to `mut input: &str` in the function definition line)

---

_[src/host.rs, line 124 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_vvEyOeteyXqOSqep:-K3_vvEyOeteyXqOSqeq:945196264) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L124)):_
`expect` panics if the option is `None`. Panic should only happen in situations that should not happen unless there is a bug. Are you confident that `numbers` here is never empty? What about `Host::parse("")`? (Same comment for the other `.expect()` call below.)

---

_[src/host.rs, line 126 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_vSuxH7b_YlYfjPh-:-K3_vSuxH7b_YlYfjPh0:-540838155) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L126)):_
I don’t think this should return https://url.spec.whatwg.org/#syntax-violation says that a syntax violation is not fatal and does not terminate the parser. In this was a compiler, it would be called a warnings. Since we don’t have a way of logging warnings at the moment, please ignore when the spec says syntax violation, or simply add a comment.

---

_[src/host.rs, line 128 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_xRR9grsLSBemePyV:-K3_xRR9grsLSBemePyW:994790903) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L128)):_
I don’t know if `u32::pow` is guaranteed to compile down to a bit shift here. Please use `<<` instead. (Multiplying by 256 is the same as shifting by 8 bits.) Same for the other `.pow()` below.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/138)_

pyfisch:
_[src/host.rs, line 89 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_tL1JEEY0ctkmLGRI:-K3a3JT0zu8YEe3Y545p:2756778) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L89)):_
Wow, I didn't knew the borrow checker would be so smart and accept this. :)

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/138)_

pyfisch:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

_[src/host.rs, line 124 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_vvEyOeteyXqOSqep:-K3a4AYZOo1t9XYu19L3:621943253) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L124)):_
It is a bug in the code if the list is empty. `Host::parse("")` does not work because empty hosts are rejected by the host parser. The spec assumes that the list will not be empty because it says `Let ipv4 be the last item in numbers. ` there is no handling for the empty list case. I also tried to come up with an example to produce an empty list but failed.

---

_[src/host.rs, line 126 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_vSuxH7b_YlYfjPh-:-K3aF5qcPH_OI7XaYhQw:110520) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L126)):_
https://url.spec.whatwg.org/#concept-ipv4-parser step 9 says "If any but the last item in numbers is greater than 255, return failure." So this is no syntax violation but a failure.

---

_[src/host.rs, line 128 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_xRR9grsLSBemePyV:-K3aFPcneDGelRmScKc-:-499146009) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L128)):_
Done.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/138)_

SimonSapin:
Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[src/host.rs, line 126 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3_vSuxH7b_YlYfjPh-:-K3aR0SLk973PMVK823G:974883214) ([raw file](https://github.com/servo/rust-url/blob/43a6a4018f70b76f5551d3f043c4bc94734c163a/src/host.rs#L126)):_
Ah, sorry, I was looking as step 8 and did not read carefully enough.

---

_[src/host.rs, line 123 \[r2\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3aS4cIYbJUsqPJGC_E:-K3aS4cIYbJUsqPJGC_F:-193409653) ([raw file](https://github.com/servo/rust-url/blob/c8881024b8ca2b2b3803fefb00b4dacb27a2b533/src/host.rs#L123)):_
Shouldn’t this be a strict `>`?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/138)_

pyfisch:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

_[src/host.rs, line 123 \[r2\]](https://reviewable.io:443/reviews/servo/rust-url/138#-K3aS4cIYbJUsqPJGC_E:-K3aV_Yn2ln115Xdss4t:-499146009) ([raw file](https://github.com/servo/rust-url/blob/c8881024b8ca2b2b3803fefb00b4dacb27a2b533/src/host.rs#L123)):_
Done.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/138)_

SimonSapin:
@bors-servo r+

Thanks!

frewsxcv:
Worth mentioning that `IpAddr` in `std` is now deprecated

https://github.com/rust-lang/rust/pull/30187

SimonSapin:
This uses Ipv4Addr and Ipv6Addr (which are stable and not deprecated) separately in a custom enum (with a third Domain variant), not the deprecated IpvAddr enum.

frewsxcv:
Ah, thought the enum was used in here, carry on

PR #139, opened 2015-11-20
Move all tests to tests/ folder.

pyfisch:
The folder is provided by Cargo and makes it easier to find tests.

Question: I found that src/parser.rs contains many parsers, should I move `parse_ipv4addr` and `parse_ipv6addr` to the parsers file?

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/139)


SimonSapin:
I’m not exactly opposed to this change, but I don’t understand: what’s the point? Are they really hard to find where they are now? http://doc.crates.io/guide.html#tests describes the two places to have tests as "should be unit tests" and "should be integration tests" but the difference is not really clear to me.

`parser.rs` contains the implementation of https://url.spec.whatwg.org/#url-parsing. It’s in a separate file because it’s so big.

pyfisch:
My point is just that all other crates I know use the tests/ directory to for integration tests and I consider the tests using test files as integration tests.

SimonSapin:
Chatting on IRC, other people seem to agree with this change.

@bors-servo r+

PR #151, opened 2015-12-04
Make it possible to define new encode sets in other crates.

SimonSapin:
**Do not merge yet**

One one hand, this is technically a breaking change (for example if a dependent crate had a function with an `EncodeSet` parameter) so this should be version 0.6.0

On the other hand, it’s *probably* not breaking for *most* users, and updating lots of `Cargo.toml` files in many repositories to deal with `expected Url, found Url` type errors is a pain.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/151)


SimonSapin:
This was merged as part of #176.

PR #152, opened 2015-12-05
IDNA support

SimonSapin:
@valenting, I rebased your PR #119 and then worked on it some. Could you have a look?

Two things are known missing (and have FIXME comments):

* The check for combining marks
* “The Bidi Rule” http://tools.ietf.org/html/rfc5893#section-2

There are some tests that are supposed to return an error but currently don’t. (The corresponding `assert!` is commented out.) Some are because of the bidi rule, but I’m not sure about others. (For example `a\u200Cb`.)

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/152)


src/idna.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// http://www.unicode.org/reports/tr46/#Validity_Criteria
+fn validate(label: &str, flags: Uts46Flags) -> Result<(), Error> {
+    // Input is from nfc(), so it must be in NFC?
+    // Can not contain '.' since the input is from .split('.')
+    if {
+        let mut chars = label.chars();
+        let _first = chars.next();
+        let _first = chars.next();
```
pyfisch: should probably be `_second`?

src/idna.rs
```diff
@@ -0,0 +1,192 @@
...
+fn is_combining_mark(c: char) -> bool {
+    false  // FIXME General_Category=Mark
+}
+
+/// http://www.unicode.org/reports/tr46/#Validity_Criteria
+fn validate(label: &str, flags: Uts46Flags) -> Result<(), Error> {
+    // Input is from nfc(), so it must be in NFC?
+    // Can not contain '.' since the input is from .split('.')
+    if {
+        let mut chars = label.chars();
```
pyfisch: You can use `let mut chars = label.chars().skip(2);`

http://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip

PR #153, opened 2015-12-10
Introduce Url::join

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/153)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "0.5.1"
+version = "0.5.2"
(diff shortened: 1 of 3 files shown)


SimonSapin:
r=me with a version number bump

SimonSapin:
Sorry, I should have been more precise. This should be 0.5.2, not 0.6.0 since it’s a backward compatible addition. (And updating all the things for semver-incompatible versions is a pain.)

nox:
@SimonSapin Sorry myself, I know the rules, just forgot that semver with 0.y.z is different than with 1.y.z.

PR #155, opened 2015-12-26
Add doc comment for percent_encoding::from_hex

cmbrandenburg:
[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/155)


SimonSapin:
I don't think this was intended to be a public API, but oh well

PR #159, opened 2016-01-18
making uuid field on OpaqueOrigin public

Chandler:
by default, fields on tuple structs are private[1]. In this case, an instance of OpaqueOrigin is not very useful if the only field is private and there's no getter.

[1] https://github.com/rust-lang/rfcs/blob/master/text/0001-private-fields.md

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/159)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -204,7 +204,7 @@ pub struct Url {
 
 /// Opaque identifier for URLs that have file or other schemes
 #[derive(PartialEq, Eq, Clone, Debug)]
-pub struct OpaqueOrigin(Uuid);
+pub struct OpaqueOrigin(pub Uuid);
 
 #[cfg(feature="heap_size")]
 known_heap_size!(0, OpaqueOrigin);


Ms2ger:
The point is that you can only compare these for equality and change them. Exposing the internal representation is explicitly not something we want to allow.

Chandler:
@Ms2ger hmm, I was confused, now I'm more confused. If this value is private, what will print out in servo when you call `location.origin` ?

I assumed it would be the value of the uuid in the OpaqueOrigin but it seems that's not totally obvious.
https://github.com/whatwg/url/issues/21
https://github.com/jsdom/whatwg-url/issues/10

I'm working on adding origin to servo's location and url APIs. I guess I can just have it return "protocol://" like in chrome?

Chrome:
> new URL("accc://abcsdd.com/adf").origin
"accc://"

Ms2ger:
https://html.spec.whatwg.org/multipage/browsers.html#unicode-serialisation-of-an-origin
> If the origin in question is not a scheme/host/port tuple, then return the literal string "null" and abort these steps.

KiChjang:
@jdm has an open PR that works on the origin issue: servo/servo#8658

PR #169, opened 2016-02-13
Change return type of password_mut to Option<&mut Option<String>>

KiChjang:
This is necessary because by default, if the URL ':' delimiter was not found in the URL, then password_mut() would return None, causing a later mutation to the password field in RelativeSchemeData impossible.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/169)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "0.5.5"
+version = "0.6.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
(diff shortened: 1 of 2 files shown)


SimonSapin:
This is sensible, but I have an in-progress rewrite that makes it obsolete: https://github.com/servo/rust-url/compare/1.0 (The password and other components will not be independent `String`s anymore. `*_mut` methods will be replaced with setters, though I’m not certain of the exact API yet.)

So making a breaking release if another one is coming soon may not be worth it. In the mean time, as discussed on IRC, you can use the `Url::relative_scheme_data_mut()` method and access the `RelativeSchemeData::password` field directly.

KiChjang:
Ah ok, the rewrite makes more sense then.

Issue #166, opened 2016-02-03
Panic when parsing a `.` in file URLs


alexcrichton:
Parsing the url `file://./foo` will cause rust-url to panic, for example:

``` rust
let _url = url::Url::parse("file://./foo");
```

yields:

```
thread '<main>' panicked at 'a non-empty list of numbers', ../src/libcore/option.rs:335
stack backtrace:
   1:     0x7fda97eb0f40 - sys::backtrace::tracing::imp::write::haa19c02b4de52f3bG0t
   2:     0x7fda97eb3075 - panicking::log_panic::_<closure>::closure.41218
   3:     0x7fda97eb2af0 - panicking::log_panic::h527fe484e9de8fe1W7x
   4:     0x7fda97ead303 - sys_common::unwind::begin_unwind_inner::h51f64b1a34c60827fTs
   5:     0x7fda97ead418 - sys_common::unwind::begin_unwind_fmt::h0845853a1913f45blSs
   6:     0x7fda97eb05a1 - rust_begin_unwind
   7:     0x7fda97ede0cf - panicking::panic_fmt::h3967afc085fe8067LFK
   8:     0x7fda97e8ffb5 - option::_<impl>::expect::expect::h6499254404020416113
                        at ../src/libcore/macros.rs:29
   9:     0x7fda97e8c02d - host::parse_ipv4addr::h8bc077b4a80e7f06FIa
                        at src/host.rs:186
  10:     0x7fda97e886b8 - host::_<impl>::parse::hd99c67cae0091fc0Uya
                        at src/host.rs:61
  11:     0x7fda97e98e65 - parser::parse_file_host::ha2c614830ab90204nKb
                        at src/parser.rs:469
  12:     0x7fda97e926b0 - parser::parse_relative_url::h4e24ca8208217f655mb
                        at src/parser.rs:217
  13:     0x7fda97e85497 - parser::parse_url::hdcd76e516d93c3d01eb
                        at src/parser.rs:122
  14:     0x7fda97e80a2b - _<impl>::parse::h6996f44ab86e41c8w1p
                        at src/lib.rs:437
  15:     0x7fda97e809eb - _<impl>::parse::h6fc311753b000c53hlq
                        at src/lib.rs:566
  16:     0x7fda97e80998 - main::h13ef09143d6a2ad6faa
                        at src/main.rs:4
  17:     0x7fda97eb2894 - sys_common::unwind::try::try_fn::h11901883998771707766
[...]

PR #170, opened 2016-02-21
Do not panic when parsing '.' in URL host: #166

danlrobertson:
In response to #166. The parser should not panic when parsing URLs such as `file://./foo`.

Really this was just an excuse to dig into `rust-url` and learn more about it, so tips, comments, and critiques are welcomed!

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/170)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -58,6 +58,9 @@ impl Host {
         ][..]).is_some() {
             return Err(ParseError::InvalidDomainCharacter)
         }
+        if domain.is_empty() {
+            return Err(ParseError::RelativeUrlWithoutBase)
+        }
         match parse_ipv4addr(&domain[..]) {
             Ok(Some(ipv4addr)) => Ok(Host::Ipv4(ipv4addr)),
             Ok(None) => Ok(Host::Domain(domain.to_ascii_lowercase())),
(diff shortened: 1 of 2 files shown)


danlrobertson:
Done.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/host.rs, line 62 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/170#-KB16ST8l2iJmKjmxca_:-KB16ST8l2iJmKjmxcaa:-1490546468) ([raw file](https://github.com/servo/rust-url/blob/a269edfcc42291b19d7a101cfae472c975da17f4/src/host.rs#L62)):*
Couldn't find much in the spec about this, so the same error from `Url::from_directory_path` is returned.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/170#-:-KB16Rn8eWjrQ7cjhMlH:-499146009)*

SimonSapin:
Thanks for the PR! It turns out the underlying cause of the issue was a bit more subtle, so I’ll close this in favor of https://github.com/servo/rust-url/pull/171.

SimonSapin:
Specifically, a domain made of just a dot is turned into the empty string by IDNA code.

For the spec, see also https://github.com/whatwg/url/issues/79

danlrobertson:
Thanks! Great comment on the spec too!

PR #171, opened 2016-02-22
Don’t remove leading dots in domain names

### LINKED ISSUE #166: Panic when parsing a `.` in file URLs

SimonSapin:
Fixes #166, closes #170.

http://www.unicode.org/reports/tr46/#ToASCII has algorithms like:

> - Break the result [a domain] into labels at U+002E FULL STOP.
> - […]
> - join the labels using U+002E FULL STOP as a separator

Before this PR, the IDNA code does "join" with code like:

``` rust
     let mut result = String::new();
     for label in /* ... */ {
         if result.len() > 0 {
             result.push('.');
         }
         result.push_str(/* ... */)
     }
```

This writes a dot before every label, except the first. But this code seems to be buggy: if all labels so far are empty, it won’t push dots to separate empty labels. In other words, leading dots are stripped.

I’ve fixed this apparent bug by using a boolean, but a number of tests started failing. They apparently expect leading dots to be stripped. I don’t see any thing in the spec to support this behavior.

@valenting did you mean to strip leading dots in this code? Am I missing something in the spec?

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/171)


valenting:
I don't recall if my implementation happened to match the unit tests, or if I changed the implementation to match the unit tests.
But I do agree that the spec doesn't specify this, and your patch makes more sense.
We should raise this issue with the unicode group, to get the unit tests fixed.

valenting:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/servo/rust-url/171#-KBxxOwXdbMIO-8sTygL:-KBxxOwY8FnefgP6wCv9:-855967974) (no related file):*
Sorry I never completed this review. As I said, this makes sense, and I think we should land it.

---

*[tests/tests.rs, line 119 \[r1\]](https://reviewable.io:443/reviews/servo/rust-url/171#-KBxwuZEMx1D3KnVM96y:-KBxwuZEMx1D3KnVM96z:-18477990) ([raw file](https://github.com/servo/rust-url/blob/9d0fd21ab002e59479610166956b04e98c7d186f/tests/tests.rs#L119)):*
Interestingly, not how Firefox parses this URL :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-url/171)*

Manishearth:
Just realized that this is the same as https://github.com/servo/rust-url/pull/256 .

Both chrome and firefox do not strip any dots, and the spec agrees with this. We should land either this or #256

We're not running the IDNA tests on CI, btw, so we already have a bunch that fail.

Manishearth:
> But I do agree that the spec doesn't specify this

The spec says to "join using full stops", and the "join" operation usually means putting the given delimeter between the segments, even if they are empty. `["","",""].join('.') == ".."`

SimonSapin:
What bothers me about this change, and why this PR isn’t merged yet, is that it makes some IDNA tests fail. It looks like these tests might be wrong? But the Unicode Technical Committee has been completely unresponsive to reports of spec issues so far, so I don’t know what to do here.

Manishearth:
The tests are wrong -- they don't agree with the spec nor the implementations.

Manishearth:
Anyway, r=me on this change

dtolnay:
This is still waiting for a response from the Unicode Technical Committee right?

behnam:
I have submitted a new PR, based on this, which also has a fix for the conformance tests: https://github.com/servo/rust-url/pull/337

About the spec and test data issues, please see discussion on the issue: https://github.com/servo/rust-url/issues/166

With those, I recommend closing this old PR.

Hoverbear:
Closing this as requested.

PR #177, opened 2016-03-04
New IntoUrl trait

cmbrandenburg:
Here's a useful trait the `url` crate could provide, otherwise many other crates may duplicate the same functionality.

The `IntoUrl` trait allows a function to accept a URL argument that's either a `Url` instance or a string. For example:

```rust
fn foo<U: IntoUrl>(url: U);

foo("http://localhost/");
foo(Url::parse("http://localhost/").unwrap());
```

I didn't write this. This code comes from Hyper ([see code here](https://github.com/hyperium/hyper/blob/45ec6fd8841786b2625e7240507a967bd27a6050/src/client/mod.rs#L384-L406), [documentation here](http://hyper.rs/hyper/hyper/client/trait.IntoUrl.html)). I touched up the doc comments and added some unit tests.

My CouchDB crate wraps the same `IntoUrl` trait, so there are at least two crates out there providing this functionality. It would be helpful to consolidate this effort into the `url` crate.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-url/177)


nox:
Why don't you just do `fn foo<T>(input: Into<Url>)`?

sfackler:
@nox The conversion needs to be fallable.

SimonSapin:
I’ve read IRC talk about having fallible conversion traits in `std`. If that’s indeed happening, I’d rather implement that than add an ad-hoc trait.

Also, do you typically need to take ownership of an `Url`? If you only read from it, it’d be nice to somehow accept `&Url` too (without cloning).

cmbrandenburg:
@SimonSapin In my case, yes, I need to take ownership. I'm not sure about Hyper, though my guess is they have a good reason. In my case, the application constructs a client object that points to a CouchDB server via URL. The client instance _could_ be constrained by the lifetime of a `Url` stored elsewhere—hence your `&Url` idea—but it's more convenient for the application to store the `Url` within the client instance and not worry about lifetimes. This is especially poignant when there exist multiple simultaneous client instances.

Does anyone have any links to these forthcoming fallible conversion traits in `std`?

SimonSapin:
@sfackler is reportedly working on writing up an RFC, and has experiments at https://github.com/sfackler/rust/tree/try-from

cmbrandenburg:
@SimonSapin Thanks!

PR #183, opened 2016-04-04
Update uuid.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/183)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "0.5.7"
+version = "0.5.8"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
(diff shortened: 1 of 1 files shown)


Ryman:
FWIW, this now requires a later version of rust then previously (now requires >1.6) due to uuid 2.0 having `no_std` support, so I think bumping to 0.6 might have been nicer here.

SimonSapin:
Is there a reason to use older versions of Rust?

FWIW with #176 this crate will require Rust 1.7+.

Ryman:
Personally, I don't mind using latest stable, but I think it affects users who may only want to use a packaged version, or have an OS which isn't tier1 and has slower releases or something? It likely has more problems for end-users compiling from source than for developers of the library itself.

I'm not entirely sure what the community consensus on this is but it is a kind of breaking change, so some kind of signalling, via either version bump or opt-in feature, would be neat imo.

Requiring 1.7 for the 1.0 bump (#176) would be totally fine as it wouldn't break dependent crates unexpectedly, i.e. they have to opt in to Cargo letting them move from 0.6 -> 1.0 (unless they use `*`).

SimonSapin:
Alright, let’s reconsider this if it becomes more than hypothetical.

PR #191, opened 2016-04-26
Fixed AppVeyor link

perlun:
The actual badge URL is still wrong. I don't think this project is set up there since it was moved from Manishearth's acount to the servo organization, so maybe that needs fixing as well.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/191)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 rust-url
 ========
 
-[![Travis build Status](https://travis-ci.org/servo/rust-url.svg?branch=master)](https://travis-ci.org/servo/rust-url) [![Appveyor build status](https://ci.appveyor.com/api/projects/status/ulkqx2xcemyod6xa?svg=true)](https://ci.appveyor.com/project/Manishearth/rust-url)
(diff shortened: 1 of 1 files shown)


SimonSapin:
@Manishearth or @larsbergstrom, do you know how to get a correct URL?

untitaker:
You need to reregister the project, Appveyor is unable to handle GitHub renames.

SimonSapin:
This is not a github rename. @Manishearth first got AppVeyor set up for servo/rust-url. I believe it showed up as https://ci.appveyor.com/project/Manishearth/rust-url because AppVeyor doesn’t support github organizations well, or didn’t at the time.

untitaker:
Ah, sorry for the noise then.

larsbergstrom:
Can't really do that, AFAIK. AppVeyor chooses an account to associate the "project" with (in this case servo/rust-url) and then puts it in that account's name/GH-project. So, I can move it to the paid account, but then it'll just have larsbergstrom in it :-(

Hoverbear:
Closing this because the changed link is invalid while the existing link is valid. If this changes in the future we can either dig up this PR or just make the change.

Thanks so much for the contribution!

perlun:
@Hoverbear - that's fine. Apparently, this is caused by a bug at the AppVeyor side so let's live with it for now.

PR #194, opened 2016-05-02
Add Url::path_segments_mut, fix #188

### LINKED ISSUE #188: Manipulating path segments
`Url` currently has `pop_path_segments` and `push_path_segment` methods, but they have a design problem reported by @sfackler. A path is at minimum `/` which is considered like a single path segment that happens to be empty (like trailing slashes in general), so `push_path_segment("foo")` make the path be `//foo` with two slashes.

Fixing this may involve breaking API changes, so instead I’ll make these methods private (effectively removing them) before publishing 1.0.0, in order to unblock it. I’ll add them back (or something else to provide this functionality) in 1.1.0.

SimonSapin:
@sfackler @maximih @jgillich How does this look? Would these new API help for your use cases? Can they be improved?

@nox, please review and but don’t land yet.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/194)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "1.0.1"
+version = "1.1.0"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,12 +142,14 @@ use std::str;
 
 pub use origin::{Origin, OpaqueOrigin};
 pub use host::{Host, HostAndPort, SocketAddrs};
+pub use path_segments::PathSegmentsMut;
 pub use parser::ParseError;
 pub use slicing::Position;
 
 mod encoding;
 mod host;
 mod origin;
+mod path_segments;
 mod parser;
 mod slicing;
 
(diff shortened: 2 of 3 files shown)


sfackler:
This looks great!

maximih:
I think it looks good even though I find this a bit unexpected:

```rust
extern crate url;

fn main() {

    let mut my_url = url::Url::parse("http://this/is/a").unwrap();

    let new_url = my_url.join("./test").unwrap();
    println!("{}", new_url.as_str());

    my_url.path_segments_mut().unwrap().push("./test");
    println!("{}", my_url.as_str());
}
```

Which prints:

```
http://this/is/test
http://this/is/a/.%2Ftest
```

where I was expecting `http://this/is/a/test` so that it's consistent with `join()` implementation.

Anyhow, it's not a big deal as I can fi it by using `my_url.path_segments_mut().unwrap().push("test")`

So I am happy with the implementation. Thanks for your effort

SimonSapin:
It is by design that `PathSegmentsMut::push` escapes slashes and does not interpret `.` or `..` specially: it adds exactly one segment, whatever the characters in that segment. The point is not to do the same as `join`, you can use `join` for that.

Actually, I wonder what `push(".")` or `push("..")` should do. With the code is it currently is, `Url::parse(url.as_str()).unwrap() == url` would not hold, since `.` and `..` segments are normalized during parsing. @nox, what do you think?

nox:
> @nox, what do you think?

That's a good question. Maybe it should panic on `.` and `..`?

Btw, r=me for the rest.

SimonSapin:
I've added a commit to ignore . and make .. behave like pop()

SimonSapin:
@bors-servo r=nox

(IRC)

PR #199, opened 2016-05-27
Fixed deriving etc.

izgzhen:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/199)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -7,7 +7,7 @@
 // except according to those terms.
 
 use std::cmp;
-use std::fmt::{self, Formatter, Write};
+use std::fmt::{self, Formatter};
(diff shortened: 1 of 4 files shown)


SimonSapin:
Not deriving these impls is deliberate, in order to only serialize/parse a single string per URL instead of all the fields of the structs. These fields are implementation details that should not be exposed outside of the crate, and there are a lot of invariants that need to be maintained between them: things break if the fields are inconsistent.

izgzhen:
Ok I see, this makes sense. Then how about the `Origin` one?

SimonSapin:
I’ve replied there.

PR #200, opened 2016-05-27
Improve Origin

izgzhen:
First change is an addition of `same_origin`, which is more explicit than `==` reload I believe.

Second change is introduction of `same_origin_domain` and unavoidably, the optional domain field. Although spec is not very precise on what `domain` can do here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/200)

### DIFF
--- a/src/origin.rs
+++ b/src/origin.rs
@@ -24,7 +24,7 @@ pub fn url_origin(url: &Url) -> Origin {
         },
         "ftp" | "gopher" | "http" | "https" | "ws" | "wss" => {
             Origin::Tuple(scheme.to_owned(), url.host().unwrap().to_owned(),
-                url.port_or_known_default().unwrap())
+                          url.port_or_known_default().unwrap(), None)
         },
         // TODO: Figure out what to do if the scheme is a file
         "file" => Origin::new_opaque(),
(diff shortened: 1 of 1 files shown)


src/origin.rs
```diff
@@ -39,8 +39,9 @@ pub enum Origin {
     /// A globally unique identifier
     Opaque(OpaqueOrigin),
 
-    /// Consists of the URL's scheme, host and port
-    Tuple(String, Host<String>, u16)
+    /// Consists of the URL's scheme, host and port, and an optional domain
+    /// https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-tuple
```
Manishearth: I suspect the html spec origin might diverge from the URL spec origin, this crate is URL spec only

src/origin.rs
```diff
@@ -91,6 +92,27 @@ impl Origin {
             }
         }
     }
+
+    /// https://html.spec.whatwg.org/multipage/browsers.html#same-origin
+    pub fn same_origin(&self, other: &Origin) -> bool {
+        self == other
+    }
+
+    /// https://html.spec.whatwg.org/multipage/browsers.html#same-origin
```
Manishearth: https://html.spec.whatwg.org/multipage/browsers.html#same-origin-domain

Ms2ger:
I don't think we should add the domain bit, since we're not planning on using it any time soon in Servo.

izgzhen:
Ok cool, so let's close it for now? or hang it for a while

Manishearth:
Yeah, we only need == in servo for now. If we need the domain thing we can revisit this or wrap it in servo

SimonSapin:
Even when we do implement the domain bit in Servo, I’d be in favor of having it in Servo’s `script::origin::Origin` wrapper type (or similar) rather than in rust-url.

izgzhen:
Good, I will close this for now.

PR #201, opened 2016-05-27
Changes needed for gecko-integration

valenting:
This pull request contains 2 features that are needed in the C-API for use in Gecko:
1. Feature that allows to only percent decode characters that are valid in the hostname. 
This introduces the ```only_percent_decode_hostname_valid``` feature.
2. From<u32> for Position in order to use slicing in the C API.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/201)

### DIFF
--- a/src/percent_encoding.rs
+++ b/src/percent_encoding.rs
@@ -110,6 +110,14 @@ define_encode_set! {
     }
 }
 
+define_encode_set! {
+    /// This encode set is used to decide which characters are allowed in the hostname.
+    pub HOSTNAME_ENCODE_SET = [SIMPLE_ENCODE_SET] | {
+        ' ', '!', '"', '#', '$', '%', '&', '\'', '(', ')', '+', ',', '/', ':', ';',
+        '<', '=', '>', '?', '@', '[', '\\', ']', '^', '`', '{', '|', '}', '~'
+    }
+}
+
 /// Return the percent-encoding of the given bytes.
 ///
 /// This is unconditional, unlike `percent_encode()` which uses an encode set.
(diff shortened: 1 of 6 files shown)


SimonSapin:
I’ll comment more tomorrow, but why the first change? It deviates from the spec. If it’s necessary for web-compat, the spec should be changed.

valenting:
I forgot to mention the reasoning behind this. Indeed it deviates from the spec (which is why I made it a feature), but that is needed to support old addons in Gecko, which access their content using a URL such as resource://ADDON_ID/path - where ADDON_ID usually contains a percent encoded ```@``` character.
If we feel this isn't a change we want to support at all, I could use a separate fork for the Gecko stuff.

SimonSapin:
Does Gecko’s URL parser for web content have the same behavior? If that’s the case and Gecko is not willing to align with the spec, maybe the spec should be changed. Could you file an issue on https://github.com/whatwg/url/issues/ with the proposed parser changes and this reasoning?

Instead of a compile-time switch it could be a per-URL run-time one: add a method to `ParseOptions` and use it like: `Url::options().only_percent_decode_hostname_valid().parse(some_str)`. What do you think?

I’ll comment inline on implementation details.

src/percent_encoding.rs
```diff
@@ -110,6 +110,14 @@ define_encode_set! {
     }
 }
 
+define_encode_set! {
+    /// This encode set is used to decide which characters are allowed in the hostname.
+    pub HOSTNAME_ENCODE_SET = [SIMPLE_ENCODE_SET] | {
```
SimonSapin: Encode sets are normally used for percent encoding, not decoding. So I’d rather not expose this in the public API. In `after_percent_sign` you could "inline" `!HOSTNAME_ENCODE_SET.contains(c)` to `!(matches(c, ' ' | '!' | '"' | /* ...*/ '~') || c < '\u{20}' || c > '\u{7E}')`, or invert the logic and use `matches!(c, 'a'...'z' | 'A'...'Z' | '0'...'9' | '.' | '-' | '_' | '*')`. (I think that’s the complementary set, but I’m not sure `*` should be there.)

src/slicing.rs
```diff
@@ -77,6 +78,7 @@ impl Index<Range<Position>> for Url {
 /// `BeforeScheme` and `AfterFragment` are always the start and end of the entire URL,
 /// so `&url[BeforeScheme..X]` is the same as `&url[..X]`
 /// and `&url[X..AfterFragment]` is the same as `&url[X..]`.
+#[repr(u32)]
```
SimonSapin: Why `u32`? `u8` is more than large enough for 16 variants.

src/slicing.rs
```diff
@@ -97,6 +99,14 @@ pub enum Position {
     AfterFragment
 }
 
+impl From<u32> for Position {
+    fn from(f: u32) -> Self {
+        unsafe {
+            mem::transmute(f)
```
SimonSapin: I think this `transmute` is unsound. This method itself is not `unsafe`, and there’s nothing stopping users from calling it with a value that doesn’t have a corresponding enum variant. This should return an option, and avoid transmute. That code would be tedious and error-prone to write by hand, but it can be generated with a macro:

```rust
macro_rules! c_like_enum {
    ( $type_name: ident: $( $variant: ident, )+ ) => {
        c_like_enum! {
            $type_name
            [ $($variant)* ]
            [ ]
            next = 0
        }
    };

    (
        $type_name: ident
        [ $next_variant: ident $( $variant: ident )* ]
        [ $($assigned_variant: ident = $value: expr),* ]
        next = $next_value: expr
    ) => {
        c_like_enum! {
            $type_name
            [ $($variant)* ]
            [ $($assigned_variant = $value,)* $next_variant = $next_value ]
            next = $next_value + 1
        }
    };

    (
        $type_name: ident
        [ ]
        [ $($variant: ident = $value: expr),+ ]
        next = $next_value: expr
    ) => {
        #[derive(Copy, Clone, Debug)]
        #[repr(u8)]
        pub enum $type_name {
            $($variant = $value,)+
        }

        impl $type_name {
            #[allow(non_upper_case_globals)]
            pub fn from_u8(value: u8) -> Option<Self> {
                $(
                    const $variant: u8 = $value;
                )+
                match value {
                    $(
                        $variant => Some($type_name::$variant),
                    )+
                    _ => None
                }
            }
        }
    };
}

c_like_enum! { Position:
    BeforeScheme,
    AfterScheme,
    BeforeUsername,
    AfterUsername,
    BeforePassword,
    AfterPassword,
    BeforeHost,
    AfterHost,
    BeforePort,
    AfterPort,
    BeforePath,
    AfterPath,
    BeforeQuery,
    AfterQuery,
    BeforeFragment,
    AfterFragment,
}

nox:
Why not use an enum from C too for Position?

valenting:
The resource://ADDON_ID issue will be fixed in Firefox, so we shouldn't worry about that here.
I'll probably open another PR for the slicing bits.

nox:
> The resource://ADDON_ID issue will be fixed in Firefox,

@valenting And you deliver this information on my birthday? You have a good timing sense, thanks. :D

PR #205, opened 2016-06-25
Document common changes needed during a 0.x -> 1.x upgrade

carols10cents:
Hi! I had to upgrade another crate from 0.x to 1.x, so I decided to actually write up what I did. Then I went and looked at the crates that @SimonSapin upgraded and found some more. I also spot checked a few reverse dependency crates that have done this upgrade. So I'm sure there are more changes that I'm missing, but these should be the most common ones.

There are still quite a few crates that are using url 0.x listed on crates.io, I hope this will help when they decide to upgrade! 🎉

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/205)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,5 @@ rust-url
 URL library for Rust, based on the [URL Standard](https://url.spec.whatwg.org/).
 
 [Documentation](https://servo.github.io/rust-url/url/index.html)
+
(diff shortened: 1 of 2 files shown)


carols10cents:
Closing in favor of #206!

SimonSapin:
Since I used your commit as-is let’s let github mark it as merged :)

carols10cents:
Ok! :)

PR #216, opened 2016-07-30
Allow serde 0.8 and remove use of compiler plugins

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/216)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,12 +27,11 @@ rustc-serialize = "0.3"
 
 [features]
 query_encoding = ["encoding"]
-heap_size = ["heapsize", "heapsize_plugin"]
+heap_size = ["heapsize"]
 
 [dependencies]
 encoding = {version = "0.2", optional = true}
 heapsize = {version = ">=0.1.1, <0.4", optional = true}
-heapsize_plugin = {version = "0.1.0", optional = true}
 idna = { version = "0.1.0", path = "./idna" }
 matches = "0.1"
 rustc-serialize = {version = "0.3", optional = true}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "1.1.1"
+version = "1.2.0"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,10 +30,10 @@ query_encoding = ["encoding"]
 heap_size = ["heapsize", "heapsize_plugin"]
 
 [dependencies]
-idna = { version = "0.1.0", path = "./idna" }
+encoding = {version = "0.2", optional = true}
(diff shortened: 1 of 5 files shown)


nox:
The serde feature is actually used outside of Servo, and it would break at least [feeds-to-pocket](https://github.com/FraGag/feeds-to-pocket).

nox:
This would also break [wolfram-alpha-rs](https://github.com/indiv0/wolfram-alpha-rs) and [xkcd-rs](https://github.com/indiv0/xkcd-rs).

@FraGag and @indiv0: Would you mind having to relax the serde bound in your own crates at the same time as we merge this here, so that we avoid a 2.0.0 bump in rust-url?

FraGag:
Well this is annoying. If I relax the requirement on serde in feeds-to-pocket but stay on url 1.1.1, then feeds-to-pocket picks up serde 0.8.x while url stays on serde 0.7.x, which causes build errors. On the other hand, if I don't relax the requirement on serde in feeds-to-picket but update url to your branch, then it's the other way around: feeds-to-pocket stays on 0.7.x while url picks up 0.8.x, which also causes build errors. If I update *both at once*, then it's fine. :\ If https://github.com/rust-lang/cargo/issues/2064 was fixed, then this wouldn't be an issue.

feeds-to-pocket is an executable, not a library, so the package comes with its Cargo.lock, and unless a user runs `cargo update`, in theory the crate should still build, because my Cargo.lock currently references url 1.1.1. Still, I'll publish an update with relaxed requirements after url 1.2.0 is published.

nox:
@FraGag Given feeds-to-pocket is an executable, there is no need for relaxed requirements there, even if you update rust-url after this is published, it won't make it use serde 0.8 as long as you don't also update serde AFAIK.

SimonSapin:
Even though publishing this change can break projects that don’t have a `Cargo.lock` file because Cargo over-optimitic in using the latest possible version of serde, as far as I understand it is possible then to downgrade with something like `cargo update -p serde --precise 0.7.15`.

As such, I believe this does not qualify as a breaking API change, and so 1.2.0 rather than 2.0.0 is fine. (Or even 1.1.2, but let’s not split hairs about this, Cargo treats them the same.)

@bors-servo r+

FraGag:
Just as a heads up (in particular to @nox): I just reinstalled feeds-to-pocket v0.1.1 locally with `cargo install` and now it fails to build. Apparently, crates uploaded to crates.io never contain Cargo.lock, even if it's not in the project's .gitignore. As a result, Cargo had to recalculate the dependencies and selected Serde 0.7 for feeds-to-pocket and Serde 0.8 for rust-url, which causes build errors because `url::Url` implements Serde 0.8's `serde::Serialize` trait, whereas feeds-to-pocket needs it to implement Serde 0.7's `serde::Serialize` trait.

So this did end up being a breaking change, but I agree with @SimonSapin that this kind of change should not be treated as a breaking change. Rather, Cargo could and should be improved such that situations like this do not cause breakage in the future (see e.g. https://github.com/rust-lang/cargo/issues/2064, which I also mentioned above).

PR #217, opened 2016-08-04
doc: replace bad grammar

tshepang:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/217)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -368,7 +368,7 @@ impl Url {
 
     /// Return the origin of this URL (https://url.spec.whatwg.org/#origin)
     ///
-    /// Note: this return an opaque origin for `file:` URLs, which causes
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -368,7 +368,7 @@ impl Url {
 
     /// Return the origin of this URL (https://url.spec.whatwg.org/#origin)
     ///
-    /// Note: this return an opaque origin for `file:` URLs, which causes
+    /// Note: the return value is an opaque origin for `file:` URLs, which causes
```
KiChjang: Equivalently, "this returns" would fix all of the problems here. Your new wording makes the "which causes..." part sound very strange since it isn't clear on what the "which" is referring to.

tshepang:
fixed https://github.com/servo/rust-url/pull/218

PR #223, opened 2016-09-01
Truncate the serialization in Url::take_after_path

### LINKED ISSUE #222: Pushing empty path segment results in doubled query parameters
A testcase:

```diff
diff --git a/tests/unit.rs b/tests/unit.rs
index 55906b8..676f2a3 100644
--- a/tests/unit.rs
+++ b/tests/unit.rs
@@ -75,6 +75,14 @@ fn new_path_windows_fun() {
 
 
 #[test]
+fn append_trailing_slash() {
+    let mut url: Url = "http://localhost:6767/foo/bar?a=b".parse().unwrap();
+    url.path_segments_mut().unwrap().push("");
+    assert_eq!(url.to_string(), "http://localhost:6767/foo/bar/?a=b");
+}
+
+
+#[test]
 fn new_directory_paths() {
	 if cfg!(unix) {
		 assert_eq!(Url::from_directory_path(Path::new("relative")), Err(()));
```

Test results (manually formatted):

	thread '.append_trailing_slash' panicked at 'assertion failed: `(left == right)` (
		left: `"http://localhost:6767/foo/bar?a=b/?a=b"`,
		right: `"http://localhost:6767/foo/bar/?a=b"`
	)', tests/unit.rs:81
	note: Run with `RUST_BACKTRACE=1` for a backtrace.

emilio:
Fixes #222

I did this under the assumption that extending with an empty segment is a no-op. If we want it to cause something like `foo.com/bar//baz`, then the bit in `path_segments_mut` needs to be removed.

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/223)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -791,7 +791,11 @@ impl Url {
 
     fn take_after_path(&mut self) -> (u32, String) {
         match (self.query_start, self.fragment_start) {
-            (Some(i), _) | (None, Some(i)) => (i, self.slice(i..).to_owned()),
+            (Some(i), _) | (None, Some(i)) => {
+                let after_path = self.slice(i..).to_owned();
+                self.serialization.truncate(i as usize);
+                (i, after_path)
+            },
             (None, None) => (to_u32(self.serialization.len()).unwrap(), String::new())
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -326,7 +326,7 @@ impl Url {
             } else {
                 assert_eq!(self.byte_at(self.host_end), b':');
                 let port_str = self.slice(self.host_end + 1..self.path_start);
-                assert_eq!(self.port, Some(port_str.parse::<u16>().unwrap()));
+                assert_eq!(self.port, Some(port_str.parse::<u16>().expect("Couldn't parse port?")));
             }
             assert_eq!(self.byte_at(self.path_start), b'/');
         } else {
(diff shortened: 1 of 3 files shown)


untitaker:
Empty path segments are used for the eventual trailing slash in paths.

emilio:
@untitaker: Fair enough, removed that bit.

untitaker:
Also the testcase I provided in #222 still appears to fail :)

Really apprechiate your work btw, I never got around to do this myself but
wanted to.

emilio:
@untitaker: That's because that `is_empty()` bit, and because I'm stupid and forgot to push :)

untitaker:
Hmm the AppVeyor build fails!

untitaker:
(also it probably would make sense to add my testcase to this PR)

Phrohdoh:
This also fixes #227.

emilio:
@Hoverbear: The AppVeyor build fails at parsing the first URL in that test (`/`), because `path_to_file_url_segments_windows` expects a disk prefix or otherwise returns an error.

I "fixed" it by conditionally compiling that test. I'm pretty sure that failure has nothing to do with these changes (I doubt it passed CI in the first place?).

emilio:
Yeah, indeed [all the closed PRs](https://github.com/servo/rust-url/pulls?q=is%3Apr+is%3Aclosed) since the addition of that test fail on AppVeyor.

PR #231, opened 2016-10-20
Teach to_file_path to understand percent-encoded drive letters on Windows.

### LINKED ISSUE #230: to_file_path on Windows doesn't handle percent-encoded drive letter
Version 1.2.1

```rust
let succeeds = Url::parse("file:///c:/foo/bar").unwrap();
println!("{:?}", succeeds.to_file_path()); // Ok("c:\\foo\\bar")

let fails = Url::parse("file:///c%3a/foo/bar").unwrap();
println!("{:?}", fails.to_file_path()); // Err(())
```

Arnavion:
Fixes #230

---

I thought about using `percent_encoded` instead of matching `%3a` bytes individually, but I figured it would be overkill. If you think otherwise I'm happy to change it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/231)


SimonSapin:
Converting to/from file paths is not covered by https://url.spec.whatwg.org/ , so it’s probably fine to tweak that behavior.

Looks good, thanks!

@bors-servo r+

PR #237, opened 2016-10-30
Fix support for for IPv4-in-IPv6 syntax.

SimonSapin:
r? @Manishearth

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/237)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "1.2.2"
+version = "1.2.3"
 authors = ["The rust-url developers"]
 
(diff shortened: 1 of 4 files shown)


Manishearth:
Not sure if I've understood the code properly, might need to look at the context more. r=me though, will go through in depth later.

Could you also bump the version?

Manishearth:
ah, I understand now. r=me with or without the version bump

SimonSapin:
Also importing new tests from WPT, which I assume have been reviewed upstream.

@bors-servo r=Manishearth

PR #239, opened 2016-11-10
Add a few doc examples.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/239)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -930,6 +930,20 @@ impl Url {
     ///
     /// If this URL is cannot-be-a-base, does not have a host, or has the `file` scheme;
     /// do nothing and return `Err`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use url::Url;
+    ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -234,6 +234,16 @@ impl Url {
...
     /// Parse a string as an URL, with this URL as the base URL.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let url = Url::parse("https://example.net").unwrap();
+    /// let url = url.join("foo").unwrap();
+    /// assert_eq!(url.as_str(), "https://example.net/foo");
```
SimonSapin: Please change this to showcase trailing slashes: `https://example.net/foo` joined with `bar` is `https://example.net/bar`, but `https://example.net/foo/` joined with `bar` is `https://example.net/foo/bar`

r=me with this.

Issue #243, opened 2016-12-02
Panic when calling set_host with None on URL


frewsxcv:
```
let mut url = Url::parse("https://example.net/hello").unwrap();
url.set_host(None).unwrap();
```

```
thread 'main' panicked at 'begin <= end (19 <= 6) when slicing `https://example.net/hello`', ../src/libcore/str/mod.rs:1719
stack backtrace:
   1:        0x10d66ffaa - std::sys::imp::backtrace::tracing::imp::write::h383d0c200a1dd210
   2:        0x10d67fd6f - std::panicking::default_hook::{{closure}}::h9445b5c503f9824b
   3:        0x10d67cf2f - std::panicking::default_hook::h36fefcc1f50aa494
   4:        0x10d67d566 - std::panicking::rust_panic_with_hook::hf54dd023dee7ef13
   5:        0x10d67d404 - std::panicking::begin_panic::ha78ff90bed4768ad
   6:        0x10d67d322 - std::panicking::begin_panic_fmt::hc7561fdea0f5f71c
   7:        0x10d67d287 - rust_begin_unwind
   8:        0x10d6bfc10 - core::panicking::panic_fmt::hb7aed24f08b1bad6
   9:        0x10d6c0ae5 - core::str::slice_error_fail::hc148aa73da59a300
  10:        0x10d4d3314 - core::str::traits::<impl core::ops::Index<core::ops::Range<usize>> for str>::index::hc91b0c453a2b9a37
  11:        0x10d4c9474 - <collections::string::String as core::ops::Index<core::ops::Range<usize>>>::index::hfbbcd72b44dfd783
  12:        0x10d4caaba - collections::string::String::drain::h1868f57c1d4390c5
  13:        0x10d4ee1e2 - url::Url::set_host::hcdae4ea78ed9facd
  14:        0x10d4c8ca9 - rust_out::main::ha208b69ccbc11839
  15:        0x10d68032a - __rust_maybe_catch_panic
  16:        0x10d67ddf6 - std::rt::lang_start::h26c24604fceba8d3
  17:        0x10d4c8d89 - main
', ../src/librustdoc/test.rs:315
stack backtrace:
   1:        0x113674faa - std::sys::imp::backtrace::tracing::imp::write::h383d0c200a1dd210
   2:        0x113684d6f - std::panicking::default_hook::{{closure}}::h9445b5c503f9824b
   3:        0x113681e35 - std::panicking::default_hook::h36fefcc1f50aa494
[...]

PR #244, opened 2016-12-02
Prevent panic when calling `url::Url::set_host` with `None`.

frewsxcv:
Fixes https://github.com/servo/rust-url/issues/243.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/244)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -998,12 +998,18 @@ impl Url {
         }
 
         if let Some(host) = host {
+            if host == "" && SchemeType::from(self.scheme()).is_special() {
+                return Err(ParseError::EmptyHost);
+            }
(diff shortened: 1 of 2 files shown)


valenting:
Note this line from the [URL spec](https://url.spec.whatwg.org/#hostname-state):
> If url is special and buffer is the empty string, return failure. 

So both set_host(None) and set_host("") should probably fail if SchemeType::from(self.scheme()).is_special()

So while the test should not crash, the result should not be https:/hello
The same test would be correct though, if the protocol were not special.

frewsxcv:
@SimonSapin Do you remember why there is logic here for the 'cannot-be-a-base-URL' flag? At least for the link that @valenting sent, it doesn't look like that logic needs to be here.

PR #249, opened 2016-12-11
Add Redox support

jackpot51:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/249)


SimonSapin:
I don't know redox, but this looks reasonable.

@bors-servo r+

PR #252, opened 2016-12-17
Add faster serde serialization, sacrificing invariant-safety for speed in release mode; test that it roudtrips

Manishearth:
r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/252)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -16,52 +16,10 @@ use parser::{ParseResult, ParseError};
 use percent_encoding::percent_decode;
 use idna;
 
-#[cfg(feature="serde")]
-use serde;
-
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum HostInternal {
-    None,
-    Domain,
-    Ipv4(Ipv4Addr),
-    Ipv6(Ipv6Addr),
-}
+include!("codegen/host.rs");
 
 #[cfg(feature = "heapsize")]
 known_heap_size!(0, HostInternal);
-
-#[cfg(feature="serde")]
-impl serde::Serialize for HostInternal {
-    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
-        match *self {
-            HostInternal::None => 0u8.serialize(serializer),
-            HostInternal::Domain => 1u8.serialize(serializer),
-            HostInternal::Ipv4(addr) => {
-                2u8.serialize(serializer)?;
-                addr.serialize(serializer)
-            }
-            HostInternal::Ipv6(addr) => {
-                3u8.serialize(serializer)?;
-                addr.serialize(serializer)
-            }
-        }
-    }
-}
-
-#[cfg(feature="serde")]
-impl serde::Deserialize for HostInternal {
-    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error> where D: serde::Deserializer {
-        use serde::{Deserialize, Error};
-        let discr: u8 = Deserialize::deserialize(deserializer)?;
-        match discr {
-            0 => Ok(HostInternal::None),
-            1 => Ok(HostInternal::Domain),
-            2 => Ok(HostInternal::Ipv4(Deserialize::deserialize(deserializer)?)),
-            3 => Ok(HostInternal::Ipv6(Deserialize::deserialize(deserializer)?)),
-            _ => Err(D::Error::unknown_variant("Unknown variant")),
-        }
-    }
-}
 
 impl<S> From<Host<S>> for HostInternal {
     fn from(host: Host<S>) -> HostInternal {
--- a/src/host.rs
+++ b/src/host.rs
@@ -16,6 +16,9 @@ use parser::{ParseResult, ParseError};
 use percent_encoding::percent_decode;
 use idna;
 
+#[cfg(feature="serde")]
+use serde;
+
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum HostInternal {
     None,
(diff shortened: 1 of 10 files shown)


Manishearth:
r? @erickt on the custom serialization

I didn't use the `serialize_struct` and `serialize_struct_elt` stuff because this isn't going to be used for json serialization. Should I have?

erickt:
cc @dtolnay

Don't do this, it'll break formats like json that need to wrap aggregates in `{...}` or `[..]`. Instead, if you want zero overhead with [bincode](https://github.com/TyOverby/bincode/blob/master/src/serde/writer.rs#L200), serialize into a tuple, as in:

```
fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
    (&self.serialization, schema_end, ...).serialize(serializer)
}
```

We got a [ticket](https://github.com/serde-rs/serde/issues/637) open to write a container-level attribute that'll do this for you.

src/host.rs
```diff
@@ -27,6 +30,39 @@ pub enum HostInternal {
...
 known_heap_size!(0, HostInternal);
 
+#[cfg(feature="serde")]
+impl serde::Serialize for HostInternal {
+    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
+        match *self {
+            HostInternal::None => 0u8.serialize(serializer),
+            HostInternal::Domain => 1u8.serialize(serializer),
+            HostInternal::Ipv4(addr) => {
+                2u8.serialize(serializer)?;
```
erickt: would be better as (2u8, addr).serialize(serializer)

src/host.rs
```diff
@@ -27,6 +30,39 @@ pub enum HostInternal {
...
+                addr.serialize(serializer)
+            }
+            HostInternal::Ipv6(addr) => {
+                3u8.serialize(serializer)?;
```
erickt: same here

erickt:
Actually, `serde_macros` + bincode would generate the same code here as what you manually wrote, and would be safe to use with multiple formats, so I recommend just using that if possible.

Manishearth:
Picked up the code generator from #254, and added a travis check to ensure that its in sync.


r? @SimonSapin

erickt:
FYI, once we implement https://github.com/serde-rs/serde/issues/642, you should be able to replace your custom serde impls for `Url` with this hook.

nox:
@Manishearth I'm very much against that patch. No invalid URL should be produced. Why was that needed?

Manishearth:
Split it out as an extra pair of methods

nox:
Why must we put into the repos generated code?

Manishearth:
Need it to work on stable and don't want heavy dependencies on aster/syntex. It's temporary till macros 1.1 comes out.

Trying to keep the dependency list small for now.

SimonSapin:
> Need it to work on stable and don't want heavy dependencies on aster/syntex. It's temporary till macros 1.1 comes out.

Doesn’t serde with macros 1.1 use aster and syntax?

Manishearth:
No, it uses syn, and doesn't need build scripts.

nox:
serde can use syn with a build script on stable without needing syntex. I would rather we do that instead of checking in generated code.

src/lib.rs
```diff
@@ -1502,10 +1510,61 @@ impl rustc_serialize::Decodable for Url {
     }
 }
 
+#[cfg(feature="serde")]
+impl Url {
+    /// Serialize the URL efficiently, for use with IPC
+    pub fn serialize_efficient<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
```
nox: Please don't name if `efficient`. Its main property is that it is risky, not efficient.

src/lib.rs
```diff
@@ -1502,10 +1510,61 @@ impl rustc_serialize::Decodable for Url {
...
+            username_end: username_end,
+            host_start: host_start,
+            host_end: host_end,
+            host: host,
+            port: port,
+            path_start: path_start,
+            query_start: query_start,
+            fragment_start: fragment_start
+        };
+        #[cfg(debug_assertions)] {
```
nox: Without `debug_assertions`, type errors in that block won't be emitted.  Do `if cfg!(debug_assertions)` instead.

Manishearth:
> serde can use syn with a build script on stable without needing syntex. 

Fair. I also want to avoid a build script. And I want to avoid new dependencies for as long as possible, until https://bugzilla.mozilla.org/show_bug.cgi?id=1322798 is resolved. Right now the state of rust-url dependencies is acceptable for Gecko, but until that issue is resolved I'd prefer to wait before adding extra dependencies that can be avoided.

src/lib.rs
Manishearth: Renamed to _unsafe

nox:
> Fair. I also want to avoid a build script. And I want to avoid new dependencies for as long as possible, until https://bugzilla.mozilla.org/show_bug.cgi?id=1322798 is resolved. Right now the state of rust-url dependencies is acceptable for Gecko, but until that issue is resolved I'd prefer to wait before adding extra dependencies that can be avoided.

I would rather we make things more complicated on the Gecko side than on the Rust side for every contributor stumbling upon generated code.

Manishearth:
The only generated code is for the host struct, which isn't modified very often.

Manishearth:
And if we use syn+build script on this side you still have to deal with generated code. And debugging gets harder to do.

SimonSapin:
Closing in favor of #259, which is largely based on this.

PR #254, opened 2016-12-17
Add an external Serde code generator

erickt:
This builds upon #252 to use an external serde code generator to avoid forcing downstream components to use `serde_codegen`. If `Url` and `Host` ever need to be modified, just update `src/codegen/url.rs.in` and `src/codegen/host.rs.in`, then run `cd serde-code-generator && cargo run`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/254)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,8 @@ test = false
 [dev-dependencies]
 rustc-test = "0.1"
 rustc-serialize = "0.3"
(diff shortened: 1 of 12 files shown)


SimonSapin:
Thanks Erick! That’s an interesting pattern.

Closing as #259 removes the need for this, though.

PR #256, opened 2016-12-18
Handle empty domain segments correctly in idna

### LINKED ISSUE #255: "https://../foo" should not lose the dots when parsed
Parsing "https://../foo" gets us "https:///foo". This doesn't seem correct, and no browser has this behavior (they all produce "https://../foo")

cc @SimonSapin

Manishearth:
Spec asks us to "Otherwise join the labels using U+002E FULL STOP as a separator". The current code checking if the result was non-empty would fail if the segments themself were non-empty. It should be equivalent to `.split('.').map(.. processing ..).join('.')`. Implementing it procedurally since the code is already pretty procedural and it avoids extra allocations.

Should `processing()` return an iterator? We seem to be needlessly deconstructing and reconstructing the same string twice. OTOH that would mean more allocations.

r? @SimonSapin

fixes #255

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/256)

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -224,10 +224,12 @@ fn processing(domain: &str, flags: Flags, errors: &mut Vec<Error>) -> String {
     }
     let normalized: String = mapped.nfc().collect();
     let mut validated = String::new();
+    let mut first = true;
     for label in normalized.split('.') {
-        if validated.len() > 0 {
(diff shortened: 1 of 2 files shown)


SimonSapin:
Closing in favor of #171, I’ll comment there.

PR #259, opened 2016-12-19
Add (de)serialization of internal representation, to avoid re-parsing

SimonSapin:
+ random drive-by fixes.

The trick used here to avoid a complex `Deserialize` impl (either hand-written on in a cumbersome generated file) is to use existing impls, namely those of `Option` and `IpAddr`.

@Manishearth, another change from your PR is checking invariants unconditionally (including release mode) so there there is no need for separate `(de)serialize_unsafe` methods. Does this run-time cost sound acceptable?

r? @Manishearth

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/259)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ test = false
 [dev-dependencies]
 rustc-test = "0.1"
 rustc-serialize = "0.3"
+serde_json = ">=0.6.1, <0.9"
 
 [features]
 query_encoding = ["encoding"]
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 test:
-	cargo test --features "query_encoding serde rustc-serialize"
-	[ x$$TRAVIS_RUST_VERSION != xnightly ] || cargo test --features heapsize
+	cargo test --features "query_encoding serde rustc-serialize heapsize"
 	(cd idna && cargo test)
 
 .PHONY: test
--- a/src/host.rs
+++ b/src/host.rs
@@ -27,6 +27,38 @@ pub enum HostInternal {
 #[cfg(feature = "heapsize")]
 known_heap_size!(0, HostInternal);
 
+#[cfg(feature="serde")]
+impl ::serde::Serialize for HostInternal {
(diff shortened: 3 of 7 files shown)


erickt:
Looks good to me from the serde angle.

Manishearth:
>  so there there is no need for separate (de)serialize_unsafe methods. Does this run-time cost sound acceptable?


@nox disagrees. I am okay with this situation, sort of.

Manishearth:
r+ otherwise. Good idea.

nox:
I asked for separate methods because otherwise rust-url becomes pretty much useless to deserialize from JSON or TOML in the wild, where no one cares about the internal representation in rust-url.

SimonSapin:
IMO `serde::Serialize` and `serde::Deserialize` should be used together. By "in the wild", I assume you mean a situation whene one of the two ends is not serde. In these situations, `Url::as_str` and `Url::parse` should be used instead.

nox:
What do you mean? Many people use serde to serialise configuration and whatnot. Our serde support ought to serialise as values which make sense in most contexts of serde. The internal representation isn't one of them. What about formats which don't even have a serialiser?

Also, this is a breaking change.

SimonSapin:
Alright, I’ve changed the PR to leave the existing impls alone and add new inherent methods instead. r? @nox

Cargo.toml
```diff
@@ -38,4 +39,4 @@ heapsize = {version = ">=0.1.1, <0.4", optional = true}
 idna = { version = "0.1.0", path = "./idna" }
 matches = "0.1"
 rustc-serialize = {version = "0.3", optional = true}
-serde = {version = ">=0.6.1, <0.9", optional = true}
+serde = {version = "0.8.20", optional = true}
```
SimonSapin: nox correctly points out that this is a breaking change. @Manishearth why this change? Can we do without it?

Manishearth: Because bincode needs 0.8.20 and cargo isn't smart enough to coalesce the versions in a test-only dependency.

SimonSapin: Can we use JSON instead for the test and revert this?

Manishearth: Go ahead

PR #264, opened 2016-12-28
Added FromStr impl and tests for HostAndPort

amikhalev:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/264)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -11,6 +11,7 @@ use std::cmp;
 use std::fmt::{self, Formatter};
 use std::io;
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs};
+use std::str::FromStr;
 use std::vec;
 use parser::{ParseResult, ParseError};
 use percent_encoding::percent_decode;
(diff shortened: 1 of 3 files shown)


SimonSapin:
I’m somewhat hesitant to take this since there might be subtle variations of the exact parsing algorithm, and the URL Standard doesn’t specify one. The closest thing would be the setter of [the `host` attribute](https://url.spec.whatwg.org/#dom-url-host) of JavaScript `URL` objects. It’s not obvious to me whether the way that is specified is equivalent to the way you implemented it here.

Which brings the most important question: why do you want this? In what context would you use it? Is there a spec with a parsing algorithm or grammar?

SimonSapin:
> The only issue is that it is a bit unwieldy to construct a HostAndPort. This PR should make
>
> ```rust
> let addr = HostAndPort {
>     host: Host::Domain("localhost"), port: 80,
> };
> ```
>
> into
>
> ```rust
> let addr: HostAndPort = "localhost:80".parse().unwrap();
> ```

It sounds like you just want a simple constructor, but `Host::parse` does much more than that. It will decode any percent-encoding, do Unicode normalization, encode to Punycode if there are non-ASCII characters, …

I’d recommend making your own type and a constructor method for them. Parsing the host part could be something along the lines of "try `std::net::IpAddr::parse`, and if that returns an error take the input as-is and consider it a domain".

But come to think of it, do you even need to parse IP addresses? Could you pass around `(&str, u16)` or `(String, u16)`? The `std::net::ToSocketAddrs` trait has an impl for `(&str, u16)` that does this parsing already, in case you eventually need to resolve this host to an IP address.

dtolnay:
Given that there is no spec with a parsing algorithm or grammar, we are not planning to provide this functionality in `url` right? Let's close the PR.

SimonSapin:
Closing for lack of response from @amikhalev about what exactly is desired.

PR #265, opened 2016-12-30
Make mod host public

valenting:
This just makes the mod public, so I can call into Host::parse() from rust_url_capi.
@Manishearth @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/265)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,13 +143,13 @@ pub use parser::ParseError;
 pub use slicing::Position;
 
 mod encoding;
-mod host;
 mod origin;
 mod path_segments;
 mod parser;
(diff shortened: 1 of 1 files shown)


nox:
You can already call it. https://docs.rs/url/1.2.4/url/enum.Host.html#method.parse

A module doesn't need to be public for public methods in it to be exposed outside, what matters is whether the type on which the method is is exposed or not.

valenting:
Oh, I hadn't noticed the enum is public. @nox thanks for pointing it out. Sorry for the noise.

nox:
@valenting No problem. :)

PR #267, opened 2017-01-02
Only compress sequences of 2 or more zero segments for IPV6 hosts

### LINKED ISSUE #266: IPv6 serialization: Only compress 0 sequences longer than 1
https://url.spec.whatwg.org/#concept-ipv6-serializer - Step 3:
> If there is no sequence of address’s 16-bit pieces that are 0 longer than one, set compress pointer to null. 

As a consequence, http://[fe80:cd00:0:cde:1257:0:211e:729c] should be serialized as it is, not http://[fe80:cd00::cde:1257:0:211e:729c]

@Manishearth @SimonSapin

Manishearth:
(fixes #266)

cc @valenting

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/267)


SimonSapin:
Did the spec change?

Are the test changes from upstream wpt? If not please also send a PR upstream.

Manishearth:
It's always been in the spec (https://github.com/whatwg/url/commit/074fb085d14ade670715497342cd0a071c5e683c#diff-2502abf17adfd0b7f0904c0cbe40ddd7R543)

> If not please also send a PR upstream.


Yeah, I will soon, wanted to wait for review.

PR #269, opened 2017-01-03
Fix warnings.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/269)


Ms2ger:
Looks like the import is used, but should be behind `#[cfg(feature="serde")]`.

SimonSapin:
Indeed, Travis fails:

```rust
error: no method named `description` found for type `parser::ParseError` in the current scope
    --> src/lib.rs:1523:45
     |
1523 |             serde::Error::invalid_value(err.description())
     |                                             ^^^^^^^^^^^
     |
     = help: items from traits can only be used if the trait is in scope; the following trait is implemented but not in scope, perhaps add a `use` for it:
     = help: candidate #1: `use std::error::Error;`

frewsxcv:
Fixed.

PR #271, opened 2017-01-15
add Url::parse_with_params()

kybishop:
Allows us to build a `Url` from user-supplied params without the boilerplate of constructing then modifying the `Url`.

This is in hopes of improving the ergenomics of the reqwest library: https://github.com/seanmonstar/reqwest/pull/45#issuecomment-272589321

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/271)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "1.2.5"
+version = "1.3.0"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,7 @@ use host::HostInternal;
 use parser::{Parser, Context, SchemeType, to_u32};
 use percent_encoding::{PATH_SEGMENT_ENCODE_SET, USERINFO_ENCODE_SET,
                        percent_encode, percent_decode, utf8_percent_encode};
+use std::borrow::Borrow;
 use std::cmp;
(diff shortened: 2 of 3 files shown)


SimonSapin:
Looks good! Just a few small things: please add `assert_eq!(url.as_str(), …)` to the doc-test, remove `.iter()` in the doc-test and add `&` to pass a borrowed array (to demonstrate `IntoIterator` vs `Iterator`), and if you want this on crates.io increment the version number in `Cargo.toml` to 1.3.0.

emilio:
Does it need to be 1.3.0? It's not a breaking change, only adds an API.

SimonSapin:
Yes, exactly. A breaking change would be 2.0.0.

kybishop:
Hurray for semver 😃 

@SimonSapin changes made and commits squashed!

KiChjang:
I think @emilio means whether it can be `1.2.6` instead.

kybishop:
@KiChjang @emilio. Hey all. Apologies if rust-url isn't following true semver, but if it is: "MINOR version when you add functionality in a backwards-compatible manner" - [semver.org](http://semver.org/). Patch version is typically reserved for bug fixes.

Happy to match the version to whatever you guys would like

SimonSapin:
Yeah, 1.2.6 vs 1.3.0 is the same as far as Cargo is concerned. IMO we’d be just fine with just two components in the version number. But since semver is the convention in the rust ecosystem let’s do that.

I was thinking of `assert_eq!` in the doc-test itself to show an example of what the resulting URL looks like, but let’s call this good enough.

@bors-servo r+

Issue #273, opened 2017-01-25
Add support for serde 0.9


maximih:
[Serde 0.9 was just released](https://github.com/serde-rs/serde/releases/tag/v0.9.0) and has some nice new features. 

Missing support in the rust-url stops me from upgrading. Is there any work ongoing to add support for it?

Thanks

SimonSapin:
I’m not sure what’s the best way to approach this. Serde support is part of the public API exposed by the `url` crate, so I think that upgrading to 0.9 is a breaking change. That means it’ll have to be in `url` version 2.0.0, which would be disruptive as upgrades will need to be coordinated through Servo’s dependency graph.

maximih:
I have the implementation ready. Should I update the version to 2.0.0?

SimonSapin:
@nox mentioned an alternative: having serde support in a separate crate, similar to this one: https://docs.rs/hyper_serde/0.2.0/hyper_serde/

maximih:
Wouldn't it still be a breaking change to remove serde from rust-url? Or you want to have only the new serde versions in the url-serde or whatever will be called and keep old serde versions as optional feature in rust-url?

SimonSapin:
I don’t know, both are options.

maximih:
Let me know how you want to have it and I am happy to do the implementation.

SimonSapin:
Alright. In that case, please send a PR to this repository that adds a new `url_serde` crate (in a sub-directory) that does something similar to https://github.com/nox/hyper_serde/blob/master/src/lib.rs (but with serde 0.9). Look at https://github.com/servo/rust-url/blob/bb1100baf0/src/lib.rs#L1735-L1756 for the body of the impls, in particular to deal with errors from `Url::parse`.

This should allow you to use rust-url and serde 0.9 together, though admittedly in a way that’s not as convenient as if this support had been directly in the `url` crate. You’ll also end up with two versions of serde in your `Cargo.lock` file, but only one of them will be built as long as the `serde` feature of the `url` crate is not enabled.

When the time comes to make other breaking changes to the `url` crate and publish version 2.0.0, we’ll also remove support for serde 0.8.

SimonSapin:
Fixed in #274.

PR #274, opened 2017-01-27
Add url_serde crate with ser/de support for url::Url type

maximih:
This is fixing #273

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/274)

### DIFF
--- a/url_serde/Cargo.toml
+++ b/url_serde/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+
+name = "url_serde"
+version = "0.1.0"
+authors = ["The rust-url developers"]
+
+description = "Serde support for URL types"
+documentation = "https://docs.rs/url_serde/"
+repository = "https://github.com/servo/rust-url"
+readme = "README.md"
+keywords = ["url", "serde"]
+license = "MIT/Apache-2.0"
+
+[dependencies]
+serde = "0.9.0"
+url = "1.0.0"
+
+[dev-dependencies]
+serde_json = "0.9.0"
+
+[lib]
+doctest = false
--- a/url_serde/src/lib.rs
+++ b/url_serde/src/lib.rs
@@ -1,0 +1,215 @@
+/*!
+
+This crate provides wrappers and convenience functions to make rust-url
+and Serde work hand in hand.
+
+The supported types are:
+
+* `url::Url`
+
+# How do I use a data type with a `Url` member with Serde?
+
+Use the serde attributes `deserialize_with` and `serialize_with`.
+
+```
+#[derive(serde::Serialize, serde::Deserialize)]
+struct MyStruct {
+    #[serde(deserialize_with = "url_serde::deserialize",
+    serialize_with = "url_serde::serialize")]
+    url: Url,
+}
+```
+
+# How do I encode a `Url` value with `serde_json::to_string`?
+
+Use the `Ser` wrapper.
+
+```
+serde_json::to_string(&Ser::new(&url))
+```
+
+# How do I decode a `Url` value with `serde_json::parse`?
+
+Use the `De` wrapper.
+
+```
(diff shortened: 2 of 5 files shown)


url_serde/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+description = "Serde support for URL types"
+documentation = "http://servo.github.io/rust-url/url/index.html"
+repository = "https://github.com/servo/rust-url"
+readme = "README.md"
+keywords = ["url", "serde"]
+license = "MIT/Apache-2.0"
+
+[dependencies]
+serde = ">=0.9, <0.10"
```
SimonSapin: This is equivalent to `"0.9"`, so please use that. Same for serde_json below.

url_serde/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+
+* `url::Url`
+
+# How do I use a data type with a `Url` member with Serde?
+
+Use the serde attributes `deserialize_with` and `serialize_with`.
+
+```
+struct MyStruct {
+    #[serde(deserialize_with = "url_serde::deserialize",
```
SimonSapin: If you tried that this works in your project?

url_serde/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+use std::fmt;
+use std::ops::{Deref, DerefMut};
+use std::error::Error;
+use serde::{Deserialize, Serialize, Serializer, Deserializer};
+use url::{Url};
+
+
+/// Serialises `value` with a given serializer.
+///
+/// This is useful to serialize Hyper types used in structure fields or
```
SimonSapin: Replace Hyper with rust-url

url_serde/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+/// tuple members with `#[serde(serialize_with = "hyper_serde::serialize")]`.
```
SimonSapin: Replace hyper_serde with url_serde

url_serde/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where S: Serializer
+    {
+        Ser(&self.0).serialize(serializer)
+    }
+}
+
+
+#[test]
```
SimonSapin: Please add a `(cd url_serde && cargo test)` line after `(cd idna && cargo test)` in `Makefile`, so that this test is run on Travis-CI. (Note that the indentation needs to be an actual tab character, not spaces.)

url_serde/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+
+name = "url_serde"
+version = "0.1.0"
+authors = ["The rust-url developers"]
+
+description = "Serde support for URL types"
+documentation = "http://servo.github.io/rust-url/url/index.html"
```
SimonSapin: Please change the documentation URl to https://docs.rs/url_serde/

url_serde/src/lib.rs
maximih: Yes, it does. I would like to add a test case for it but it will need rust nightly until rust 1.15 is released. Is it ok to have a test that run on nightly only?

SimonSapin: We want to keep running the rest of the tests on stable Rust on Travis. We could add a Cargo feature and enable it based on `$TRAVIS_RUST_VERSION`, but I think it’s not worth the bother. I ok with just the non-automated test that you did.

maximih:
PR updated with your comments @SimonSapin

SimonSapin:
Looks great, thanks!

@bors-servo r+

SimonSapin:
And it’s up: https://crates.io/crates/url_serde

PR #276, opened 2017-02-04
Derive Eq and Hash for Serde<T> and add SerdeUrl type alias.

lvillani:
This should make Serde<Url> usable as an HashMap key, since that requires Eq and Hash to be implemented (and Url supports them).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/276)

### DIFF
--- a/url_serde/src/lib.rs
+++ b/url_serde/src/lib.rs
@@ -53,6 +53,7 @@ extern crate url;
 
 use std::cmp::PartialEq;
 use std::fmt;
+use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

lvillani:
@SimonSapin would it be possible to push a new version of the crate?

SimonSapin:
Published v0.1.1. (Next time feel free to increment the version number in `Cargo.toml` in the same PR :))

lvillani:
Thanks a lot @SimonSapin!

PR #281, opened 2017-02-21
Add fuzz scripts

Manishearth:
Add a simple script for fuzzing with https://github.com/rust-fuzz/cargo-fuzz

Mostly autogenerated by cargo-fuzz; the contents of `go()` in parser.rs were
manually written.

cargo-fuzz is still WIP so the format here may change.

This can be fuzzed (on Linux only) with `cargo fuzz --target parser`

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/281)

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,3 @@
+
+target
+libfuzzer
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,15 @@
+
+[package]
+name = "url-fuzz"
+version = "0.0.1"
+authors = ["Automatically generated"]
+
+[dependencies.url]
+path = ".."
+
+[[bin]]
+name = "parse"
+path = "fuzzers/parse.rs"
+
+[workspace]
+members = ["."]
--- a/fuzz/fuzzers/parse.rs
+++ b/fuzz/fuzzers/parse.rs
@@ -1,0 +1,16 @@
+#![no_main]
+
+extern crate fuzzer_sys;
+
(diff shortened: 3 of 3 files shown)


fuzz/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[dependencies.url]
+path = ".."
+
+[[bin]]
+name = "parse"
+path = "fuzzers/parse.rs"
+
+[workspace]
+members = ["."]
```
SimonSapin: Instead of creating a new workspace here, this crate should be added to the workspace at the root of this repository.

fuzz/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+target
+libfuzzer
```
SimonSapin: With the shared workspace (see other comment) there shouldn’t be a `target` directory here. As to `libfuzzer`, I’d prefer it to be `fuzz/libfuzzer` in `.gitignore` at the root of the repo.

Manishearth:
Both those files are autogenerated by cargo-fuzz (the workspace=. is a generic solution so that the fuzz crate doesn't break when placed within a system that has an arbitrary workspace setup). It's better to not share target directories since we build with `RUSTFLAGS` and this may end up polluting dependencies in the regular target dir with sanitizer coverage. IIRC rustc does not rebuild when rustflags change.

SimonSapin:
I think Cargo does rebuild when RUSTFLAGS changes, but ok.

@bors-servo r+

PR #284, opened 2017-02-28
Windows: implement conversion to/from UNC paths

wfraser:
Currently, taking a path like `\\fileserver\share\foo\bar.txt` and trying to make a URL out of it results in an error. However, this *can* be turned into a file URL: `file://fileserver/share/foo/bar.txt`. Note the lack of the usual triple-slash.

The reverse is also implemented: a `file:` URL with a non-empty host is taken to be a UNC path.

This conversion is only enabled on Windows.

I was motivated to do this after trying to run Cargo in a remote directory. Currently it errors out because it uses `rust-url` to convert all its paths to URLs, and fails to convert its working directory to one. I made a build of Cargo using these changes, and it was then able to build in a directory on a remote machine!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/284)


SimonSapin:
Landed with a rebase and some fixes in #360. Sorry this took so long!

PR #291, opened 2017-03-23
reduce space required by idna data tables

froydnj:
The `Range` type defined by uts46 takes up 20 bytes on a 32-bit platform and 32 bytes on a 64-bit platform.  This bloat is due to `Mapping` using a full word-width discriminant for its type and native string slices, which are unnecessarily large.  The use of native string slices in this context also require many load-time relocations on some platforms when idna is used in a static library, another unnecessary cost.

Instead of this scheme, we can define our own string slice type that indexes into a generated table.  This string slice type enables us to use a smaller discriminant type for `Mapping`, and therefore reduces the
size of `Range` to 16 bytes on a 32-bit platform (I think) and 16 bytes on a 64-bit platform.  For x86-64 Linux, this change reduces total idna size by ~130K, a significant savings, and makes more of idna read-only.

Further reductions are possible by splitting the table into smaller variants for the basic multilingual plane vs. everything else; I will leave that reduction for a different PR.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/291)


SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #292, opened 2017-04-18
Fixed passwords getting percent_encoded

Thomspoon:
When testing mysql database connectors, I came across this squirrelly response:

`thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidPort', /Users/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-mac/build/src/libcore/result.rs:86`

My connection string was similar to:

`mysql://root:pass#word@example.com`

Which immediately prompted me to wonder what was happening. I cloned the repo for the connector I was using and started debugging, before I dug deeper and found it was due to the url::Url type panicking. So I delved even deeper and ended up here.

I found that `parse_userinfo` was breaking at the `#`, which would then cause some errors downstream when parsing the rest of the url. 

After this fix, I found that the password, since it wasn't expecting these characters, was `percent_encoding` these characters, so the connection string would turn into:

`mysql://root:pass%23word@example.com`

My proposed change would allow additional special characters in the password, accepting the `?` and `#` characters that were previously disallowed. [IBM](https://www.ibm.com/support/knowledgecenter/en/SSAW57_8.5.5/com.ibm.websphere.nd.doc/ae/csec_chars.html) and [various](https://docs.oracle.com/cd/E11223_01/doc.910/e11197/app_special_char.htm#BABGCBGA) [others](https://www.owasp.org/index.php/Password_special_characters) promote the usage of specifically `#` and `?` in passwords, which is why I think those restrictions should be taken out. 

This is my first pull request, so please let me know if this was too much.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/292)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -16,7 +16,7 @@ use encoding::EncodingOverride;
 use host::{Host, HostInternal};
 use percent_encoding::{
     utf8_percent_encode, percent_encode,
-    SIMPLE_ENCODE_SET, DEFAULT_ENCODE_SET, USERINFO_ENCODE_SET, QUERY_ENCODE_SET,
+    SIMPLE_ENCODE_SET, DEFAULT_ENCODE_SET, QUERY_ENCODE_SET,
     PATH_SEGMENT_ENCODE_SET
 };
 
@@ -685,7 +685,7 @@ impl<'a> Parser<'a> {
                     }
                     last_at = Some((char_count, remaining.clone()))
                 },
-                '/' | '?' | '#' => break,
+                '/' => break,
                 '\\' if scheme_type.is_special() => break,
                 _ => (),
             }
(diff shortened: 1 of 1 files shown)


Thomspoon:
On second though, this only works for my use-case. All of the SQL connector repos use the URL crate, so I'm kind of wondering what the workaround for this would be. `Url::parse_no_percent_encode`?

SimonSapin:
`#` is not simply disallowed as an arbitrary restriction, it is a separator for the "fragment" component of an URL. For example, `structs` is the fragment of `https://docs.rs/url/1.4.0/url/#structs`. As you can see on Travis-CI, your change makes some tests fail.

Using percent-encoding is the correct way.

Thomspoon:
Ah, that makes sense.

 I'm aware why the tests fail, the question now is, would a better way around this problem be to have a `parse_without_encoding` function? The url crate is such a strong crate, that it wouldn't make sense to have repos such as [rust-mysql-simple](https://github.com/blackbeam/rust-mysql-simple) change their codebase when they could change one function to have the same effect. I'm sure a lot of crates could benefit from that if they are using their own scheme.

PR #296, opened 2017-04-22
Add serde 1.0 support

### LINKED ISSUE #282: url_serde: Update documentation to use #[serde(with = "url_serde")]
Added in [Serde 0.9.8](https://github.com/serde-rs/serde/releases/tag/v0.9.8). There should be no code change required on your end.

```diff
 #[derive(Serialize, Deserialize)]
 struct MyStruct {
-    #[serde(serialize_with = "url_serde::serialize",
-            deserialize_with = "url_serde::deserialize")]
+    #[serde(with = "url_serde")]
     url: Url,
 }
```

### LINKED ISSUE #295: url_serde: Update to serde 1.0 and make use of serde remote-derive
Serde 1.0 was just released and has a new nice feature that we can make use of: [remote-derive](https://serde.rs/remote-derive.html). This would improve the ease of use for url_serde.

maximih:
This commit is adding support for serde 1.0 and update docs
to address #282.

Fix #282
Fix #295 

Note1: I had to bump the version of `url_serde` to `0.2` as the changes are not backward compatible and will cause custom derive to fail if the users are using `serde 0.9`.

Note2: I have decided not to use `remote-derive` serde feature for now. I will look into it more and maybe to another PR later.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/296)

### DIFF
--- a/url_serde/src/lib.rs
+++ b/url_serde/src/lib.rs
@@ -22,8 +22,7 @@ Use the serde attributes `deserialize_with` and `serialize_with`.
 ```
 #[derive(serde::Serialize, serde::Deserialize)]
 struct MyStruct {
-    #[serde(deserialize_with = "url_serde::deserialize",
-    serialize_with = "url_serde::serialize")]
+    #[serde(with = "url_serde")]
     url: Url,
 }
 ```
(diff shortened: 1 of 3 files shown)


url_serde/src/lib.rs
```diff
@@ -179,8 +178,8 @@ impl<T> De<T> where De<T>: serde::Deserialize {
 }
 
 /// Deserializes this URL from a `serde` stream.
-impl Deserialize for De<Url> {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer {
+impl<'de> Deserialize<'de> for De<Url> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
         let string_representation: String = Deserialize::deserialize(deserializer)?;
```
SimonSapin: With serde 1.0 we might be able to use `Cow<'de, str>` here instead of `String` (https://serde.rs/borrow.html) and make one less memory allocation in most cases. But that can be done later.

SimonSapin:
Thanks for this PR! Looks good.

As discussed in #295 we’re not quite sure yet how to deal with serde 1.0 in all of Servo’s dependency graph, but that shouldn’t block this. We can always publish other 0.1.x versions of `url_serde` if necessary.

@bors-servo r+

SimonSapin:
https://crates.io/crates/url_serde/0.2.0 is up.

PR #321, opened 2017-05-05
Implement `Hash` for `Origin` (fix #302)

### LINKED ISSUE #302: Implement `Hash` for `Origin`
Seems like this could reasonable be used as a key, and `Hash` is a trait we implement eagerly.

cGuille:
Hello,

I'm a Rust / Open Source beginner, so just explain if anything is wrong. :-)

I am giving a try at issue #302. I just derived the `Hash` for the `Origin` type.

I have zero experience with testing in Rust so I just tried to demonstrate that it works, but I do not know how to write a good test for this.

What should I do now?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/321)

### DIFF
--- a/src/origin.rs
+++ b/src/origin.rs
@@ -50,7 +50,7 @@ pub fn url_origin(url: &Url) -> Origin {
 ///   the URL does not have the same origin as any other URL.
 ///
 /// For more information see https://url.spec.whatwg.org/#origin
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
 pub enum Origin {
     /// A globally unique identifier
     Opaque(OpaqueOrigin),
@@ -123,7 +123,7 @@ impl Origin {
 }
 
 /// Opaque identifier for URLs that have file or other schemes
-#[derive(Eq, PartialEq, Clone, Debug)]
+#[derive(Eq, PartialEq, Hash, Clone, Debug)]
 pub struct OpaqueOrigin(usize);
 
 #[cfg(feature = "heapsize")]
--- a/tests/unit.rs
+++ b/tests/unit.rs
@@ -372,3 +372,45 @@ fn define_encode_set_scopes() {
 
     m::test();
 }
+
+#[test]
+/// https://github.com/servo/rust-url/issues/302
+fn test_origin_hash() {
+    use std::hash::{Hash,Hasher};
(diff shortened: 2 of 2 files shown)


skade:
@cGuille Wait for one of the maintainers to provide review :). Most maintainers of the URL crate live in the US, so this might take until later.

Ignore the nightly error in the linked travis test suite, rust nightly is currently broken, not this lib.

emilio:
@SimonSapin is french fwiw, hopefully he can take a look soon-ish.

Can you test also that opaque origins get different hashes? (It's trivial, but worth having that in the test-suite too).

cGuille:
Hello,

There is actually no hurry at all. :-)

I am OK with adding more tests ; I will just have to figure out how.
I do not like using `HashMap` to indirectly test the `Hash` trait ; it  would be nice to have someone else's opinion about this.

SimonSapin:
@skade, unfortunately, most of the maintainers of the url crate are me :) I’m usually in UTC+1~2.

@cGuille You can hash without a hashmap like so:

```rust
use std::collections::hash_map::DefaultHasher;
use std::hash::Hash;

fn hash<T: Hash>(value: T) -> u64 {
    let mut hasher = DefaultHasher::new();
    value.hash(&mut hasher);
    hasher.finish()
}
```

To make the test more interesting, you could use different URLs that have the same origin. Please also test that hashing is consistent with `==` equality.

Thanks!

cGuille:
Thank you!

I updated the PR to test hash values directly instead of using `HashMap`, but I do not know how to test that hashing is consistent with equality. Do you have something in mind?
Should I just check that equalities of the origins in my test are the same as the equalities of their hashes?

SimonSapin:
Test that things that compare equal with `==` have the same hash. Test that things that compare unequal have a different hash. The latter is not guaranteed because there could be hashing collisions, but it is unlikely enough that it’s alright to test in a couple fixed-input test cases.

cGuille:
What do you think about this? Are there enough URL samples?

cGuille:
I solved the conflict with `master` and rebased the branch upon it.

Tell me if there is anything left to do regarding this PR.

cGuille:
I rebased upon master to solve this second conflict.

Can this be merged?

brson:
@bors-servo r+

bors-servo probably doesn't listen to me...

brson:
Sorry for the delay @cGuille !

Issue #299, opened 2017-05-04
Replace `()` error types with specific types that implement Error


brson:
There are a number of methods in url that return `Result<_, ()>`. `()` though does not implement `Error` and so does not inter operate cleanly with callers that want to treat the result as `Error` (like error-chain).

In today's Rust the best pattern for this is probably to create a single `Error` enum for the entire crate to share and return it everywhere.

Would require a major version bump.

dtolnay:
Yep we struggled with this in #323. It makes it hard to handle these errors concisely.

Enet4:
Here's a list of public methods returning `()` as an error:

1.  [`Url#path_segments_mut`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1038)
1. [`Url#set_port`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1089)
1. [`Url#set_ip_host`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1258)
1. [`Url#set_password`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1274)
1. [`Url#set_username`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1323)
1. [`Url#set_scheme`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1410)
1. [`Url#from_file_path`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1460)
1. [`Url#from_directory_path`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1495)
1. [`Url#to_file_path`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L1575)

Although [`Url#path_segments`](https://github.com/servo/rust-url/blob/780410c0f161f57bc1a58e40fab1930b703456da/src/lib.rs#L854) returns an `Option`, this is not very consistent when we look at its "_mut" counterpart `path_segments_mut`, which returns a `Result`.

[...]

PR #323, opened 2017-05-05
Replace unwrap() in doctests with `?`

### LINKED ISSUE #312: Modify examples to not call `unwrap`
Several examples in `url` call `unwrap`. It's preferable to use `?` to avoid teaching bad practices to new Rust coders. Modify the examples to have some hidden scaffolding to set up the `Result` type and use `?` instead.

Enet4:
Fixes #312.
 - All results yielding `ParseError` in the doctests should now be handled with the ? operator.
 - Some other specific results are ~unwrapped with `expect()`~ mapped to a placeholder string message, such as in URLs that are not _cannot-be-a-base_. Users are then expected to adjust this mapping depending on their use case. Once #299 is resolved, we'll be able to propagate these normally.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/323)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,10 +26,11 @@ Let’s parse a valid URL and look at its components.
 
 ```
 use url::{Url, Host};
-
+# use url::ParseError;
+# fn run() -> Result<(), ParseError> {
 let issue_list_url = Url::parse(
     "https://github.com/rust-lang/rust/issues?labels=E-easy&state=open"
-).unwrap();
+)?;
 
 
 assert!(issue_list_url.scheme() == "https");
@@ -44,6 +45,8 @@ assert!(issue_list_url.path_segments().map(|c| c.collect::<Vec<_>>()) ==
 assert!(issue_list_url.query() == Some("labels=E-easy&state=open"));
 assert!(issue_list_url.fragment() == None);
 assert!(!issue_list_url.cannot_be_a_base());
+# Ok(())
+# }
 ```
 
 Some URLs are said to be *cannot-be-a-base*:
@@ -52,8 +55,10 @@ and their "path" is an arbitrary string rather than slash-separated segments:
 
 ```
 use url::Url;
+# use url::ParseError;
 
-let data_url = Url::parse("data:text/plain,Hello?World#").unwrap();
+# fn run() -> Result<(), ParseError> {
+let data_url = Url::parse("data:text/plain,Hello?World#")?;
 
 assert!(data_url.cannot_be_a_base());
 assert!(data_url.scheme() == "data");
@@ -61,6 +66,8 @@ assert!(data_url.path() == "text/plain,Hello");
 assert!(data_url.path_segments().is_none());
 assert!(data_url.query() == Some("World"));
 assert!(data_url.fragment() == Some(""));
+# Ok(())
+# }
 ```
 
 
@@ -84,10 +91,14 @@ Use the `join` method on an `Url` to use it as a base URL:
 
 ```
 use url::Url;
+# use url::ParseError;
 
-let this_document = Url::parse("http://servo.github.io/rust-url/url/index.html").unwrap();
-let css_url = this_document.join("../main.css").unwrap();
-assert_eq!(css_url.as_str(), "http://servo.github.io/rust-url/main.css")
+# fn run() -> Result<(), ParseError> {
+let this_document = Url::parse("http://servo.github.io/rust-url/url/index.html")?;
+let css_url = this_document.join("../main.css")?;
+assert_eq!(css_url.as_str(), "http://servo.github.io/rust-url/main.css");
+# Ok(())
+# }
 */
 
 #[cfg(feature="rustc-serialize")] extern crate rustc_serialize;
@@ -218,8 +229,12 @@ impl Url {
     ///
     /// ```rust
     /// use url::Url;
+    /// # use url::ParseError;
     ///
-    /// let url = Url::parse("https://example.net").unwrap();
+    /// # fn run() -> Result<(), ParseError> {
+    /// let url = Url::parse("https://example.net")?;
+    /// # Ok(())
+    /// # }
     /// ```
     #[inline]
     pub fn parse(input: &str) -> Result<Url, ::ParseError> {
@@ -234,9 +249,13 @@ impl Url {
     ///
     /// ```rust
     /// use url::Url;
+    /// # use url::ParseError;
     ///
+    /// # fn run() -> Result<(), ParseError> {
     /// let url = Url::parse_with_params("https://example.net?dont=clobberme",
-    ///                                  &[("lang", "rust"), ("browser", "servo")]);
+    ///                                  &[("lang", "rust"), ("browser", "servo")])?;
+    /// # Ok(())
+    /// # }
     /// ```
     #[inline]
     pub fn parse_with_params<I, K, V>(input: &str, iter: I) -> Result<Url, ::ParseError>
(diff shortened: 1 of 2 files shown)


KiChjang:
Why are there so many `#` symbols added?

Enet4:
@KiChjang They define lines of code that we want to consider in the tests but hide in the documented output. This is needed here to hide the boilerplate. The code needs to be wrapped in a function that returns a result in order to use the `?` operator, but that should be implicit in the examples.

budziq:
There seam to be few `unwrap` calls left in the doctests immediately surrounding the change.
Also, would it not be beneficial to leave the `run` boilerplate at least once? As url does not provide convenience `Result` it was actually easier for me look into the source code to find the proper return type than analyze the docs in their current form.

Enet4:
> There seam to be few unwrap calls left in the doctests immediately surrounding the change.

Can you specify where? Note that lines hidden from the docs do not need to have their errors captured. Moreover, a good part of the unwraps still around are unwrapping `Option`, not `Result`.

> Also, would it not be beneficial to leave the run boilerplate at least once? As url does not provide convenience Result it was actually easier for me look into the source code to find the proper return type than analyze the docs in their current form.

That would be debatable. Although a `Result` alias is not provided, the documented methods do show the error type `ParseError`, which is exported at the root of the crate. I suppose the crate could consider making the `Result` alias.

src/path_segments.rs
```diff
@@ -18,13 +18,17 @@ use Url;
...
-/// let mut url = Url::parse("mailto:me@example.com").unwrap();
+/// # use url::{Url, ParseError};
+///
+/// # fn run() -> Result<(), ParseError> {
+/// let mut url = Url::parse("mailto:me@example.com")?;
 /// assert!(url.path_segments_mut().is_err());
 ///
-/// let mut url = Url::parse("http://example.net/foo/index.html").unwrap();
+/// let mut url = Url::parse("http://example.net/foo/index.html")?;
 /// url.path_segments_mut().unwrap().pop().push("img").push("2/100%.png");
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
```diff
@@ -60,10 +64,14 @@ impl<'a> PathSegmentsMut<'a> {
     /// Example:
     ///
     /// ```rust
-    /// # use url::Url;
-    /// let mut url = Url::parse("https://github.com/servo/rust-url/").unwrap();
+    /// # use url::{Url, ParseError};
+    /// 
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("https://github.com/servo/rust-url/")?;
     /// url.path_segments_mut().unwrap().clear().push("logout");
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
```diff
@@ -81,14 +89,18 @@ impl<'a> PathSegmentsMut<'a> {
     /// Example:
     ///
     /// ```rust
-    /// # use url::Url;
-    /// let mut url = Url::parse("https://github.com/servo/rust-url/").unwrap();
+    /// # use url::{Url, ParseError};
+    /// 
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("https://github.com/servo/rust-url/")?;
     /// url.path_segments_mut().unwrap().push("pulls");
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
```diff
@@ -81,14 +89,18 @@ impl<'a> PathSegmentsMut<'a> {
...
     /// assert_eq!(url.as_str(), "https://github.com/servo/rust-url//pulls");
     ///
-    /// let mut url = Url::parse("https://github.com/servo/rust-url/").unwrap();
+    /// let mut url = Url::parse("https://github.com/servo/rust-url/")?;
     /// url.path_segments_mut().unwrap().pop_if_empty().push("pulls");
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
```diff
@@ -138,23 +150,31 @@ impl<'a> PathSegmentsMut<'a> {
...
-    /// # use url::Url;
-    /// let mut url = Url::parse("https://github.com/").unwrap();
+    /// # use url::{Url, ParseError};
+    /// 
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("https://github.com/")?;
     /// let org = "servo";
     /// let repo = "rust-url";
     /// let issue_number = "188";
     /// url.path_segments_mut().unwrap().extend(&[org, repo, "issues", issue_number]);
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
```diff
@@ -138,23 +150,31 @@ impl<'a> PathSegmentsMut<'a> {
...
     /// a segment is ignored if it is `"."` or `".."`:
     ///
     /// ```rust
-    /// # use url::Url;
-    /// let mut url = Url::parse("https://github.com/servo").unwrap();
+    /// # use url::{Url, ParseError};
+    /// 
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("https://github.com/servo")?;
     /// url.path_segments_mut().unwrap().extend(&["..", "rust-url", ".", "pulls"]);
```
budziq: `unwrap()` still used in doctest

src/path_segments.rs
Enet4: That is something I would resolve in another PR: `path_segments_mut` returns `Result<_,()>`. With the API like this, we cannot capture it with the `?` without mapping the error or rolling a custom error type. Note that using `()` as error type is not recommended by the Rust API guidelines ([C-MEANINGFUL-ERR](https://github.com/brson/rust-api-guidelines#c-meaningful-err)).

src/path_segments.rs
Enet4: Same issue, error type is `()`.

src/path_segments.rs
Enet4: Same issue, error type is `()`.

Enet4:
For the reasons stated [above](https://github.com/servo/rust-url/pull/323#discussion_r115454520), I suggest that #299 is resolved before those unwraps are replaced.

dtolnay:
Yes the `()` errors make error handling obnoxious. We would all love a fix for #299.

But as it is, `unwrap()` is not the way you would handle these in real code. I think the examples should show realistic error handling even if it is awkward, because people will paste these examples into their own code.

Enet4:
Ok, I will look into them today, possibly devising an error type for the purpose. I also wish to help addressing #299, since it appears we could highly benefit from better error handling.

dtolnay:
Thanks! That eliminates the remainder of the `unwrap`s.

Some of these have a bit more error-handling noise than I would find valuable. For example the following one from `PathSegmentsMut::clear` is 12 lines of error handling to 4 lines of example code.

```rust
use url::Url;
use std::error::Error;
use std::fmt;

#[derive(Debug)]
struct CannotBeBaseError;
impl fmt::Display for CannotBeBaseError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        f.write_str(self.description())
    }
}
impl Error for CannotBeBaseError {
    fn description(&self) -> &str { "cannot be a base" }
}

let mut url = Url::parse("https://github.com/servo/rust-url/")?;
url.path_segments_mut().map_err(|_| CannotBeBaseError)?
    .clear().push("logout");
assert_eq!(url.as_str(), "https://github.com/logout");
```

Since `Box<Error>` implements `From<&str>`, what do you think of the following? In contrast to `unwrap`, this still leaves the user an obvious place to plug in their own error handling after they copy+paste this code. They can't forget to do it like they probably would with `unwrap`.

```rust
use url::Url;

let mut url = Url::parse("https://github.com/servo/rust-url/")?;
url.path_segments_mut().map_err(|_| "cannot be a base")?
    .clear().push("logout");
assert_eq!(url.as_str(), "https://github.com/logout");
```

Enet4:
I agree that the current approach is verbose (not to mention that it definitely looks like something that should have been an error variant in the first place!). I'll follow the advice of using `&'static str` as a placeholder error and push a commit right away.

src/lib.rs
```diff
@@ -44,6 +45,8 @@ assert!(issue_list_url.path_segments().map(|c| c.collect::<Vec<_>>()) ==
 assert!(issue_list_url.query() == Some("labels=E-easy&state=open"));
 assert!(issue_list_url.fragment() == None);
 assert!(!issue_list_url.cannot_be_a_base());
+# Ok(())
+# }
```
bjorn3: `fn run` is not called

src/lib.rs
```diff
@@ -52,15 +55,19 @@ and their "path" is an arbitrary string rather than slash-separated segments:
 
 ```
 use url::Url;
+# use url::ParseError;
 
-let data_url = Url::parse("data:text/plain,Hello?World#").unwrap();
+# fn run() -> Result<(), ParseError> {
```
bjorn3: Same here

Enet4:
@bjorn3 Oops, you are right! I appended calls to `run()` in the latest commit.

SimonSapin:
This looks great. Thanks all!

@bors-servo r+

PR #327, opened 2017-05-06
Add support for Serde 1.0

### LINKED ISSUE #304: Move serde 1.0 support in-crate
Today serde 1.0 is supported in its own `url_serde` crate, so that url can support both old serde and new serde. @dtolnay has some plan to hack support for both in-crate, though presumably serde_url will need to continue on until a major bump.

### LINKED ISSUE #320: Figure out a way to support multiple versions of serde in-crate
This is a requirement for doing #304 without a breaking change.

@dtolnay suggested a plan along these lines:

* Make a `serde_1_0` crate that depends on serde 1.0 and only contains `pub use serde::*;`. Make other similar crates for other mutually-incompatible versions of serde.  (I’d rather the serde project maintain these crates, if they’re willing.)
* Have `url` optionally depend on these crates, with Cargo feature flags named `serde_1_0` etc.
* Implement multiple versions of the `Serialize` and `Deserialize` traits. For example `impl serde_1_0::Serialize for Url {…} impl serde_0_9::Serialize for Url {…}` etc.
* Publish a new, compatible version of `url_serde` that is marked `#[deprecated]` with a message that point to the feature flags.

Open question: can `#[derive]` be used in this setup? How?

These crates with a version number in the name are a hack to work around a limitation of Cargo: a crate can not depend on multiple crates with the same name. (That is, multiple mutually-incompatible versions of the same crate.) A better solution would be to make Cargo support that scenario directly. This probably involves locally renaming a dependency for disambiguating `extern crate`.

Help wanted to write a Cargo RFC.

dtolnay:
Fixes #304, fixes #320.

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/327)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ matches = "0.1"
 percent-encoding = { version = "1.0.0", path = "./percent_encoding" }
 rustc-serialize = {version = "0.3", optional = true}
 serde = {version = ">=0.6.1, <0.9", optional = true}
+serde1 = {version = "1.0", optional = true}
 
 [[bench]]
 name = "parse_url"
--- a/src/host.rs
+++ b/src/host.rs
@@ -27,38 +27,6 @@ pub enum HostInternal {
 #[cfg(feature = "heapsize")]
 known_heap_size!(0, HostInternal);
 
-#[cfg(feature="serde")]
-impl ::serde::Serialize for HostInternal {
-    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: ::serde::Serializer {
-        // This doesn’t use `derive` because that involves
-        // large dependencies (that take a long time to build), and
-        // either Macros 1.1 which are not stable yet or a cumbersome build script.
-        //
-        // Implementing `Serializer` correctly for an enum is tricky,
-        // so let’s use existing enums that already do.
-        use std::net::IpAddr;
-        match *self {
-            HostInternal::None => None,
-            HostInternal::Domain => Some(None),
-            HostInternal::Ipv4(addr) => Some(Some(IpAddr::V4(addr))),
-            HostInternal::Ipv6(addr) => Some(Some(IpAddr::V6(addr))),
-        }.serialize(serializer)
-    }
-}
-
-#[cfg(feature="serde")]
-impl ::serde::Deserialize for HostInternal {
-    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error> where D: ::serde::Deserializer {
-        use std::net::IpAddr;
-        Ok(match ::serde::Deserialize::deserialize(deserializer)? {
-            None => HostInternal::None,
-            Some(None) => HostInternal::Domain,
-            Some(Some(IpAddr::V4(addr))) => HostInternal::Ipv4(addr),
-            Some(Some(IpAddr::V6(addr))) => HostInternal::Ipv6(addr),
-        })
-    }
-}
-
 impl<S> From<Host<S>> for HostInternal {
     fn from(host: Host<S>) -> HostInternal {
         match host {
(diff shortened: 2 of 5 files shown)


SamWhited:
LGTM; thanks for the comments on the copy/pasted code that could be custom derived, saved me trying to figure out why derive wasn't used.

dtolnay:
I fixed the conflict with #331.

drusellers:
Looking forward to serde 1.0 support landing

SimonSapin:
I find the code in `src/serde.rs`, especially `Deserilize` impls for enum, to be overly verbose. It sounds like `extern crate serde;` in the code generated by `derive` is the only reason we can’t use `derive` instead. I’ve filed https://github.com/serde-rs/serde/issues/953 to request a serde feature that hopefully will fix that. I’d much prefer to use derive here.

SimonSapin:
@drusellers In the meantime, https://docs.rs/url_serde/0.2.0/url_serde/ supports serde 1.0 today.

konstin:
Unfortunately, url_serde doesn't work on existing types, e.g. `Vec<Url>`, so using serde in my project is currently blocked by this PR.

Object905:
@konstin you can use `Vec<url_serde::SerdeUrl>` instead.

killercup:
I really wanted to do a cool demo in a talk where I fetch <http://readrust.net/rust2018/feed.json>, first show a nice list of posts, and as a bonus quickly extract the domain to see ho many people are using gist.github.com as their blog. Not having serde 1.0 support basically destroys that bonus feature :(

SimonSapin:
@killercup See https://github.com/servo/rust-url/pull/327#issuecomment-308139040 above.

killercup:
@SimonSapin that's why I wrote "I really wanted to do a *cool* demo" 😉 

Adding two crates and using

```rust
#[serde(with = "url_serde")]
url: Url,
```

is not nearly as cool as just writing `url: Url,`.

SimonSapin:
Sorry that SemVer spaghetti is not cool :) I’m considering doing url 2.0 at some point to fix this and other things, but for Servo to use be able to update it’d first need to update hyper from 0.10 to 0.11, which seems to be blocked on websocket support: https://github.com/servo/servo/issues/18518.

But since I’d rather not maintain two branches of url and I seem to be the one doing all that work anyway, url 2.0 will wait.

killercup:
Ah, I see. That's indeed a pretty big chunk of spaghetti you have on your plate!

I kinda thought using the serde09 crate would resolve this, but I guess the issues run quite a lot deeper than that.

dtolnay:
@SimonSapin I rebased and removed the 1.0 impls for Host\<S\> and HostInternal, because I believe most users are really only looking for Serde impls for url::Url. That one does not require serde_derive because we are serializing the url in string form.

LegNeato:
Any update on this? Is this landable? Did something similar land?

est31:
The 2.0 branch [has been opened](https://github.com/servo/rust-url/issues/463#issuecomment-510963476). Per [prior statements](https://github.com/servo/rust-url/issues/304#issuecomment-430619148), this PR should be closed and instead a PR against that branch filed.

I have made an alternative serde 1.0 implementation that's closer to the current implementation: https://github.com/servo/rust-url/compare/2.0...est31:serde_1.0

I think that this PR's serialization should be used instead where url's are just serialized as strings.

SimonSapin:
Thanks for the PR, David. The version looks more reasonable than the original.

I’m going to close this since I took https://github.com/servo/rust-url/pull/512 instead, to be part of url 2.0.0. Hopefully it’ll be easy enough for the ecosystem to migrate, but if not we can consider reviving this PR in the `1.x` branch.

PR #328, opened 2017-05-07
Add Display for HostAndPort

### LINKED ISSUE #300: Implement `Display` for `Host` and `HostAndPort`
Probably using the obvious serialization for "host:port". Looks like `HostAndPort` impl needs to be parameterized over `S: Display`.

tomecki:
Fix #300

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/328)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -192,6 +192,15 @@ impl<'a> HostAndPort<&'a str> {
     }
 }
 
+impl<S: AsRef<str>> fmt::Display for HostAndPort<S> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        try!(self.host.fmt(f));
+        try!(f.write_str(":"));
+        self.port.fmt(f)
+    }
+}
+
+
 impl<S: AsRef<str>> ToSocketAddrs for HostAndPort<S> {
     type Iter = SocketAddrs;
 
--- a/src/host.rs
+++ b/src/host.rs
@@ -166,8 +166,8 @@ impl<S: AsRef<str>> fmt::Display for Host<S> {
             Host::Domain(ref domain) => domain.as_ref().fmt(f),
             Host::Ipv4(ref addr) => addr.fmt(f),
             Host::Ipv6(ref addr) => {
-                f.write_str("[")?;
-                write_ipv6(addr, f)?;
+                try!(f.write_str("["));
+                try!(write_ipv6(addr, f));
                 f.write_str("]")
             }
         }
--- a/src/host.rs
+++ b/src/host.rs
@@ -194,8 +194,8 @@ impl<'a> HostAndPort<&'a str> {
 
 impl<S: AsRef<str>> fmt::Display for HostAndPort<S> {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        try!(self.host.fmt(f));
-        try!(f.write_str(":"));
+        self.host.fmt(f)?;
+        f.write_str(":")?;
         self.port.fmt(f)
     }
 }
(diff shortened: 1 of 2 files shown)


src/host.rs
```diff
@@ -192,6 +192,15 @@ impl<'a> HostAndPort<&'a str> {
     }
 }
 
+impl<S: AsRef<str>> fmt::Display for HostAndPort<S> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        try!(self.host.fmt(f));
```
SamWhited: I wonder if there's an easy way to write this all at once so that we return an Err Result and don't write anything to the formatter, or we return Ok and the formatter has the full thing? I don't think Display makes that guarantee, but it seems like a "nice to have".

SimonSapin: @SamWhited I don’t understand what you’re proposing. Could you explain some more or give an example?

In any case, it is incorrect for a `Display::fmt` impl to return `Err` other than for propagating errors that ultimately come from the `Formatter`. Formatting is supposed to be infallible, it’s only writing to the output stream (for example to a file on a full disk) that can fail.

SamWhited: Yes, sorry, on rereading that was confusing.

I'm suggesting that if possible we do a single, atomic, write to the formatter. Right now we might write `host:` and then something bad happens and writing the port fails so we end up with a half written thing.

It's probably not actually an issue.

Pistahh: what about using the "new" ? operator?

~~~
self.host.fmt(f)?;
self.write_str(":")?;
self.port.fmt(f)
~~~

SamWhited: I was about to suggest that as a nit pick :)

SimonSapin: `Display` impls are pretty much never "atomic" in that sense. On the contrary, they are designed to be composable like this. The only way I can think of to make on "atomic" would be to first format to a `String` and then write that, but that would add unnecessary costly memory allocations. A large number of them, for deeply nested `Display::fmt` calls.

SimonSapin: Let’s do this separately, on the whole repo: https://github.com/servo/rust-url/issues/330

SamWhited: ooh yah, an allocation would be worse. LGTM then.

SimonSapin: Since #331 has landed first, please switch to `?`. Thanks!

tests/unit.rs
```diff
@@ -255,6 +255,30 @@ fn test_form_serialize() {
...
+                host: Host::Ipv6(Ipv6Addr::new(
+                    0x2001, 0x0db8, 0x85a3, 0x08d3, 0x1319, 0x8a2e, 0x0370, 0x7344
+                )),
+                port: 1337
+            },
+            "[2001:db8:85a3:8d3:1319:8a2e:370:7344]:1337"
+        )
+    ];
+
+    for &(ref input, result) in &data {
```
SimonSapin: Avoid `for` loops like this for unit tests. When a test fail it will show the line number for `assert_eq!`, but that doesn’t show which of the entries failed.

In this case I think duplicating the `assert_eq!` + `format!` line is fine, since it’s not long and there’s only two test cases. In a case when de-duplicating testing code is desirable, consider using a macro like `assert_from_file_path!` in this file. This makes panic messages use the line number of the macro usage.

tomecki:
I addressed your comments, and added the `?` operator to Display for Host. Thank you very much for your input!

tomecki:
Oops, I noticed, that `?` has been added to Display for Host, added a revert commit.

sigmavirus24:
:lgtm: This looks good. Thanks for the extra tests too :)

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/328#-:-KjmA3w6Mmou_6IzeWul:bfdr521)*

tmccombs:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/328)*

SimonSapin:
Looks good.

@bors-servo r+

PR #329, opened 2017-05-09
Upgrade unicode-bidi to 0.3.0

behnam:
Replace calls to deprecated API (`bidi_class()`) with the new function
name.

Depends on https://github.com/servo/unicode-bidi/pull/27

r? @emilio @mbrubeck

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/329)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "url"
-version = "1.4.0"
+version = "1.4.1"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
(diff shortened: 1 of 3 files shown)


SimonSapin:
Looks good to me, with the path dependency reverted to crates.io.

behnam:
Thanks, @SimonSapin . Yeah, I'll fix up `Cargo.toml` after we have `0.3.0` out and update this diff to land.

behnam:
In some iterations, we landed on having the part of `unicode-bidi` API that's being used here unchanged into `0.3` version, so this diff is not necessary anymore.

PR #331, opened 2017-05-09
Replace usage of try! with ?

### LINKED ISSUE #330: Untry
Run https://github.com/japaric/untry on every `.rs` file in this repository to replace usage of the `try!(…)` macro with the `…?` operator.

SamWhited:
Replaces all uses of `try!` with the `?` operator.

Fixes #330

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/331)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -781,7 +781,7 @@ impl Url {
     /// # use std::io;
     ///
     /// fn connect(url: &Url) -> io::Result<TcpStream> {
-    ///     TcpStream::connect(try!(url.with_default_port(default_port)))
+    ///     TcpStream::connect(url.with_default_port(default_port)?)
     /// }
     ///
     /// fn default_port(url: &Url) -> Result<u16, ()> {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -797,13 +797,13 @@ impl Url {
...
+            host: self.host()
                            .ok_or(())
-                           .or_else(|()| io_error("URL has no host"))),
-            port: try!(self.port_or_known_default()
+                           .or_else(|()| io_error("URL has no host"))?,
+            port: self.port_or_known_default()
                            .ok_or(())
                            .or_else(|()| f(self))
-                           .or_else(|()| io_error("URL has no port number")))
+                           .or_else(|()| io_error("URL has no port number"))?
```
SimonSapin: Please re-align these lines. (`untry` should have shown a warning about multi-line replacement.)

SamWhited: Fixed; is there a specific set of rustfmt settings that should be used? I tried to use it originally but it realigned a ton of other stuff so I left any sort of alignment issues alone. Thanks!

SimonSapin: rustfmt does a lot more things than just replacing `try!`. If we’re gonna go down that road I’d prefer to go all the way and enforce rustfmt-like style in CI. But that seems more disruptive than I wanted to get into right now, so I recommended untry instead.

SimonSapin:
Great, thanks!

@bors-servo r+

src/lib.rs
SamWhited: Ah, I see; untry gave me some warning about beign deprecated in favor of rustfmt; anyways, doesn't matter, fixed now. Thanks!

PR #335, opened 2017-05-10
Fix #305 Implement Debug for many types

### LINKED ISSUE #305: Implement Debug for many types
We generally implement Debug for all types in libraries that are intended for general use.

Add `Debug` for HostAndPort, ParseOptions, PathSegmentsMut, SocketAddrs, UrlQuery, ByteSerialize, Parse, ParseIntoOwned, Serializer, `*_ENCODE_SET`, PercentEncode, PercentDecode.

imor:
For most types it was a simple matter of adding #[derive(Debug)] but
ParseOptions needed a manual implementation because the type of
log_syntax_violation is Option<&'a Fn(&'static str)> and Fn doesn't
implement Debug. log_syntax_violation is formatted as Some(Fn(&'static
str)) or None depending upon its value.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/335)

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -13,6 +13,7 @@
 #[cfg(feature = "query_encoding")] extern crate encoding;
 
 use std::borrow::Cow;
+#[cfg(feature = "query_encoding")] use std::fmt::{self, Write, Debug, Formatter};
 
 #[cfg(feature = "query_encoding")] use self::encoding::types::{DecoderTrap, EncoderTrap};
 #[cfg(feature = "query_encoding")] use self::encoding::label::encoding_from_whatwg_label;
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -95,7 +95,7 @@ impl Debug for EncodingOverride {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "EncodingOverride {{ encoding: ")?;
         match self.encoding {
-            Some(_) => write!(f, "Some(EncodingOverride) }}"),
+            Some(e) => write!(f, "{} }}", e.name()),
             None => write!(f, "None }}")
         }
     }
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -19,7 +19,7 @@ use std::borrow::Cow;
 #[cfg(feature = "query_encoding")] pub use self::encoding::types::EncodingRef;
 
 #[cfg(feature = "query_encoding")]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct EncodingOverride {
     /// `None` means UTF-8.
     encoding: Option<EncodingRef>
@@ -91,7 +91,7 @@ impl EncodingOverride {
 
 
 #[cfg(not(feature = "query_encoding"))]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct EncodingOverride;
 
 #[cfg(not(feature = "query_encoding"))]
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -19,7 +19,7 @@ use std::borrow::Cow;
 #[cfg(feature = "query_encoding")] pub use self::encoding::types::EncodingRef;
 
 #[cfg(feature = "query_encoding")]
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone)]
 pub struct EncodingOverride {
     /// `None` means UTF-8.
     encoding: Option<EncodingRef>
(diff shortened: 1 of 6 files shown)


saghm:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/encoding.rs, line 98 at r1](https://reviewable.io:443/reviews/servo/rust-url/335#-Kjs8CneImvzHu6G6ha0:-Kjs8Cnfb5iKpuRuXaSB:b-48mrg5) ([raw file](https://github.com/servo/rust-url/blob/01e70b9f69bf0b21246fcbd26e37a94719eda74d/src/encoding.rs#L98)):*
> ```Rust
>         write!(f, "EncodingOverride {{ encoding: ")?;
>         match self.encoding {
>             Some(_) => write!(f, "Some(EncodingOverride) }}"),
> ```

Could this instead call [`name`](https://docs.rs/encoding/0.2.33/encoding/types/trait.Encoding.html#tymethod.name) on the wrapped encoding ref? I think it might be useful to be able to see the the type of the encoding if it's present.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/335)*

saghm:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/335)*

src/lib.rs
```diff
@@ -213,6 +213,16 @@ impl<'a> ParseOptions<'a> {
     }
 }
 
+impl<'a> Debug for ParseOptions<'a> {
```
turnage: The body of this function can just be

    (self.base_url, self.encoding_override).fmt(f)

src/encoding.rs
```diff
@@ -89,9 +90,19 @@ impl EncodingOverride {
     }
 }
 
+#[cfg(feature = "query_encoding")]
+impl Debug for EncodingOverride {
```
turnage: You can just derive Debug on the structs compiled for each feature.

src/percent_encoding.rs
```diff
@@ -77,7 +77,7 @@ macro_rules! define_encode_set {
 }
 
 /// This encode set is used for the path of cannot-be-a-base URLs.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 #[allow(non_camel_case_types)]
 pub struct SIMPLE_ENCODE_SET;
```
turnage: You need to derive Debug in the encoding set macro to get the rest of *_ENCODE_SET types.

dtolnay:
The `write!` approach isn't a great way to do Debug implementations because it doesn't respect the various flags on the [`Formatter`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html).

```rust
use std::fmt::{self, Debug};

struct Good { imor: i32 }
impl Debug for Good {
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        formatter.debug_struct("Good")
            .field("imor", &self.imor)
            .finish()
    }
}

struct Bad { imor: i32 }
impl Debug for Bad {
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        write!(formatter, "Bad {{ imor: {:?} }}", self.imor)
    }
}

fn main() {
    let good = Good { imor: 0 };
    let bad = Bad { imor: 0 };
    println!("default: {:?}", good);
    println!("indented: {:#?}", good);
    println!();
    println!("default: {:?}", bad);
    println!("indented: {:#?}", bad);
}
```

```
default: Good { imor: 0 }
indented: Good {
    imor: 0
}

default: Bad { imor: 0 }
indented: Bad { imor: 0 }
```

Is there a way to do these only using `derive(Debug)` and `debug_struct`?

src/encoding.rs
imor: Can't derive Debug in this case because encoding field has type Option<EncodingRef> and EncodingRef trait isn't Debug

src/percent_encoding.rs
imor: Right, will do.

src/lib.rs
imor: What about the log_syntax_violation field? Shouldn't it be included?

imor:
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/encoding.rs, line 98 at r1](https://reviewable.io:443/reviews/servo/rust-url/335#-Kjs8CneImvzHu6G6ha0:-Kjwb5D3skJPZLBIh_sY:b-896fix) ([raw file](https://github.com/servo/rust-url/blob/01e70b9f69bf0b21246fcbd26e37a94719eda74d/src/encoding.rs#L98)):*
<details><summary><i>Previously, saghm (Saghm Rossi) wrote…</i></summary><blockquote>

Could this instead call [`name`](https://docs.rs/encoding/0.2.33/encoding/types/trait.Encoding.html#tymethod.name) on the wrapped encoding ref? I think it might be useful to be able to see the the type of the encoding if it's present.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/335)*

src/percent_encoding.rs
imor: Done.

SimonSapin:
Thanks!

@bors-servo r+

PR #340, opened 2017-05-14
Add examples Url methods #309

### LINKED ISSUE #309: Add examples to `Url` methods
These are missing examples:  Url::options, query, query_pairs, fragment, set_fragment, set_query, set_path, path_segments, set_ip_host, set_password, set_username.

hngnaig:
Fix #309 

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/340)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -320,6 +320,24 @@ impl Url {
     }
 
     /// Return a default `ParseOptions` that can fully configure the URL parser.
+    ///
+    /// # Examples
+    ///
+    /// Get default `ParseOptions`, then change base url
+    ///
+    /// ```rust
+    /// use url::Url;
+    /// # use url::ParseError;
+    /// # fn run() -> Result<(), ParseError> {
+    /// let options = Url::options();
+    /// let api = Url::parse("https://api.example.com")?;
+    /// let base_url = options.base_url(Some(&api));
+    /// let version_url = base_url.parse("version.json")?;
+    /// assert_eq!(version_url.as_str(), "https://api.example.com/version.json");
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
     pub fn options<'a>() -> ParseOptions<'a> {
         ParseOptions {
             base_url: None,
@@ -936,6 +954,25 @@ impl Url {
     /// For cannot-be-a-base URLs, this is an arbitrary string that doesn’t start with '/'.
     /// For other URLs, this starts with a '/' slash
     /// and continues with slash-separated path segments.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::{Url, ParseError};
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let url = Url::parse("https://example.com/api/versions?page=2")?;
+    /// assert_eq!(url.path(), "/api/versions");
+    ///
+    /// let url = Url::parse("https://example.com")?;
+    /// assert_eq!(url.path(), "/");
+    ///
+    /// let url = Url::parse("https://example.com/countries/việt nam")?;
+    /// assert_eq!(url.path(), "/countries/vi%E1%BB%87t%20nam");
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
     pub fn path(&self) -> &str {
         match (self.query_start, self.fragment_start) {
             (None, None) => self.slice(self.path_start..),
@@ -975,6 +1012,11 @@ impl Url {
     ///
     /// let url = Url::parse("data:text/plain,HelloWorld")?;
     /// assert!(url.path_segments().is_none());
+    ///
+    /// let url = Url::parse("https://example.com/countries/việt nam")?;
+    /// let mut path_segments = url.path_segments().ok_or_else(|| "cannot be base")?;
+    /// assert_eq!(path_segments.next(), Some("countries"));
+    /// assert_eq!(path_segments.next(), Some("vi%E1%BB%87t%20nam"));
     /// # Ok(())
     /// # }
     /// # run().unwrap();
@@ -989,6 +1031,29 @@ impl Url {
     }
 
     /// Return this URL’s query string, if any, as a percent-encoded ASCII string.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
+    /// fn run() -> Result<(), ParseError> {
+    /// let url = Url::parse("https://example.com/products?page=2")?;
+    /// let query = url.query();
+    /// assert_eq!(query, Some("page=2"));
+    ///
+    /// let url = Url::parse("https://example.com/products")?;
+    /// let query = url.query();
+    /// assert!(query.is_none());
+    ///
+    /// let url = Url::parse("https://example.com/?country=español")?;
+    /// let query = url.query();
+    /// assert_eq!(query, Some("country=espa%C3%B1ol"));
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
     pub fn query(&self) -> Option<&str> {
         match (self.query_start, self.fragment_start) {
             (None, _) => None,
@@ -1005,6 +1070,28 @@ impl Url {
 
     /// Parse the URL’s query string, if any, as `application/x-www-form-urlencoded`
     /// and return an iterator of (key, value) pairs.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use std::borrow::Cow;
+    ///
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let url = Url::parse("https://example.com/products?page=2&sort=desc")?;
+    /// let mut pairs = url.query_pairs();
+    ///
+    /// assert_eq!(pairs.count(), 2);
+    ///
+    /// assert_eq!(pairs.next(), Some((Cow::Borrowed("page"), Cow::Borrowed("2"))));
+    /// assert_eq!(pairs.next(), Some((Cow::Borrowed("sort"), Cow::Borrowed("desc"))));
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    ///
+
     #[inline]
     pub fn query_pairs(&self) -> form_urlencoded::Parse {
         form_urlencoded::parse(self.query().unwrap_or("").as_bytes())
@@ -1023,6 +1110,25 @@ impl Url {
     ///
     /// **Note:** the parser did *not* percent-encode this component,
     /// but the input may have been percent-encoded already.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let url = Url::parse("https://example.com/data.csv#row=4")?;
+    ///
+    /// assert_eq!(url.fragment(), Some("row=4"));
+    ///
+    /// let url = Url::parse("https://example.com/data.csv#cell=4,1-6,2")?;
+    ///
+    /// assert_eq!(url.fragment(), Some("cell=4,1-6,2"));
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
     pub fn fragment(&self) -> Option<&str> {
         self.fragment_start.map(|start| {
             debug_assert!(self.byte_at(start) == b'#');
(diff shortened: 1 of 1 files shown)


hngnaig:
cc @dtolnay and @SimonSapin

src/lib.rs
```diff
@@ -861,6 +886,20 @@ impl Url {
     }
 
     /// Return this URL’s query string, if any, as a percent-encoded ASCII string.
```
dtolnay: Let's show a third example where percent-encoding is necessary.

src/lib.rs
```diff
@@ -813,6 +824,20 @@ impl Url {
...
     /// and continues with slash-separated path segments.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let url = Url::parse("https://example.com/api/versions?page=2").unwrap();
+    ///
+    /// assert_eq!(url.path(), "/api/versions");
```
dtolnay: I think this would be more readable at a glance if the asserts were grouped with the previous line:

```rust
use url::Url;

let url = Url::parse("https://example.com/api/versions?page=2").unwrap();
assert_eq!(url.path(), "/api/versions");

let url = Url::parse("https://example.com").unwrap();
assert_eq!(url.path(), "/");
```

src/lib.rs
```diff
@@ -281,6 +281,17 @@ impl Url {
...
 
     /// Return a default `ParseOptions` that can fully configure the URL parser.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let parse_options = Url::options();
+    /// let base = parse_options.parse("https://example.net/api/").unwrap();
```
dtolnay: Please come up with an example that requires ParseOptions. This one would just use `Url::parse`.

src/lib.rs
```diff
@@ -901,6 +955,23 @@ impl Url {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com/data.csv").unwrap();
+    /// {
+    ///     url.set_fragment(Some("cell=4,1-6,2"));
+    /// }
+    /// assert_eq!(url.fragment(), Some("cell=4,1-6,2"));
```
dtolnay: Please add an assert on `url.as_str()` after each of these to show what the full url looks like.

src/lib.rs
```diff
@@ -936,6 +1007,18 @@ impl Url {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com/products").unwrap();
+    /// {
+    ///     url.set_query(Some("page=2"));
+    /// }
+    /// assert_eq!(url.query(), Some("page=2"));
```
dtolnay: Please add an assert on `url.as_str()` after this.

src/lib.rs
```diff
@@ -1011,6 +1094,24 @@ impl Url {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com").unwrap();
+    /// {
+    ///     url.set_path("api/comments");
+    /// }
+    /// assert_eq!(url.path(), "/api/comments");
```
dtolnay: Please add an assert on `url.as_str()` after each of these.

src/lib.rs
```diff
@@ -1011,6 +1094,24 @@ impl Url {
...
 
     /// Change this URL’s path.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com").unwrap();
+    /// {
```
dtolnay: This nesting is not needed.

src/lib.rs
```diff
@@ -936,6 +1007,18 @@ impl Url {
...
 
     /// Change this URL’s query string.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com/products").unwrap();
+    /// {
```
dtolnay: This nesting is not needed.

src/lib.rs
```diff
@@ -901,6 +955,23 @@ impl Url {
...
 
     /// Change this URL’s fragment identifier.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("https://example.com/data.csv").unwrap();
+    /// {
```
dtolnay: This nesting is not needed.

src/lib.rs
```diff
@@ -1255,6 +1356,22 @@ impl Url {
...
     ///
     /// Compared to `Url::set_host`, this skips the host parser.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("http://example.com").unwrap();
+    /// {
```
dtolnay: This nesting is not needed.

src/lib.rs
```diff
@@ -1255,6 +1356,22 @@ impl Url {
...
+    ///     url.set_ip_host("127.0.0.1".parse().unwrap());
+    /// }
+    /// assert_eq!(url.host_str(), Some("127.0.0.1"));
```
dtolnay: Please add an assert on `url.as_str()` after this.

src/lib.rs
```diff
@@ -1255,6 +1356,22 @@ impl Url {
...
+    ///
+    /// let mut url = Url::parse("mailto:rms@example.com").unwrap();
+    /// let result = url.set_ip_host("127.0.0.1".parse().unwrap());
+    /// assert!(result.is_err());
```
dtolnay: Since the error type isn't meaningful, please add a brief comment here reminding why this is an error.

src/lib.rs
```diff
@@ -1320,6 +1456,20 @@ impl Url {
...
     ///
     /// If this URL is cannot-be-a-base or does not have a host, do nothing and return `Err`.
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let mut url = Url::parse("mailto:rmz@example.com").unwrap();
+    /// let result = url.set_username("user1");
+    /// assert!(result.is_err());
```
dtolnay: Please add a comment saying why this is an error.

src/lib.rs
```diff
@@ -1320,6 +1456,20 @@ impl Url {
...
+    ///
+    /// let mut url = Url::parse("ftp://:secre1@example.com").unwrap();
+    /// let result = url.set_username("user1");
+    /// assert!(result.is_ok());
+    /// assert_eq!(url.username(), "user1");
```
dtolnay: Please add an assert on `url.as_str()`.

dtolnay:
Nice work. One more comment that applies to all of these: the examples should show realistic error handling. In real code you probably would not want any of these `unwrap()` calls to crash your application. Use `?` instead - see #323 for some examples.

brson:
Ping @hngnaig

hngnaig:
I just fix code, but i missing test for method `path`, but i will update soon, let me know if any wrong. Thanks @dtolnay.

hngnaig:
I need help about function `path`, as doc, method will return `&str`, that percent-encoded ASCII string. But i assert function shown as bellow, that no percent-encoded.
```rust
use url::Url;
let url = Url::parse("https://example.com/foo/bar/baz").unwrap();
assert_eq!(url.path(), "/foo/bar/baz");
```
I didn't really understand what it is. I think i should assert as bellow.

```rust
use url::Url;
use url::percent_encoding;
let url = Url::parse("https://example.com/foo%21bar%21baz").unwrap();
let encoding_path = percent_encoding::percent_encode(b"/foo/bar/baz", PATH_SEGMENT_ENCODE_SET);
let path_str = encoding_path.collect::<String>();
assert_eq!(url.path(), path_str);
```
Thanks for help me.

src/lib.rs
```diff
@@ -281,6 +281,20 @@ impl Url {
     }
 
     /// Return a default `ParseOptions` that can fully configure the URL parser.
+    ///
+    /// # Examples
+    ///
+    /// Get default `ParseOption`, then change base url
```
SimonSapin: `ParseOptions` is plural. (`s` missing.)

src/lib.rs
```diff
@@ -281,6 +281,20 @@ impl Url {
...
+    ///
+    /// ```rust
+    /// use url::Url;
+    ///
+    /// let options = Url::options();
+    /// let api = Url::parse("https://api.example.com").ok();
+    /// let base_url = options.base_url(api.as_ref());
```
SimonSapin: Please replace `.ok()` with `.unwrap()` and `api.as_ref()` with `Some(&api)`, to show more typical usage.

SimonSapin:
@hngnaig Your examples for `path` happen not to use percent-encoding. If you’d like to add an example that does, try something like:

```rust
let url = Url::parse("https://example.com/Été hiver/").unwrap();
assert_eq!(url.path(), "/%C3%89t%C3%A9%20hiver");
```

No need to use `percent_encode` explicitly in that example.

src/lib.rs
Enet4: Actually, according to the other examples (and as highlighted by @dtolnay), we should not be using `unwrap()` either. Let's use `?` instead, thus forcing users to handle the errors properly.

SimonSapin:
Your `Url::path` example looks good, thanks!

However, I’m not sure how it happened, but this PR is now adding 115 new commits. Please squash them into one commit. (Unless you think it’s valuable to split your changes into multiple commits, but they should not duplicate unrelated changes.)

----

As an aside, I recommend this git alias to visualize the commit graph, I use it a lot. In `~/.gitconfig`, add:

```ini
[alias]
graph = log --oneline --graph --all --decorate
```

Now the `git graph` command shows branch in your repository.

---

Now, the usual way to squash is `git rebase -i <some-base-commit>` (`-i` is also `--interactive`) and using the `fixup` instruction in the editor that comes up. But this might be tricky with so many commits and conflicts, so try this instead:

* Check out your branch if it isn’t already.
* Use `git graph` above or some other way to find which commit of the `master` branch your branch is up-to-date with. As far as I can tell this should be d19d5d0, but double-check.
* Run `git reset --soft <that-commit>`, for example `git reset --soft d19d5d0`. This changes your branch to be "at" that commit, but keeps all of the checked out files intact.
* Run `git commit`. Unfortunately with this method you will need to rewrite the commit message.
* Run `git push -f`. `-f` is also `--force`, it will overwrite the remote branch (in your GitHub fork) regardless of what was there before, even if this effectively removes commits. This should be use very carefully, but it is needed in this case.

hngnaig:
Thanks @SimonSapin for helping me. I squashed all commit, and reset them at the last commit [d19d5d](https://github.com/servo/rust-url/commit/d19d5d0b23b3688b00f1c2748520ba3f90e66d3a). However, there are still errors, I'm trying to check the code and fix the errors.

SimonSapin:
I can take over the git wrangling if you’d like. Let me know.

hngnaig:
💯 After trying to fix the error, pull request has been completely done @SimonSapin 💯

KiChjang:
@hngnaig Unfortunately your PR still contains merge commits and isn't properly squashed.

SimonSapin:
It also has unrelated whitespace changes in `fuzz/fuzzers/parse.rs` and `percent_encoding/lib.rs`

hngnaig:
cc @KiChjang, @SimonSapin i squashed all commit, and remove all whitespace in `fuzz/fuzzers/parse.rs` and `percent_encoding/lib.rs`

src/lib.rs
```diff
@@ -1427,6 +1594,38 @@ impl Url {
...
     /// Compared to `Url::set_host`, this skips the host parser.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use url::{Url, ParseError};
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("http://example.com")?;
+    /// url.set_ip_host("127.0.0.1".parse().unwrap());
```
Enet4: Is this `unwrap` acceptable here? If not, we could change the return type to `Result<(), Box<Error>>` and map errors into a `&'static str`.

SimonSapin:
Thanks you!

@bors-servo r+

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[src/lib.rs, line 1605 at r3](https://reviewable.io:443/reviews/servo/rust-url/340#-Kk4vvtjd4u1WNwTcXoT-r3-1605:-KnAVWRIr-kRcDPAFQtz:b9lufgo) ([raw file](https://github.com/servo/rust-url/blob/dd1f2328a20885bd8149d54f8ac8d47501e683a1/src/lib.rs#L1605)):*
<details><summary><i>Previously, Enet4 (Eduardo Pinho) wrote…</i></summary><blockquote>

Is this `unwrap` acceptable here (I'm asking because `127.0.0.1` is guaranteed to work)? If not, we could change the return type to `Result<(), Box<Error>>` and map errors into a `&'static str`.
</blockquote></details>

In normal code I think it’s fine to use `.unwrap()` when we know that it won’t panic (unless some code is buggy), for example when the input is fixed like here. In the spirit of #312 however we should probably avoid it in examples. Let’s do this later in a follow-up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/340#-:-KnAVxEZGPXyzy1RcSW0:b-wct6jx)*

src/lib.rs
hngnaig: I think it's acceptable because 127.0.0.1 is valid static ip address. But it will be much better if change return type to `Result<(), Box<Error>>`.

```
fn run() -> Result<(), Box<Error>> {
  url.set_ip_host("127.0.0.1".parse()?);
}
```

PR #343, opened 2017-05-17
[idna] Upgrade unicode-bidi to 0.3.0

behnam:
For the needs of `idna`, `unicode-bidi:0.3.0` is backwards compatible,
but this upgrade is needed to that `servo` can upgrade to the new
version.

Servo upgrade: https://github.com/servo/servo/pull/16779

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/343)

### DIFF
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -19,6 +19,6 @@ rustc-test = "0.1"
 rustc-serialize = "0.3"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


behnam:
This is a simple version bump, blocking https://github.com/servo/servo/pull/16779.

r? @mbrubeck

PR #345, opened 2017-05-21
[WIP] Modify examples to not call `unwrap`

luisbg:
This is for issue #312 

Just fixing one case to check this is what the issue is about and that I understand the intended fix.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/345)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -267,13 +267,17 @@ impl Url {
     /// ```rust
     /// use url::Url;
     ///
-    /// let base = Url::parse("https://example.net/a/b.html").unwrap();
-    /// let url = base.join("c.png").unwrap();
+    /// fn foo() -> Result<(), url::ParseError> {
+    /// let base = Url::parse("https://example.net/a/b.html")?;
+    /// let url = base.join("c.png")?;
     /// assert_eq!(url.as_str(), "https://example.net/a/c.png");  // Not /a/b.html/c.png
     ///
-    /// let base = Url::parse("https://example.net/a/b/").unwrap();
-    /// let url = base.join("c.png").unwrap();
+    /// let base = Url::parse("https://example.net/a/b/")?;
+    /// let url = base.join("c.png")?;
     /// assert_eq!(url.as_str(), "https://example.net/a/b/c.png");
+    ///
+    /// Ok(())
+    /// }
     /// ```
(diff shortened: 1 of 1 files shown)


behnam:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/lib.rs, line 270 at r1](https://reviewable.io:443/reviews/servo/rust-url/345#-Kkmf020Z-5VUBuS55Qk:-Kkmf020Z-5VUBuS55Ql:bv9z4k2) ([raw file](https://github.com/servo/rust-url/blob/752423ae35672ffeea31e464273a0c2cacecf1d3/src/lib.rs#L270)):*
> ```Rust
>     /// use url::Url;
>     ///
>     /// fn foo() -> Result<(), url::ParseError> {
> ```

If you like to keep the function definition part of the example, then the content of the block should be indented.

But, rustdoc has a nice way of letting you hide these *execution* details, using `# ` at the beginning of a line.

For example, this part would be:
```
    /// # fn foo() -> Result<(), url::ParseError> {
    /// ...
    /// # Ok(())
    /// # }
```

More here: https://doc.rust-lang.org/book/documentation.html#documenting-macros

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/345)*

behnam:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/345)*

luisbg:
Cool!

I will fix this tomorrow, probably. Currently traveling for work meeting so might not have the time to get around to it.

mgeisler:
I believe you should also include a call to the `foo()` function -- otherwise the code in question is only compiled, but not actually executed.

luisbg:
I am looking at this again. Will report soon. Sorry for the delay.

Enet4:
#312 has been addressed in #323 (see the latest [Url#join](https://github.com/servo/rust-url/blob/master/src/lib.rs#L292-L305) for instance). Are there any new examples with `unwrap` since then?

SimonSapin:
This is done in #323. Thanks for the PR anyway!

PR #346, opened 2017-05-22
Build with Rust 1.13 on Travis

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/346)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ rust:
   - nightly
   - beta
   - stable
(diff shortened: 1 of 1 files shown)


behnam:
I have borken builds for `<1.17` in `unicode-bidi:0.3.0`, which is the new dependency for `idna`. I'm working on a fix now.

See: https://travis-ci.org/servo/servo/jobs/234906288 and https://github.com/servo/servo/pull/16779

PR #347, opened 2017-05-22
move percent_encoding to its own crate

seanmonstar:
- I arbitrarily chose `percent_encode`, but I could update it to be `percent_encoding` if preferred.
- The macro `define_encode_set` duplicated, to maintain backwards compatibility for `url`.
- I set it 1.0.0 since this is a public dependency of `url`, and so it can't make API breaking changes easily without affecting `url`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/347)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,6 +98,7 @@ assert_eq!(css_url.as_str(), "http://servo.github.io/rust-url/main.css")
 #[cfg(feature="heapsize")] #[macro_use] extern crate heapsize;
 
 pub extern crate idna;
+pub extern crate percent_encoding;
 
 use encoding::EncodingOverride;
 #[cfg(feature = "heapsize")] use heapsize::HeapSizeOf;
@@ -131,7 +132,6 @@ mod parser;
 mod slicing;
 
 pub mod form_urlencoded;
-pub mod percent_encoding;
 pub mod quirks;
 
 /// A parsed URL record.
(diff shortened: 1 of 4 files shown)


url_serde/src/lib.rs
```diff
@@ -69,6 +69,10 @@ use std::ops::{Deref, DerefMut};
 use std::str;
 use url::{Url, Host};
 
+define_encode_set! {
+    /// This encode set is used in the URL parser for query strings.
+    pub QUERY_ENCODE_SET = [::url::percent_encoding::SIMPLE_ENCODE_SET] | {' ', '"', '#', '<', '>'}
+}
```
nox: Why does this need to be done here?

src/lib.rs
```diff
@@ -1881,3 +1881,48 @@ impl<'a> Drop for UrlQuery<'a> {
...
+/// define_encode_set! {
+///     /// This encode set is used in the URL parser for query strings.
+///     pub QUERY_ENCODE_SET = [SIMPLE_ENCODE_SET] | {' ', '"', '#', '<', '>'}
+/// }
+/// # fn main() {
+/// assert_eq!(utf8_percent_encode("foo bar", QUERY_ENCODE_SET).collect::<String>(), "foo%20bar");
+/// # }
+/// ```
+#[macro_export]
+macro_rules! define_encode_set {
```
nox: Isn't there a way to reexport a macro from a different crate?

percent_encode/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "percent_encode"
```
nox: I'll let @SimonSapin decide on the crate name.

nox: > 11:10 &lt;•SimonSapin> nox: if we’re splitting hairs, I think "encoding" is the general mechanism and the two specific (opposite) operations are "encode" and "decode"

Let's name it `percent_encoding`.

SimonSapin:
> I arbitrarily chose percent_encode, but I could update it to be percent_encoding if preferred.

I’m kind of splitting hairs, but I’d prefer `percent_encoding`. I think "encoding" refers to the general mechanism, which includes two opposite operations: "encode" and "decode".

src/lib.rs
SimonSapin: IIRC the only way on current stable Rust is `pub use some_crate::*;`, which of course also re-exports other things.

url_serde/src/lib.rs
seanmonstar: Um, I have no idea why this is here...

src/lib.rs
seanmonstar: There's `#[macro_reexport]`, which is unstable, and there's a the glob trick, which I'm still not certain is on purpose, but still, it re-exports too many things, so duplicating the macro seems required...

seanmonstar:
Renamed to `percent_encoding`, stray code in `url_serde` removed.

seanmonstar:
@SimonSapin or @nox, r?

src/lib.rs
seanmonstar: @nox you said in IRC you prefer the glob trick. To reiterate, that requires a `pub use percent_encoding::*` inside the `url` root. That would suddenly make all the things normally in that module/crate pollute the url namespace. Suddenly there is `url::SIMPLE_ENCODE_SET`, besides `url::percent_encoding::SIMPLE_ENCODE_SET` and friends.

nox: Can't you just put the macro in a submodule of `percent_encoding` and glob import that module from both `percent_encoding` and `url`?

seanmonstar: That's possible, but puts another public (useless) module in the `percent_encoding` crate. Of all the possible options, I find just redefining the macro in `url` to be the best.

nox: Given both crates are in the same repository, IMO it is better to just make a `#[doc(hidden)]` module and avoid the duplication of the macro, but I'll let @SimonSapin the final say.

seanmonstar: When macros 2.0 arrive (already major pieces in nightly!), they can be named explicit, so it could eventually just become in url `pub use percent_encoding::define_encode_set;`.

nox: Unfortunately I don't think these two crates can really afford to rely on hot new stuff, given how central it is to the ecosystem.

seanmonstar: We can cross that bridge when we get there, but this crate has already made upgrades to 1.15, and other than using the newest the day after, it seems to be fine to upgrade to newer version.

SimonSapin:
Landed in #362. Thanks!

PR #355, opened 2017-06-08
Documentation and examples enhancements of percent encoding

### LINKED ISSUE #298: Document the percent_encoding module
This module encapsulates a crate-specific pattern that could be explained better.

Some things that might clarify my questions about this module:

- Module level explanation of what "percent encoding" is and the role of `EncodeSet`
- Module level examples of creating and using encode sets
- Do users ever need to create their own encode sets?
- When do users need to use this module?
- Examples on the free functions and EncodeSet
- Documentation of which characters are in each predefined EncodeSet

AndyGauge:
Hopefully fixes #298 
Made some adjustments to align with @SimonSapin recommendations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/355)

### DIFF
--- a/src/percent_encoding.rs
+++ b/src/percent_encoding.rs
@@ -6,6 +6,39 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! URLs use special chacters to indicate the parts of the request.  For example, a forward slash
+//! indicates a path.  In order for that charcter to exist outside of a path separator, that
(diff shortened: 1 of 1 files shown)


AndyGauge:
The appveyor check seems to be a system problem.  

> `error: could not exec the linker `gcc`: The system cannot find the file specified. (os error 2)`

Issue #353, opened 2017-06-05
Percent encode set that encodes $ as %24?


golddranks:
I'm using `rust-url` to percent encode the username and password in my `postgresql://username:password@domain/dbname` style URL. However, my password contains a special character `$`, and no one of the percent encoding sets provided by this crate seems to encode that.

`$` is considered a "Reserved Character" ( https://tools.ietf.org/html/rfc3986#section-2.2 ), so I think it should be reasonable to encode it when it isn't used as a delimiter. I don't know of any current use of it as a delimiter in URLs, but the `libpq` that parses the URL complains about it, and accepts the password only as fully encoded.

Maybe there could be a `STRICT_ENCODE_SET` that would encode each of the reserved characters:

      reserved    = gen-delims / sub-delims

      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"

      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

SimonSapin:
Your crate can define its own encode set: https://docs.rs/url/1.5.0/url/macro.define_encode_set.html

PR #358, opened 2017-06-12
Added RESERVED_ENCODE_SET & STRICT_ENCODE_SET & more detailed docs.

golddranks:
Added RESERVED_ENCODE_SET that corresponds to the set of reserved URL delimeter characters, and STRICT_ENCODE_SET that encodes everything expect unreserved characters. (As defined in the [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.2)) Added more detailed descriptions to the existings sets.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/358)

### DIFF
--- a/src/percent_encoding.rs
+++ b/src/percent_encoding.rs
@@ -77,6 +77,9 @@ macro_rules! define_encode_set {
 }
 
 /// This encode set is used for the path of cannot-be-a-base URLs.
+/// Contains all of the ASCII control character range, DEL and the 8-bit range outside of ASCII –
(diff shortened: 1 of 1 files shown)


golddranks:
Resolves https://github.com/servo/rust-url/issues/353

SimonSapin:
I’m opposed to adding new encode sets just because they exist in some RFC. This is why the `define_encode_set!` macro exists: you can define them in your own crate if you need them.

Closing in favor of #355 for the rest of the changes.

PR #361, opened 2017-06-13
Percent encoding docs

### LINKED ISSUE #298: Document the percent_encoding module

SimonSapin:
Fixes #298, fixes #355.

This is a rebase of #355 with some additional changes. Original work by @AndyGauge.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/361)

### DIFF
--- a/rust-url-todo
+++ b/rust-url-todo
@@ -1,0 +1,14 @@
+* standalone path parsing?
+* Test setters
+  * Test trim C0/space
+  * Test remove tab & newline
+
+
+
+#[test]
+fn test_path_segments() {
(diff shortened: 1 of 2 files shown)


brson:
Thanks @SimonSapin !

PR #364, opened 2017-06-17
idna: use standard binary_search_by

froydnj:
Doing this is arguably nicer and avoids an edge case when searching for
the last element in the table.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/364)


SimonSapin:
@bors-servo r+

Thanks!

PR #351, opened 2017-05-29
[idna] Update data to Unicode 10.0 and fix logic

behnam:
* Change the behavior of ToASCII step to match the spec and prevent failures on some cases when a domain name starts with leading dots (FULL STOPs), as requested in #166. (Another attempt on https://github.com/servo/rust-url/pull/337 and https://github.com/servo/rust-url/pull/171)

* Update `IdnaTest.txt` file to UCD 10.0 and fix Validation Rules, specially Bidi Rules, for the tests to pass.

* Add TODO marks for new flags introduced in Unicode 10.0 version of UTS#46. (http://www.unicode.org/reports/tr46/proposed.html)

* Add integration test for `rust-url` crate for the new behavior.

Fix https://github.com/servo/rust-url/issues/166

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/351)

PR #365, opened 2017-06-18
more idna data slimming

froydnj:
We can do a better job of packing the uts46 data:

* We can merge identically-mapped entries that don't have an associated string slice.  This saves ~10% space.
* We can make slices smaller and pack them into `Mapping` better, which saves 25% space on 64-bit platforms.  I think it might save half that on 32-bit platforms, but I didn't check.

Together these are good for ~42KB of space savings on a 64-bit platform.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/365)


SimonSapin:
Looks great, thanks!

This does not include an version number increment, but #351 does so if you don’t mind waiting a bit I’ll publish both together to avoid merge conflicts.

@bors-servo r+

froydnj:
> This does not include an version number increment, but #351 does so if you don’t mind waiting a bit I’ll publish both together to avoid merge conflicts.

That'd be great, thank you!

PR #366, opened 2017-06-18
set_scheme should fail if url has host or port

### LINKED ISSUE #278: Setting the scheme to "file" should fail if there's a host or port
The test case:
```rust
    let mut url: Url = "http://localhost:6767/foo/bar".parse().unwrap();
    let result = url.set_scheme("file");
    assert_eq!(url.to_string(), "http://localhost:6767/foo/bar");
    assert_eq!(result, Err(()));
```
should pass, instead it fails with:
```
   (left == right)` (left: `"file://localhost:6767/foo/bar"`, right: `"http://localhost:6767/foo/bar"`)
```
The resulting URL has a `file` scheme, and a host/port, so serialization fails.

tmccombs:
Fixes #278

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/366)

### DIFF
--- a/tests/unit.rs
+++ b/tests/unit.rs
@@ -327,6 +327,13 @@ fn issue_241() {
 }
 
 #[test]
+fn issue_278() {
+    let mut url: Url = "http://localhost:6767/foo/bar".parse().unwrap();
+    assert_eq!(url.to_string(), "http://localhost:6767/foo/bar");
+    assert_eq!(url.set_scheme("file"), Err(()));
+}
+
+#[test]
 /// https://github.com/servo/rust-url/issues/222
 fn append_trailing_slash() {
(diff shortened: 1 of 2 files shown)


tests/unit.rs
```diff
@@ -327,6 +327,13 @@ fn issue_241() {
 }
 
 #[test]
+fn issue_278() {
```
tmccombs: I'm not sure if this is the best place to put the unit test.

SimonSapin:
Thank you for the PR. The code looks ok at first glance, but this may not be the right change to. Let’s discuss in #278.

djc:
The test proposed in #278 seems to pass now, so I don't think we need this anymore. Let me know if you have a different opinion!

PR #370, opened 2017-06-25
Add HeapSizeOf impl for ParseError

KiChjang:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/370)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "1.5.0"
+version = "1.5.1"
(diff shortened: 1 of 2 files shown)


SimonSapin:
* Please use the `known_heap_size!` macro when possible
* This does not compile. You need `#[…]` syntax to add an attribute to the next item, rather than `#![…]` which adds an attribute to the thing that contains such syntax.

KiChjang:
Ack, that was embarrassing. Fixed.

SimonSapin:
Thanks!

@bors-servo r+

KiChjang:
Published this as 1.5.1.

PR #372, opened 2017-07-01
Update rustc-test to 0.2, fix #371

### LINKED ISSUE #371: `tests/unit.rs` broken on nightly builds
Travis: https://travis-ci.org/servo/rust-url/jobs/248944414

Example error:

```
error[E0560]: struct `__test::test::TestDesc` has no field named `allow_fail`
  --> tests/unit.rs:20:1
   |
20 | / fn size() {
21 | |     use std::mem::size_of;
22 | |     assert_eq!(size_of::<Url>(), size_of::<Option<Url>>());
23 | | }
   | |_^ `__test::test::TestDesc` does not have this field
```

I think it's caused by `tests/unit.rs` having test harness enabled, and the test internals having changed and not conflicting `rustc-test` package used here.

I'm not sure what's the best option for a fix:
* Disabling harness for `unit.rs` and copying in boilerplate to get the tests running; or
* Updating `rustc-test` to make it compatible.

What do you think, @SimonSapin ?

SimonSapin:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/372)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ harness = false
 test = false
 
 [dev-dependencies]
-rustc-test = "0.1"
+rustc-test = "0.2"
 rustc-serialize = "0.3"
 serde_json = ">=0.6.1, <0.9"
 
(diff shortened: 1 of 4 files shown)


SimonSapin:
Forgot to also update `idna/Cargo.toml`

@bors-servo r=nox

Issue #373, opened 2017-07-03
Parsing the hostname "xn--" returns an empty hostname, but doesn't throw an error when a hostname is required


MJDSys:
Basically, ```assert_eq!(Url::parse("ftp://xn--"), Url::parse("ftp:///"));``` fails.  It seems even though idna parsing produces an empty host, it's still not considered empty and a valid Url is constructed instead of returning an error.

Is this the expected behaviour?  Or should the empty host check not take into account idna processing.

rmisev:
It was URL spec issue. Now fixed in https://github.com/whatwg/url/pull/497

djc:
This should be fixed by 6b05499d3f42f3522fb549838b80148a4cbdce23.

PR #374, opened 2017-07-03
[idna] Add unit tests for punycode prefix edge cases

behnam:
Helps with answering question asked in
https://github.com/servo/rust-url/issues/373

The test invariants here are based on and checked against:

* Processing Step 4 of UTS#46
  (<http://www.unicode.org/reports/tr46/#ProcessingStepPunycode>)

* 6.2 Decoding procedure of Punycode RFC.
  (<https://ietf.org/rfc/rfc3492.txt>)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/374)

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -357,7 +357,7 @@ enum Error {
 ///
 /// This is opaque for now, only indicating the presence of at least one error.
 /// More details may be exposed in the future.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct Errors(Vec<Error>);
 
 /// http://www.unicode.org/reports/tr46/#ToASCII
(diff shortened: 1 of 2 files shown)


djc:
This would be good to have, though it'll have to be rebased on current master.

djc:
I rebased this, but the tests for `config.to_ascii("xn-----")` no longer pass. I'll investigate further.

annevk:
The reason they don't pass (i.e., no longer expect an error) is probably because CheckHyphens is set to false. Having said that, it's pretty clear the URL Standard algorithm for IDNA has issues as well as browsers are not fully aligned with it: https://github.com/whatwg/url/labels/topic%3A%20idna.

djc:
@annevk thanks for the hint! I've rebased this as 4c1be65d0ace810e8cc2b5fca745e8002b73698e and merged it. @behnam thanks for working on this.

PR #377, opened 2017-07-05
percent-encoding doesn't require any dev-dependencies

ignatenkobrain:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/377)


SimonSapin:
Good catch, thanks!

@bors-servo r+

Issue #314, opened 2017-05-04
Modify docs to put error conditions into `Errors` sections


brson:
Per rust documentation RFC.

llompartg:
Hey brson, first time contributing and really new to rust. 

I would be happy to help out with this, is this still open? Also I'm afraid I'll need a bit of more info as per which RFC you mean and what exactly the needed changes are.

Thanks

dtolnay:
@llompartg thanks, this is still open. He is referring to [this guideline](https://github.com/brson/rust-api-guidelines#c-error-doc).

brson:
Awesome @llompartg! @dtolnay has the link.

Basically, every method that returns errors, instead of the error condition being described in the main body of the docs (or not described at all) there should be a section like

```
# Errors

Returns an error when...
```

rap2hpoutre:
@llompartg I have just submitted a PR. If you still want to work on this issue, I can close it! (since there is no activity for a while now, so I thought I can submit something). As you want!

brson:
@rap2hpoutre looks like your pr was merged. Thanks!

PR #379, opened 2017-07-10
Add error documentation on parse function

rap2hpoutre:
See: https://github.com/servo/rust-url/issues/314

This is just a try, feel free to close or comment if it's not good. I just submit one error documentation because I want to clearly understand what's the pattern before submitting more documentation: I still have to learn, so I prefer just trying on one function for now (but I can help after that)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/379)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,6 +252,13 @@ impl Url {
     /// # }
     /// # run().unwrap();
     /// ```
+    /// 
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

Issue #363, opened 2017-06-14
PathSegmentsMut has no finish method like UrlQuery


seanmonstar:
> When building a URL, it would be convenient to be able to `finish` mutating path segments like one currently can for query pairs.
>
> ```rust
> // will NOT compile
> let mut url = Url::parse("...");
> url
>     .path_segments_mut().unwrap()
>         // ...
>     .query_pairs_mut()
>         // ...
>         .finish()
>     // ...
> ```
>
> Instead the query pairs must be mutated first, since this 'sub builder' _can_ be `finish`ed, returning a `Url`. Which (though slightly counter-intuitive) is fine, it's just a slightly jarring restriction.
> 
> It also means the mutable variable is necessary; if it behaved like `query_pairs_mut` then this could all be on one chain which could be passed directly (or via an immutable variable) into a `Client` method.

From https://github.com/seanmonstar/reqwest/issues/150
/cc @OJFord

TitiAlone:
Would like to work on this, but don't really know how to implement properly, should I implement Target for PathSegmentsMut or create a specific finish() method for PathSegmentsMut, which will output an Url?

SimonSapin:
@TitiAlone No, the `Target` trait is a hack that I wish weren’t part of the public API. It allows using `application/x-www-form-urlencoded` serialization without a `Url`, which does not apply to `PathSegmentsMut`. You can add an inherent `finish` method that returns `&'a mut Url` (not `Url`).

TitiAlone:
Just worked on this today and here's the code I made (included the doc because it's important).

```rust
/// Returns the URL reference the PathSegmentsMut was constructed with.
///
/// Example:
///
/// ```rust
/// use url::Url;
/// # use std::error::Error;
///
/// # fn run() -> Result<(), Box<Error>> {
/// let mut url = Url::parse("https://github.com/servo/rust-url/issues/363")?;
/// url.path_segments_mut()
///    .map_err(|_| "cannot be base")?
///    .clear()
///    .finish()
///    .query_pairs_mut()
///    // ...
///    .finish();
/// assert_eq!(url.as_str(), "https://github.com/");
/// # Ok(())
/// # }
/// # run().unwrap();
/// ```
pub fn finish(&mut self) -> &mut ::Url {
    self.url
}
```

This makes an error, not on `cargo build` but on `cargo test`, returning this:

```
---- src\path_segments.rs - path_segments::PathSegmentsMut<'a>::finish (line 222) stdout ----
[...]

PR #380, opened 2017-07-14
Add PathSegmentsMut::finish

LikaKavkasidze:
Resolved the issue #363, added some documentation and a unit test for what the issue author wanted to do.
I was able to resolve this issue only by changing &mut PathSegmentsMut to mut PathSegmentsMut, which is quite a huge change.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/380)


SimonSapin:
Thank you for the PR. The signature change for existing method is a breaking change and can not be done in rust-url 1.x. I’ve added this to https://github.com/servo/rust-url/issues/463.

TitiAlone:
This PR is from 2017, I thought you didn't needed theses changes anymore. Do I need to do something to make it compatible with version 2.0?

Issue #319, opened 2017-05-04
Tracking issue for crate evaluation


brson:
- [x] [Document the percent_encoding module](https://github.com/servo/rust-url/issues/298)
- [x] [Implement `Display` for `Host` and `HostAndPort`](https://github.com/servo/rust-url/issues/300)
- [x] [Implement `Default` for `ParseOptions` and `ParseOptions::new`](https://github.com/servo/rust-url/issues/301)
- [x] [Implement `Hash` for `Origin`](https://github.com/servo/rust-url/issues/302)
- [ ] [Move serde 1.0 support in-crate](https://github.com/servo/rust-url/issues/304)
- [x] [Implement Debug for many types](https://github.com/servo/rust-url/issues/305)
- [x] [Modify `define_encode_set` to support private definitions](https://github.com/servo/rust-url/issues/307)
- [x] [Add examples to `ParseOptions`](https://github.com/servo/rust-url/issues/308)
- [x] [Add examples to `Url` methods](https://github.com/servo/rust-url/issues/309)
- [x] [Improvements to `Origin` docs](https://github.com/servo/rust-url/issues/310)
- [x] [Better documentation for quirks module](https://github.com/servo/rust-url/issues/311)
- [x] [Modify examples to not call `unwrap`](https://github.com/servo/rust-url/issues/312)
- [x] [Add the `html_root_url` attribute to the crate](https://github.com/servo/rust-url/issues/313)
- [x] [Modify docs to put error conditions into `Errors` sections](https://github.com/servo/rust-url/issues/314)
- [x] [Add CI badges to Cargo.toml](https://github.com/servo/rust-url/issues/315)
- [x] [Add categories to Cargo.toml](https://github.com/servo/rust-url/issues/316)
- [x] [Bump heapsize to a relicensed version](https://github.com/servo/rust-url/issues/317)
- [x] [Explain the term "fragment" in `Url::fragment`](https://github.com/servo/rust-url/issues/318)
- [ ] Minor bump and release a new version (after all else)

This is the tracking issue for the [crate evaluation](https://public.etherpad-mozilla.org/p/rust-crate-eval-url) performed by the libs team this week.

cc @rust-lang/libs

theduke:
May be relevant: https://github.com/servo/rust-url/issues/333

theduke:
Probably relevant: #338

rap2hpoutre:
[...]

PR #382, opened 2017-07-17
add more documentation for errors

rap2hpoutre:
More documentations for errors.  See https://github.com/servo/rust-url/issues/319 and https://github.com/servo/rust-url/pull/379

_Side question: I don't know how to document errors when a function returns `Result<(), ()>` so I did not document that part. I would be happy to see one example._

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/382)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -281,6 +281,13 @@ impl Url {
     /// # }
     /// # run().unwrap();
     /// ```
+    ///
+    /// # Errors
+    ///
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

----

> I don't know how to document errors when a function returns `Result<(), ()>`

Pretty much the same, except saying "returns an error" or "returns `Err`" without being more specific.

PR #385, opened 2017-08-01
Upgrade idna dependency to 0.1.4 and publish rust-url 1.5.2

valenting:
We would like to update the dependency in Gecko to get the latest idna table.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/385)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "1.5.1"
+version = "1.5.2"
 authors = ["The rust-url developers"]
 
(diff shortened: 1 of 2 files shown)


valenting:
@SimonSapin @Manishearth What are the next steps here?

SimonSapin:
You don’t need this to get that, since Cargo considers 0.1.4 to be semver-compatible with 0.1.0. Try running this in Gecko:

```sh
(cd toolkit/library/rust && cargo update -p idna)
(cd toolkit/library/gtest/rust && cargo update -p idna)
./mach vendor rust
```

valenting:
Oh, I didn't know that. Thanks for the info 😄

PR #386, opened 2017-08-06
Implement Serialize and Deserialize on ParseError

KiChjang:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/386)

### DIFF
--- a/url_serde/src/lib.rs
+++ b/url_serde/src/lib.rs
@@ -61,13 +61,14 @@ extern crate serde;
 extern crate url;
 
 use serde::{Deserialize, Serialize, Serializer, Deserializer};
+use serde::de::{EnumAccess, Unexpected, VariantAccess, Visitor, self};
 use std::cmp::PartialEq;
 use std::error::Error;
 use std::fmt;
 use std::io::Write;
 use std::ops::{Deref, DerefMut};
 use std::str;
-use url::{Url, Host};
+use url::{Url, Host, ParseError};
 
 /// Serialises `value` with a given serializer.
 ///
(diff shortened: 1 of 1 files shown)


Hoverbear:
I updated the ticket name to reflect that it does both `Ser` and `De`.

url_serde/src/lib.rs
```diff
@@ -202,6 +231,141 @@ impl<'de> Deserialize<'de> for De<Option<Url>> {
     }
 }
 
+/// Deserializes this ParseError from a `serde` stream.
+impl<'de> Deserialize<'de> for De<ParseError> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
+        enum Variant {
```
Hoverbear: I'm a bit unhappy with the level of duplication here. Is there maybe some way we can reduce the amount of times we list all the variants? In the interest of maintainability it would be great to only need to change the code in one (or two) spots when we add a variant. At the moment it seems like we need to do it in at least 6 places.

KiChjang: The duplication here is mostly because of my own lack of knowledge of how serde works. I have no idea how the implementation body of `ParseErrorVisitor` would look otherwise.

KiChjang: Also, I'm under the impression that `url_serde` is only a temporary fix until we upgrade to serde 1.0 on master, so I guess this is fine for now?

SimonSapin:
`master` now depends on serde 1.x, so it should be possible to use serde_derive which would be much simpler than this PR. Feel free to open another PR doing that.

PR #389, opened 2017-08-23
Add missing whitespace

maghoff:
Without this whitespace, the example code would be rendered as paragraph text. This change fixes it so the example code is rendered as code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/389)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1757,6 +1757,7 @@ impl Url {
     /// ```
     ///
     /// Setup username to user1
+    ///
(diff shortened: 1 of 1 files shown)


Hoverbear:
Thanks for this @maghoff :)

maghoff:
Aha, looks like this also uncovered a broken assert in the example code. Fixing...

SimonSapin:
Thank you!

@bors-servo r+

PR #393, opened 2017-09-28
Derive `Copy` for some structures.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/393)

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -23,7 +23,7 @@ include!("uts46_mapping_table.rs");
 pub static PUNYCODE_PREFIX: &'static str = "xn--";
 
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 struct StringTableSlice {
     // Store these as separate fields so the structure will have an
     // alignment of 1 and thus pack better into the Mapping enum, below.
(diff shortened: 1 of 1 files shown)


SimonSapin:
These types are private, and it looks like nothing uses these new impls. Why this change?

frewsxcv:
My understanding was that deriving `Copy` for types would allow the compiler to better optimize and allow for more inlining. But if that's not the case, let me know and I can revert

SimonSapin:
I hadn’t heard of that. Maybe that’s the case, but I’d be surprised. Do you have a reference for that? If the code compiled without `Copy`, any place it would do an implicit copy it instead did a move. And a move is implemented the same as a copy, only you’re not allowed to use the old location afterwards. But that’s only a compile-time check.

frewsxcv:
I haven't found anything concrete about performance, so I think you're right. There is [this](https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#when-should-my-type-be-copy), but I'm guessing that's more of a suggestion for public types, and I didn't realize these types _were not_ public until you mentioned it a couple comments ago.

In any case: https://github.com/servo/rust-url/pull/399

PR #394, opened 2017-09-28
Remove unnecessary bounds checking in a couple places.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/394)


SimonSapin:
Why?

`unsafe` is inherently dangerous, and the gains here seem like they’d be tiny if they exist at all. (The optimizer might be able to elide these checks anyway.) Is some benchmark improved by this change?

SimonSapin:
Reverting in https://github.com/servo/rust-url/pull/401.

PR #403, opened 2017-10-25
Use opaque host parser for non-special schemes

valenting:
Addresses URL spec change:
https://github.com/whatwg/url/commit/30362553e9ce9fc706d3492bd61886e399fc94e2

For non-special schemes we use the opaque host parser - meaning the IDNA parser is not used, but instead we apply a UTF-8 percent encode using the simple encode set on the host input.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/403)


SimonSapin:
Please change the doc-comment of `Host::Domain` to say that it also includes opaque host names.

Next time we make breaking changes to this crate’s API, would it be useful to have `Opaque` be a separate variant in the `Host` enum? Or do users always do the same thing in both cases?

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/403#-:-KxRsaXexxMquBjc-pya:b-6y8k7i)*

valenting:
The URL spec does list opaque hosts separately from regular hosts, but I don't know if this is useful reasons other than parsing the URL.

---

Review status: 4 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/403#-:-KxliW-JxA_xz7C6mQkG:b-ktfbt6)*

SimonSapin:
@bors-servo r+

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-url/403#-:-KxnohjGABAiPo3nTW3g:bkg9nno)*

Issue #417, opened 2017-12-09
Crate doesn't compile for `wasm32-unknown-unknown`


tailhook:
The error is here:
```
error[E0425]: cannot find function `path_to_file_url_segments` in this scope
    --> target/registry/src/github.com-1ecc6299db9ec823/url-1.6.0/src/lib.rs:1934:32
     |
1934 |         let (host_end, host) = path_to_file_url_segments(path.as_ref(), &mut serialization)?;
     |                                ^^^^^^^^^^^^^^^^^^^^^^^^^ did you mean `path_to_file_url_segments_windows`?
```
Shouldn't be hard to fix, but I'm not sure how/if PR also needs to update CI?

SimonSapin:
This function has two implementations, one for `#[cfg(any(unix, target_os = "redox"))]`, and one for `#[cfg(windows)]` maybe the method that uses it should be `#[cfg(any(windows, unix, target_os = "redox"))]`?

As to CI, what’s the minimum set of dependencies and build steps to reproduce this failure?

tailhook:
Okay, it looks like it compiles with emscripten, probably because it's `unix`.

For `wasm32-unknown-unknown` I can't find a way to construct `OsStr` from bytes. There is #418 which reproduces the issue on travis CI.

I'm not sure what to do from here. Probably report but to rust-lang ? Another alternative is to temporarily disable `from_file_path` and `to_file_path` for the target. Does it sound viable?

tailhook:
Well, I figured out that I can actually convert OsStr to/from str (utf-8) on any platform. So I've done an implementation which is similar to windows (i.e. only works on valid utf-8) but works just as unix one (i.e. unixy slashes and other conventions). #418

SimonSapin:
Fixed in https://github.com/servo/rust-url/pull/419, thanks!

PR #418, opened 2017-12-10
Fix building wasm target

tailhook:
Working on #417. This PR is just to make travis build tests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/418)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
 language: rust
-rust:
-  - nightly
-  - beta
-  - stable
-  - 1.17.0
 script: make test
+
+jobs:
+  include:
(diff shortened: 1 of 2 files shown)


tailhook:
Similarly to windows wasm (non-enscripten) target now only works on utf-8 paths. Which is perfectly okay for URL's I think.

SimonSapin:
I’d prefer not to commit to any behavior of `Path` in wasm until this is more settled in Rust itself, and rather make path-related methods not exist on non-Unix non-Windows platforms.

tailhook:
Okay, so there is #419 that does just that.

PR #421, opened 2017-12-31
Add host_port and request_uir methods

mina86:
`host_port` and `request_uri` methods are useful when dealing with HTTP requests. `host_port` is what needs to be sent in Host header while `request_uri` is what needs to be sent in the request line.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/421)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1118,6 +1118,37 @@ impl Url {
         form_urlencoded::parse(self.query().unwrap_or("").as_bytes())
     }
 
+    /// Return this URL’s path and query string.  If query string is not present
+    /// return just the path; otherwise returns path followed by a question mark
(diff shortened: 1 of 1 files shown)


SimonSapin:
Aren’t these methods equivalent to `&url[Position::BeforeHost..Position::AfterPort]` and `&url[Position::BeforePath..Position::AfterQuery]` respectively? `Position` exists specifically to enable all kinds of slicing without adding lots of ad-hoc methods.

mina86:
Ha! Yes, I missed those in the documentation. host_port is not exactly equivalent to `&url[Position::BeforeHost..Position::AfterPort]` since it checks whether port is the default, but it’s close enough.

PR #426, opened 2018-01-17
Introduce Host::serialize

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/426)


SimonSapin:
As discussed on IRC the IPv6 case should use `format!("[{}]", addr)` and I’d prefer the method to be called `serialize`.

nox:
My single use case can instead go through `Display`, which makes this method redundant, closing.

PR #428, opened 2018-01-20
Define `HEADER_FIELD_ENCODE_SET`

Ralith:
This is needed for encoding UTF-8 in header fields, e.g. `Content-Disposition: attachment; filename*=UTF-8''...`. Defined in [RFC8187](https://tools.ietf.org/html/rfc8187). I'd appreciate a double-check that I got the characters right.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/428)

### DIFF
--- a/percent_encoding/lib.rs
+++ b/percent_encoding/lib.rs
@@ -156,6 +156,16 @@ define_encode_set! {
     }
 }
 
+define_encode_set! {
+    /// RFC8187 HTTP header field encoding
+    ///
+    /// Encodes all bytes that are not RFC7230 token bytes, in addition to asterisk (*), single quote ('),
+    /// and percent (%). Commonly used for encoding Unicode filenames in a `Content-Disposition` header.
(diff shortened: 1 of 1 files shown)


SimonSapin:
The `define_encode_set!` macro is intended to be usable in other crates. Is there a reason this specific set should be defined in this crate?

Ralith:
Because it's a standardized encoding commonly needed by both web services and clients, similar to the others. Correct implementations of the `Content-Disposition` header are a wonderful thing.

SimonSapin:
Still, wouldn’t this belong more in a crate like `http`? Even in things with RFCs there are many that use precent-encoding slightly differently, I’d prefer this crate not to be responsible for them all.

Ralith:
The `http` crate does not depend on this crate. Is that likely to change?

SimonSapin:
Closing as out of scope. Even with this PR, this crate does not provide an implementation (correct or not) of the `Content-Disposition` header.

PR #431, opened 2018-01-24
Fix encoding::EncodingOverride when not enabling query_encoding feature

nox:
This is a breaking change, but at the same this fixes a bug that can cause build issues anyway, so I'm on the fence this time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/431)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "1.6.1"
+version = "1.6.2"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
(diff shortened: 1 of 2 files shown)


nox:
Cc @emilio @SimonSapin @Eijebong because we all have opinions about semver.

nox:
This may be a good opportunity to define the missing methods that don't depend on the `encoding` crate for when the feature is not enabled.

SimonSapin:
What’s the build issue?

Note that the `encoding` module is not public and its content not re-exported. Now that `pub(crate)` is a thing, we should use it for its contents to make that more clear.

nox:
@SimonSapin Ooooooooh. Sorry for the brainfart. 'Tis no public, so there is no problem.

PR #433, opened 2018-02-05
Improve syntax violations interface and preserve Copy

dekellum:
This is an alternative implementation of #430 which avoids introducing an `Rc<Fn>` and preserves `Copy` for `ParseOptions`.   Thus, based on my current understanding with the prior review, it is completely backward compatible and with no breaking changes.  Its a little less elegant in that, without the `Rc<Fn>`, I can not simple wrap the deprecated log_syntax_violation Fn with an adapting closure.  But by moving the two possible function references into a new enum `ViolationFn` it is reasonably well contained.  The interface for users is effectively the same, so I think the backward compatibility advantage makes this preferable to #430.  Updated summary:

### Summary of changes

This makes programmatic use of syntax violations more practical while further centralizing this concern in the parser, which should be a maintenance win.

* New `SyntaxViolation` enum with descriptions the same as the prior violation static strings.  This enables programmatic use of these violations by variant, e.g. ignoring some, warning on others, providing translated descriptions, etc. Enum implementation uses the same macro expansion pattern as was used for `ParseError`. 

* New `syntax_violation_callback` signature in `ParseOptions` and `Parser`.

* Deprecated `log_syntax_violation` and re-implemented, passing the same violation descriptions for backward compatibility.

* Unit tests for old `log_syntax_violation` compatibility, and new `syntax_violation_callback` 

* Includes rustdoc for old and new interfaces including tested example usage of new callback, as requested in #308.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/433)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,7 +117,7 @@ pub extern crate percent_encoding;
 use encoding::EncodingOverride;
 #[cfg(feature = "heapsize")] use heapsize::HeapSizeOf;
 use host::HostInternal;
-use parser::{Parser, Context, SchemeType, to_u32};
+use parser::{Parser, Context, SchemeType, to_u32, ViolationFn};
 use percent_encoding::{PATH_SEGMENT_ENCODE_SET, USERINFO_ENCODE_SET,
                        percent_encode, percent_decode, utf8_percent_encode};
 use std::borrow::Borrow;
@@ -135,7 +135,7 @@ use std::str;
 pub use origin::{Origin, OpaqueOrigin};
 pub use host::{Host, HostAndPort, SocketAddrs};
 pub use path_segments::PathSegmentsMut;
-pub use parser::ParseError;
+pub use parser::{ParseError, SyntaxViolation};
 pub use slicing::Position;
 
 mod encoding;
@@ -186,7 +186,7 @@ impl HeapSizeOf for Url {
 pub struct ParseOptions<'a> {
     base_url: Option<&'a Url>,
     encoding_override: encoding::EncodingOverride,
-    log_syntax_violation: Option<&'a Fn(&'static str)>,
+    violation_fn: ViolationFn<'a>,
 }
 
 impl<'a> ParseOptions<'a> {
@@ -209,9 +209,47 @@ impl<'a> ParseOptions<'a> {
         self
     }
 
-    /// Call the provided function or closure on non-fatal parse errors.
+    /// Call the provided function or closure on non-fatal parse errors, passing
+    /// a static string description.  This method is deprecated in favor of
+    /// `syntax_violation_callback` and is implemented as an adaptor for the
+    /// latter, passing the `SyntaxViolation` description. Only the last value
+    /// passed to either method will be used by a parser.
+    #[deprecated]
     pub fn log_syntax_violation(mut self, new: Option<&'a Fn(&'static str)>) -> Self {
-        self.log_syntax_violation = new;
+        self.violation_fn = match new {
+            Some(f) => ViolationFn::OldFn(f),
+            None => ViolationFn::NoOp
+        };
+        self
+    }
+
+    /// Call the provided function or closure for a non-fatal `SyntaxViolation`
+    /// when it occurs during parsing. Note that since the provided function is
+    /// `Fn`, the caller might need to utilize _interior mutability_, such as with
+    /// a `RefCell`, to collect the violations.
+    ///
+    /// ## Example
+    /// ```
+    /// use std::cell::RefCell;
+    /// use url::{Url, SyntaxViolation};
+    /// # use url::ParseError;
+    /// # fn run() -> Result<(), url::ParseError> {
+    /// let violations = RefCell::new(Vec::new());
+    /// let url = Url::options()
+    ///     .syntax_violation_callback(Some(&|v| violations.borrow_mut().push(v)))
+    ///     .parse("https:////example.com")?;
+    /// assert_eq!(url.as_str(), "https://example.com/");
+    /// assert_eq!(violations.into_inner(),
+    ///            vec!(SyntaxViolation::ExpectedDoubleSlash));
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
+    pub fn syntax_violation_callback(mut self, new: Option<&'a Fn(SyntaxViolation)>) -> Self {
+        self.violation_fn = match new {
+            Some(f) => ViolationFn::NewFn(f),
+            None => ViolationFn::NoOp
+        };
         self
     }
 
@@ -221,7 +259,7 @@ impl<'a> ParseOptions<'a> {
             serialization: String::with_capacity(input.len()),
             base_url: self.base_url,
             query_encoding_override: self.encoding_override,
-            log_syntax_violation: self.log_syntax_violation,
+            violation_fn: self.violation_fn,
             context: Context::UrlParser,
         }.parse_url(input)
     }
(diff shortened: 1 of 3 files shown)


nox:
Wouldn't it be better to introduce a separate trait rather than use a different `Fn`? If we ever want to extend it, we should rather not rely on `Fn` directly.

dekellum:
The extension scenarios I could imagine would be best applied to the introduced SyntaxViolation (methods or tuple additions if needed).  My goal was to make this part of the interface _easier_ to use. Unfortunately the builder situation won't allow `FnMut`, but an alternative trait would also need to be called with a non-mutable reference, and I believe would only increase the burden to the user, who would now need to implement the trait outside of block context with the options building and parse call? Consider the additions needed for the provided new tests, or the doc example.

dekellum:
Well, per my understanding your trait example in #430, the following could adapt back to the current `Fn(SyntaxViolation)` as seen in tests and examples, so it would allow having it both ways:

```rust
impl SyntaxViolationSink for Fn(SyntaxViolation) {
    fn syntax_violation(&self, violation: SyntaxViolation) {
        self(violation)
    }
}
```
But its more interface surface if SyntaxViolationSink is exported, and I think it needs to be for external closure-based usage?

Note that the ViolationFn of this PR doesn't need to be, and isn't exported.  Otherwise it should roughly be the same amount of code.  I'm not seeing the extension advantage and I doubt there is practical uses of a Sink without the context of the other parse options, and url being parsed, and any ParseError?

dekellum:
I hope I've adequately addressed @nox's concern on extensiblity and implementation strategy.  @SimonSapin: Could you review the remaining aspects of this, now with the breaking change concern hopefully out of the way?

tests/unit.rs
```diff
@@ -482,11 +482,9 @@ fn test_windows_unc_path() {
...
 fn test_old_log_violation_option() {
     let violation = Cell::new(None);
-    let url = {
-        let vfn = |s: &str| violation.set(Some(s.to_owned()));
-        let options = Url::options().log_syntax_violation(Some(&vfn));
-        options.parse("http:////mozilla.org:42").unwrap()
-    };
+    let url = Url::options().
+        log_syntax_violation(Some(&|s| violation.set(Some(s.to_owned())))).
+        parse("http:////mozilla.org:42").unwrap();
```
nox: This is badly indented, the periods shouldn't be at the end of lines, but at the start.

src/lib.rs
```diff
@@ -209,9 +209,49 @@ impl<'a> ParseOptions<'a> {
         self
     }
 
-    /// Call the provided function or closure on non-fatal parse errors.
+    /// Call the provided function or closure on non-fatal parse errors, passing
+    /// a static string description.  This method is deprecated in favor of
+    /// `syntax_violation_callback` and is implemented as an adaptor for the
+    /// latter, passing the `SyntaxViolation` description. Only the last value
+    /// passed to either method will be used by a parser.
+    #[deprecated]
```
nox: I'm not sure the method should immediately be made deprecated.

src/lib.rs
```diff
@@ -186,7 +186,7 @@ impl HeapSizeOf for Url {
 pub struct ParseOptions<'a> {
     base_url: Option<&'a Url>,
     encoding_override: encoding::EncodingOverride,
-    log_syntax_violation: Option<&'a Fn(&'static str)>,
+    violation_fn: ViolationFn<'a>,
```
nox: Please store an `Option<ViolationFn<`'a>>` and remove the `NoOp` variant.

src/lib.rs
```diff
@@ -209,9 +209,49 @@ impl<'a> ParseOptions<'a> {
...
+    ///     syntax_violation_callback(Some(&|v| violations.borrow_mut().push(v))).
+    ///     parse("https:////example.com")?;
+    /// assert_eq!(url.as_str(), "https://example.com/");
+    /// assert_eq!(violations.into_inner(),
+    ///            vec!(SyntaxViolation::ExpectedDoubleSlash));
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
+    pub fn syntax_violation_callback<F>(mut self, new: Option<&'a F>) -> Self
```
nox: Remove `F` and pass a trait object here too.

src/parser.rs
```diff
@@ -216,11 +263,56 @@ impl<'i> Iterator for Input<'i> {
...
+impl<'a> ViolationFn<'a> {
+    pub fn call(self, v: SyntaxViolation) {
+        match self {
+            ViolationFn::NewFn(f) => f(v),
+            ViolationFn::OldFn(f) => f(v.description()),
+            ViolationFn::NoOp => {}
+        }
+    }
+
+    pub fn call_if<F>(self, v: SyntaxViolation, test: F)
```
nox: Please just use `if` where you currently call `call_if`.

src/parser.rs
```diff
@@ -216,11 +263,56 @@ impl<'i> Iterator for Input<'i> {
     }
 }
 
+/// Wrapper for syntax violation callback functions.
+#[derive(Copy, Clone)]
+pub enum ViolationFn<'a> {
```
nox: Make this `pub(crate)`.

tests/unit.rs
dekellum: Converted to periods-on-the-left style in all PR changes.

src/lib.rs
dekellum: Its not required for my use case.   But why not give deprecation warning to those (relatively few) users of this feature to migrate to Fn(SyntaxViolation), calling  `v.description()` themselves if that is all they want? If deprecated in the next minor release (merged), then this could be removed in some future major release, as cleanup.

src/lib.rs
dekellum: Simplified in 6017325. This was legacy from my prior Rc version with pass by value requiring `<F>`, but I see that it has no remaining value.

src/parser.rs
dekellum: Changing this to `pub(crate)` fails with the following errors on the parser module public, but crate private, structs `Input` and `Parser`.  I suspect that all of these cases would need to be made `pub(crate)` of just `pub`, but I'm reluctant to add that this PR, as it expands scope and risk. If it worth it, consider as a follow-on change?

```txt
error[E0446]: private type `parser::ViolationFn<'_>` in public interface
   --> src/parser.rs:166:5
    |
166 | /     pub fn with_log(original_input: &'i str, vfn: ViolationFn) -> Self {
167 | |         let input = original_input.trim_matches(c0_control_or_space);
168 | |         if vfn.is_set() {
169 | |             if input.len() < original_input.len() {
...   |
176 | |         Input { chars: input.chars() }
177 | |     }
    | |_____^ can't leak private type

error[E0446]: private type `parser::ViolationFn<'a>` in public interface
   --> src/parser.rs:319:5
    |
319 |     pub violation_fn: ViolationFn<'a>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
~~~

src/parser.rs
dekellum: `Violation#call_if` is direct replacement for the original `Parser#syntax_violation_if`.  I _think_ this was introduced as and remains a useful optimization and convenience in that it does an early check for `NoOp` (short-circuit eval of test). Beyond that I have been trying not to change every call by needing to rewrite with `if` blocks, which would be a much more extensive change of the `Parser`, and require more scrutiny.

While I could have kept and just modified the original Parser method in place, I think this PR better collects those concerns under `ViolationFn`.  I added comments to these `ViolationFn` methods to hopefully clear this up. Hope that is sufficient.

src/lib.rs
dekellum: If I did that, it would complicate things.  Firstly, I couldn't directly implement the helper methods `call` and `call_if`, given the outer type is now Option.  Secondly it would complicate those methods even if I could.  See the comments around `call_if` below.

src/parser.rs
nox: I guess that's ok, I just wanted to make sure `ViolationFn` doesn't escape the crate.

src/lib.rs
nox: I see. Why not make the methods on `Parser` instead of `ViolationFn`? This would let you use `Option<T>`, right?

dekellum: OK, for preview, in 95a7b4a I reinstate Option and move the necessary methods back to the Parser.  This change would reduce the [overall diff](/servo/rust-url/compare/master...dekellum:95a7b4a) for the feature, keeping more status-quo with the parser.   However it spreads around the `if let Some(vfn) ...` and the `*_if` function can't be used outside of the `Parser` impl. `Input` already uses `ViolationFn::call`.  If the `*_if` variant is ever needed elsewhere this will prove a liability. 

I personally still think the current PR is better without this change, but if you prefer 95a7b4a, I will include it in this branch and PR.

nox:
This looks ok. Please squash the commits accordingly and I'll r=me it.

dekellum:
Squashed to 2 commits, where 255df06 has no diff. from 96148cf, the previously reviewed branch head.

PR #435, opened 2018-02-18
Exclude unneeded files from the crate package.

RazrFalcon:
This decreases `*.crate` file size from 282KiB to 90KiB.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/435)


SimonSapin:
I personally don’t care much, but some people have files bugs on some other crates when `cargo test` doesn’t work in the result of extracting a `.crate` file from crates.io. I suppose it’d be nice to have a community-wide convention of what’s appropriate to do/expect around this.

@nox, do you have an opinion?

PR #442, opened 2018-04-03
Fix `to_file_path` for relative paths with drive letters

matklad:
Hi!

I've recently hit [this debug assertion](https://github.com/servo/rust-url/blob/04d5ec1e7f3e70c850ff2e7ba7d83d446926defe/src/lib.rs#L2414) on windows for paths like `file://C:`.

I've added a couple of tests and a "fix", which just turns the assertion into error. I have no idea how to fix this properly.

### DIFF
--- a/tests/unit.rs
+++ b/tests/unit.rs
@@ -86,6 +86,29 @@ fn new_path_windows_fun() {
     }
 }
 
+#[test]
+fn to_file_path_for_relative_windows_paths() {
+    if cfg!(windows) {
+        let url = Url::parse("file://example.com/C:foo").unwrap();
+        assert!(url.to_file_path().is_ok());
+
+        let url = Url::parse("file://example.com/C:").unwrap();
+        assert!(url.to_file_path().is_err());
+
+        let url = Url::parse("file:///E:foo").unwrap();
+        assert!(url.to_file_path().is_err());
+
+        let url = Url::parse("file:///E:").unwrap();
+        assert!(url.to_file_path().is_err());
+
+        let url = Url::parse("file://localhost/C:foo").unwrap();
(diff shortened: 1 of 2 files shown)


tests/unit.rs
```diff
@@ -86,6 +86,29 @@ fn new_path_windows_fun() {
     }
 }
 
+#[test]
+fn to_file_path_for_relative_windows_paths() {
+    if cfg!(windows) {
+        let url = Url::parse("file://example.com/C:foo").unwrap();
+        assert!(url.to_file_path().is_ok());
```
matklad: NB: unlike other tests, this is `ok`. I would think that it should be `err` as well, but `std::Path` disagrees with me and insists that `\\exaple.com\C:foo` is an absolute path. Is this a bug in stdlib maybe as well?

matklad: @retep998 perhaps you know? Consider this program:

```rust
use std::path::PathBuf;

fn main() {
    let p1 = PathBuf::from(r"C:foo");
    let p2 = PathBuf::from(r"\\example.com\C:foo");
    println!("{} {}", p1.is_absolute(), p2.is_absolute());
}
```

It currently prints `false true`. Is this the correct behavior?

retep998: `C:foo` is *relative* to the current directory for the `C` drive. `\\example.com\C:foo` is a UNC path which is absolute.

For more information on the types of paths... https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html

matklad:
ping @SimonSapin :-)  

I don't feel a pressing need for this or anything, but this issue seems somewhat interesting, because it arguably allows to sneak in relative path in release build, which might be security sensitive?

PR #443, opened 2018-04-04
Add ParseOptions::encoding_override_lookup

talklittle:
Exposes an API that takes a `&[u8]` slice, instead of requiring the `rust-encoding` crate downstream for the `encoding::EncodingRef` type.

Helpful for https://github.com/servo/servo/issues/13238

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/443)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -209,6 +209,24 @@ impl<'a> ParseOptions<'a> {
         self
     }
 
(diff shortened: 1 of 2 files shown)


talklittle:
Closing in favor of #445

PR #445, opened 2018-04-05
Add Cargo feature `query_encoding_2` to use `encoding_rs` crate

talklittle:
Fixes #444

PR #446, opened 2018-04-06
Add Cargo feature `query_encoding_2` to use `encoding_rs` crate

### LINKED ISSUE #444: Migrate to encoding_rs crate?
Servo has migrated from [rust-encoding](https://github.com/lifthrasiir/rust-encoding) to [encoding_rs](https://github.com/hsivonen/encoding_rs).

Is the plan to do the same migration for `rust-url`, when using the `query_encoding` feature?

talklittle:
This attempt is substantially different than the previous approach in #445, so opening a separate PR.

Converts `EncodingOverride` to a trait, but does not go so far as to export the trait to downstream consumers. Instead, the exported API now takes a `&[u8]` encoding label where it used to take `encoding::EncodingRef`. The old methods are still present, but deprecated.

Fixes #444

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/446)


nox:
The crate doesn't depending on any encoding-related crate anymore, closing.

PR #453, opened 2018-06-25
Improve performance of the uts46 crate

Marwes:
I may have complicated the table structure more than is desired so I'd be happy to remove it if you want. It does shrink the total size of the tables by 6000 * 8 - 1500 * 2 = 42kb though (and gives a slight performance boost).

(Calculation is done as `removed_ranges * range_size - added_indexes * index_size`)

Total performance change of parsing a small and simple url.
```
 name   old ns/iter     new ns/iter     diff ns/iter   diff %  speedup 
 short  4,819 (5 MB/s)  3,516 (7 MB/s)        -1,303  -27.04%   x 1.37 
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/453)


SimonSapin:
Looks good, thank you!

---

Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 3 of 3 files at r6, 3 of 3 files at r7, 1 of 1 files at r8, 2 of 2 files at r9, 1 of 1 files at r10.
Review status: all files reviewed, 3 unresolved discussions (waiting on @Marwes)

---

*[idna/src/make_uts46_mapping_table.py, line 130 at r5](https://reviewable.io/reviews/servo/rust-url/453#-LFwjZDK1PFLLeROd4UY:-LFwjZDK1PFLLeROd4UZ:b-iij7qj) ([raw file](https://github.com/servo/rust-url/blob/94e836c3f3cfe486a8f0e31369f6fb46e2dc32cc/idna/src/make_uts46_mapping_table.py#L130)):*
> ```Python
>     current = []
>     for r in ranges:
>         mapping = r[2]
> ```

Coding style nit: I find this hard to read when I need to keep elsewhere to find out what is at index 0, 1, 2. What do you think of using tuple deconstruction in the `for` loop?

---

*[idna/src/make_uts46_mapping_table.py, line 164 at r7](https://reviewable.io/reviews/servo/rust-url/453#-LFwl3HX5uWg_eLhrNrl:-LFwl3HX5uWg_eLhrNrm:bqa6x5w) ([raw file](https://github.com/servo/rust-url/blob/fca56943473d3c50163533d4e55d193b9c02a1cf/idna/src/make_uts46_mapping_table.py#L164)):*
> ```Python
>     block_len = len(ranges)
>     single = (1 << 15) if block_len == 1 else 0 
>     print("    %s | %s, " % (offset, single))
> ```

If this boolean / bit still needs to be stored (see previous comment), I’d prefer the bitwise OR to be done in Python, since many more people will recompile the Rust code. Also add an assert that we don’t overflow 15 bits?

---

*[idna/src/uts46.rs, line 75 at r6](https://reviewable.io/reviews/servo/rust-url/453#-LFwkjQ6704vAg7GUbEH:-LFwkjQ6704vAg7GUbEI:b-pc5vz4) ([raw file](https://github.com/servo/rust-url/blob/281958c0de441579d02d5568d4f5e0c9ac218e04/idna/src/uts46.rs#L75)):*
> ```Rust
>             &MAPPING_TABLE[offset as usize]
>         } else {
>             &MAPPING_TABLE[(offset + codepoint as u16 - TABLE[i].from as u16) as usize]
> ```

Aren’t these two code paths the same? Maybe the boolean is not needed at all?

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/rust-url/453#-:-LFwln6wAeC6NAvvuYRm:bdkiwka)*

Marwes:
Review status: all files reviewed, 3 unresolved discussions (waiting on @Marwes)

---

*[idna/src/make_uts46_mapping_table.py, line 130 at r5](https://reviewable.io/reviews/servo/rust-url/453#-LFwjZDK1PFLLeROd4UY:-LFxQYGD9Al1o-0BrdJc:b-8fx2u7) ([raw file](https://github.com/servo/rust-url/blob/94e836c3f3cfe486a8f0e31369f6fb46e2dc32cc/idna/src/make_uts46_mapping_table.py#L130)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

Coding style nit: I find this hard to read when I need to keep elsewhere to find out what is at index 0, 1, 2. What do you think of using tuple deconstruction in the `for` loop?
</blockquote></details>

:+1: 

---

*[idna/src/make_uts46_mapping_table.py, line 164 at r7](https://reviewable.io/reviews/servo/rust-url/453#-LFwl3HX5uWg_eLhrNrl:-LFxQT1HBhywGjtT3PiY:bb74ce2) ([raw file](https://github.com/servo/rust-url/blob/fca56943473d3c50163533d4e55d193b9c02a1cf/idna/src/make_uts46_mapping_table.py#L164)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

If this boolean / bit still needs to be stored (see previous comment), I’d prefer the bitwise OR to be done in Python, since many more people will recompile the Rust code. Also add an assert that we don’t overflow 15 bits?
</blockquote></details>

Sure

---

*[idna/src/uts46.rs, line 75 at r6](https://reviewable.io/reviews/servo/rust-url/453#-LFwkjQ6704vAg7GUbEH:-LFxPwDL3W34dHDlZYU4:bziimyy) ([raw file](https://github.com/servo/rust-url/blob/281958c0de441579d02d5568d4f5e0c9ac218e04/idna/src/uts46.rs#L75)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

Aren’t these two code paths the same? Maybe the boolean is not needed at all?
</blockquote></details>

It differentiates between the two cases

* One (single) mapping for an entire range
* One mapping for each element in a range (added through this PR)

In the first case multiple codepoints needs to map to `offset` so I can't do any arithmetic on it

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/rust-url/453)*

SimonSapin:
Review status: all files reviewed, 2 unresolved discussions (waiting on @SimonSapin)

---

*[idna/src/uts46.rs, line 75 at r6](https://reviewable.io/reviews/servo/rust-url/453#-LFwkjQ6704vAg7GUbEH:-LFxdM9B6YA4P6xzrjk6:bbocvh0) ([raw file](https://github.com/servo/rust-url/blob/281958c0de441579d02d5568d4f5e0c9ac218e04/idna/src/uts46.rs#L75)):*
<details><summary><i>Previously, Marwes (Markus Westerlind) wrote…</i></summary><blockquote>

It differentiates between the two cases

* One (single) mapping for an entire range
* One mapping for each element in a range (added through this PR)

In the first case multiple codepoints needs to map to `offset` so I can't do any arithmetic on it
</blockquote></details>

Ok I see, thanks.

---


*Comments from [Reviewable](https://reviewable.io/reviews/servo/rust-url/453)*

SimonSapin:
Great, thanks!

@bors-servo r+

---

Reviewed 1 of 1 files at r11, 1 of 1 files at r12, 3 of 3 files at r13, 3 of 3 files at r14.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/servo/rust-url/453#-:-LG-Wr8a4yZ3gAYUk32A:b-2mq7t1)*

PR #454, opened 2018-07-04
Fix a lifetime bug uncovered by NLL, thanks @lqd

SimonSapin:
A similar error message was: https://play.rust-lang.org/?gist=4c5f4002623da9a3df723c9540dd125e&version=nightly&mode=debug&edition=2015

```rust
error[E0509]: cannot move out of type `S<'_>`, which implements the `Drop` trait
  --> src/main.rs:15:17
   |
15 |         let S { url } = self;
   |                 ^^^ cannot move out of here
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/454)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "1.7.0"
+version = "1.7.1"
 authors = ["The rust-url developers"]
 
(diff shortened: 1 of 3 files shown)


lqd:
BTW I've also filed a diagnostics issue for the less-than-clear error message shown for the original code https://github.com/rust-lang/rust/issues/52059

PR #466, opened 2018-10-30
Add test case for ambiguous ipv6

jedahan:
I don't think this should be merged, because CI will fail, but it would be nice to start the conversation if this truly should be an error. Motivation is reducing false positives in https://github.com/jwilm/alacritty/issues/1727 .

If we determine it should be an error, I'd like to tackle fixing it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/466)

### DIFF
--- a/tests/urltestdata.json
+++ b/tests/urltestdata.json
@@ -6144,5 +6144,10 @@
     "pathname": "/test",
     "search": "?a",
     "hash": "#bc"
+  },
+  {
+    "input": "f::some::ambiguous::ipv6",
+    "base": "http://example.org/foo/bar",
+    "failure": true
   }
 ]


SimonSapin:
The parsing algorithm that we implement only looks for an IPv6 address if it’s enclosed in `[` and `]` brakets: https://url.spec.whatwg.org/#host-parsing

valenting:
Also, this test shouldn't be in urltestdata.json.
I assume we want to track [web-platform-tests](https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json) as closely as possible, in which case this should either be submitted to that repo, or converted to a regular rust unit test.
But I agree with Simon, I think current parsing behaviour for f::some::ambiguous::ipv6 is the correct one.

PR #469, opened 2018-11-06
Escape backslash in special URL path components

### LINKED ISSUE #468: `\` characters should be escaped in paths of special URLs
`\` characters are not escaped when appended to paths currently:

```rust
let mut url = "http://foobar.com".parse::<Url>().unwrap();
url.path_segments_mut().unwrap().push("foo\\bar");
assert_eq!(url.as_str(), "http://foobar.com/foo\\bar");
```

However, the [path-state](https://url.spec.whatwg.org/#path-state) section of the URL standard treats `\` as a path separator and a validation error in special URLs. I believe `push` should be percent-encoding the `\` to `%5C`.

sfackler:
Closes #468

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/469)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -19,22 +19,13 @@ use host::{Host, HostInternal};
 use percent_encoding::{
     utf8_percent_encode, percent_encode,
     SIMPLE_ENCODE_SET, DEFAULT_ENCODE_SET, USERINFO_ENCODE_SET, QUERY_ENCODE_SET,
-    PATH_SEGMENT_ENCODE_SET, EncodeSet
+    PATH_SEGMENT_ENCODE_SET
 };
 
-// The backslash (\) character is treated as a path separator in special URLs
-// so it needs to be additionally escaped in that case.
-#[derive(Clone)]
-struct SPECIAL_PATH_SEGMENT_ENCODE_SET;
-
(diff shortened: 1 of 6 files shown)


src/parser.rs
```diff
@@ -19,9 +19,24 @@ use host::{Host, HostInternal};
...
     utf8_percent_encode, percent_encode,
     SIMPLE_ENCODE_SET, DEFAULT_ENCODE_SET, USERINFO_ENCODE_SET, QUERY_ENCODE_SET,
-    PATH_SEGMENT_ENCODE_SET
+    PATH_SEGMENT_ENCODE_SET, EncodeSet
 };
 
+// The backslash (\) character is treated as a path separator in special URLs
+// so it needs to be additionally escaped in that case.
+#[derive(Clone)]
+struct SPECIAL_PATH_SEGMENT_ENCODE_SET;
```
SimonSapin: Could this use the `define_encode_set` macro?

SimonSapin:
Thanks!

@bors-servo r+

SimonSapin:
https://crates.io/crates/url/1.7.2 is up. (I didn’t bother with the `percent-encoding` crate since it’s only docs changes.)

sfackler:
Thanks!

PR #477, opened 2019-01-14
Redox support

jackpot51:
Redox uses `file:` at the beginning of any file paths. Therefore, a path like `file:///home/user/some%20file.txt` needs to turn into `file:/home/user/some file.txt`

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/477)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2351,7 +2351,11 @@ fn file_url_segments_to_pathbuf(host: Option<&str>, segments: str::Split<char>) 
         return Err(());
     }
 
-    let mut bytes = Vec::new();
+    let mut bytes = if cfg!(target_os = "redox") {
+        b"file:".to_vec()
+    } else {
+        Vec::new()
+    };
     for segment in segments {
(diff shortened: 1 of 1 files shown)


SimonSapin:
The initial value of `bytes` is the only difference with the Unix variant of this function, right? I’d prefer we keep a single function and use something like `let mut bytes = if cfg!(target_os = "redox") { b"file:".to_vec() } else { Vec::new() };`, to reduce code duplication.

Does the `path_to_file_url_segments` function need any corresponding change?

jackpot51:
Good point, I can do that

jackpot51:
I will check path_to_file_url_segments

jackpot51:
I cleaned up the patch. It is the best I can do so far. There is not much I can do with `path_to_file_url_segments` until this issue is solved: https://github.com/rust-lang/rust/issues/52331

SimonSapin:
Thanks!

@bors-servo r+

PR #481, opened 2019-01-29
Port documentation

BradyMcd:
Just a simple addition to the documentation. The fact that default port numbers are automatically elided makes perfect sense, but it caught me off and I found wasn't terribly well reflected in the documentation so I added a callout and an example on `port( )` and `set_port( )` function documentations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/481)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -915,7 +915,9 @@ impl Url {
         }
     }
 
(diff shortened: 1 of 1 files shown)


SimonSapin:
Thanks!

@bors-servo r+

Issue #483, opened 2019-02-11
Relaxed parsing mode


vorner:
Hello

I understand the library tries to follow the standard as close as possible.

However, there are URLs out there in the wild that exist, work and are rejected by this library as invalid. As an example, `http://canada-region-70-24-.static-apple-com.center/` (rejected because of the trailing `-`) ‒ if you point a browser there, you'll get a content (not a very useful one, granted).

Would it be possible to have some relaxed parsing mode (if there is, I haven't found it in the documentation) where I would still get the methods to get the host and path and query out of the URL and canonicalize it, while allowing for certain violations against how a good URL looks like? I have little choice over the URLs that I need to handle ‒ basically, whatever happens in the wild might come my way and I have to do something meaningful with it.

nox:
The spec for [parsing hosts](https://url.spec.whatwg.org/#host-parsing) says:

> 5. Let asciiDomain be the result of running domain to ASCII on domain.

The spec for the [domain to ASCII](https://url.spec.whatwg.org/#concept-domain-to-ascii) algorithm says:

> Let result be the result of running Unicode ToASCII with domain_name set to domain, UseSTD3ASCIIRules set to beStrict, CheckHyphens set to false, CheckBidi set to true, CheckJoiners set to true, Transitional_Processing set to false, and VerifyDnsLength set to beStrict.

The spec for [Unicode ToASCII](https://www.unicode.org/reports/tr46/#ToASCII) refers to the [validity criteria](https://www.unicode.org/reports/tr46/#Validity_Criteria) section which says:

> If CheckHyphens, the label must neither begin nor end with a U+002D HYPHEN-MINUS character.

But as mentioned earlier, the Unicode ToASCII algorithm is to be invoked with *CheckHyphens* set to `false`.

Seems like a `rust-url` bug to me.

nox:
https://github.com/servo/rust-url/blob/a1d8c8849fd35098c90a6a05dc772e827303ce41/idna/src/uts46.rs#L253-L262

nox:
Ouch. `idna::uts46::Flags` is a public struct with public fields, so adding one for the `CheckHyphens` flag is a breaking change for the `idna` crate, and `idna::uts46::Errors` is exposed in the `url` crate through the `From<idna::uts46::Errors>` for `url::ParseError`, meaning the breaking change should be propagated as a breaking bump in `url` too. 😕

nox:
[...]

Issue #463, opened 2018-10-17
Tracking issue for rust-url 2.0


SimonSapin:
For a while now we’ve been discussing some possible breaking changes and I’ve been reluctant, to avoid dealing with moving [the ecosystem](https://crates.io/crates/url/reverse_dependencies) to a semver-incompatible version. However it’s probably time to do it anyway, in particular to get serde 1.0 support in the `url` crate itself. To avoid doing *multiple* such migrations, it would be nice batch breaking changes together.

# This is blocked on

* [x] Having a viable migration path for Servo

  * [x] Upgrading Servo from Hyper 0.11 (which uses rust-url 1.x) https://github.com/servo/servo/issues/19825
  * [ ] Upgrading https://crates.io/crates/webdriver
  * [ ] Upgrading https://crates.io/crates/ws
  * [x] Ideally we would demonstrate building Servo with a `v2.0` branch of rust-url before publishing rust-url 2.0.0 to crates.io.

# This would unblock

* [x] Removing `serde_url`
  * [x] Closing https://github.com/servo/rust-url/pull/386
  * [x] Closing https://github.com/servo/rust-url/issues/420
* [x] Closing https://github.com/servo/rust-url/pull/446
* [ ] https://github.com/servo/servo/issues/20412

# Changes to consider

Not necessarily blockers for 2.0.0.

## Update with current spec and tests

… because this might lead to API or behavior changes. 

* [ ] https://github.com/servo/rust-url/issues/163
* [ ] https://github.com/servo/rust-url/issues/290

## Breaking API changes

* [ ] https://github.com/servo/rust-url/issues/299
* [x] https://github.com/servo/rust-url/issues/304 / https://github.com/servo/rust-url/issues/303
* [ ] Rename `Url::join`? https://github.com/servo/rust-url/issues/333
* [ ] https://github.com/servo/rust-url/issues/375
* [ ] https://github.com/servo/rust-url/pull/380 / https://github.com/servo/rust-url/issues/363
* [x] https://github.com/servo/rust-url/issues/388
* [x] https://github.com/servo/rust-url/issues/444 (remove rust-encoding support)
* [x] https://github.com/servo/rust-url/issues/455 (bump Rust version requirement)
* [ ] https://github.com/servo/rust-url/issues/460
* [ ] https://github.com/servo/rust-url/issues/462
* [x] #506
* [x] Remove everything marked as deprecated
* [x] Stop re-exporting entire crates, make crate dependencies private where possible
[...]

PR #484, opened 2019-02-11
Implement CheckHyphens

nox:
Fixes https://github.com/servo/rust-url/issues/483

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/484)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ heap_size = ["heapsize"]
 [dependencies]
 encoding = {version = "0.2", optional = true}
 heapsize = {version = ">=0.4.1, <0.5", optional = true}
-idna = { version = "0.1.0", path = "./idna" }
+idna = { version = "0.2.0", path = "./idna" }
 matches = "0.1"
 percent-encoding = { version = "1.0.0", path = "./percent_encoding" }
 rustc-serialize = {version = "0.3", optional = true}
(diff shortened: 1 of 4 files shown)


nox:
I need to write a test but lunch time is approaching. :)

nox:
I made `idna::domain_to_ascii` not check hyphens, and `idna::uts46::to_ascii` continue to check hyphens by default. Not sure that this is the right trade-off to make.

SimonSapin:
UTS 46 does not specify defaults for its flags, but checking hyphens is what previous versions did:

* https://www.unicode.org/reports/tr46/tr46-19.html#Validity_Criteria
* https://www.unicode.org/reports/tr46/tr46-17.html#Validity_Criteria

The URL standard explicitly sets `CheckHyphens` to false. So yes, I think these are the right defaults.

I think I’d rather not add a new `Config` type in the pubic API as long as `Flags` also exist. What do you think of keeping it private for now? I’ve filed https://github.com/servo/rust-url/issues/486 about cleaning this up on the next breaking change batch, and referenced it from https://github.com/servo/rust-url/issues/463.

nox:
I don't think making `Config` public is an issue, even when `Flags` goes away in a future version `Config` will still be here. I tried to make the change anyway, and I can't use  `pub(crate)` because that stabilised in 1.18.0.

PR #491, opened 2019-03-22
Added fragment percent encode set for URL fragments.

ulf5:
Hello!

I noticed that the spec here defines a fragment percent encode set:
https://url.spec.whatwg.org/#percent-encoded-bytes

> The fragment percent-encode set is the C0 control percent-encode set and U+0020 SPACE, U+0022 ("), U+003C (<), U+003E (>), and U+0060 (`). 

It seemed to me that rust-url was using the SIMPLE_ENCODE_SET while parsing the fragment and that it should be using the fragment encode set instead:
https://url.spec.whatwg.org/#url-parsing

> UTF-8 percent encode c using the fragment percent-encode set and append the result to url’s fragment. 

I made this PR which should fix it, or have I got it wrong?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/491)

### DIFF
--- a/percent_encoding/lib.rs
+++ b/percent_encoding/lib.rs
@@ -117,6 +117,14 @@ impl EncodeSet for SIMPLE_ENCODE_SET {
 }
 
 define_encode_set! {
+    /// This encode set is used in the URL parser for fragments.
+    ///
+    /// Aside from special chacters defined in the [`SIMPLE_ENCODE_SET`](struct.SIMPLE_ENCODE_SET.html),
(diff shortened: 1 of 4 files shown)


nox:
```
thread '"http://f:21/ b ? d # e " @ base "http://example.org/foo/bar"' panicked at '
"http://f:21/%20b%20?%20d%20# e"
!= expected.href
"http://f:21/%20b%20?%20d%20#%20e"
for URL "http://f:21/%20b%20?%20d%20# e"
'
```

Yep, upstream tests were changed and we are just not aligned anymore.

PR #496, opened 2019-05-04
allowing bidi domain names to start with a digit

rtf-const:
This pull request resolves the issue #489

Currently rust-url marks as invalid urls that contains bidirectional characters (domains that contain symbols both in RLT and LTR encoding) if they contains segments starting with a number.
As an example URL
http://mail.163.com.xn----9mcjf9b4dbm09f.com/iloystgnjfrgthteawvo/indexx.php
Attept to parse that url will return an error "invalid international domain name", while that url is valid and existed (url contained some malitious content and doesn't exist anymore)
While urls  http://mail.com.xn----9mcjf9b4dbm09f.com/iloystgnjfrgthteawvo/indexx.php as well as http://mail.163.com/iloystgnjfrgthteawvo/indexx.php are considered as OK. 
I guess there is some contradiction bettween different RFCs. This crate implements bidi (bidirectional unicode chars rules) check based on RFC-5893 https://tools.ietf.org/html/rfc5893#section-2 whicth says  that the first label in the segment name must be the left-to-right char, riht-to-left-char or "Arabic letter" but not "European Number"
While RFC-1123 from 1989 that specify "Requirements for Internet Hosts" explicily says that a digit is a permited domain name start https://tools.ietf.org/html/rfc1123#section-2
> the restriction on the first character is relaxed to allow either a
      letter or a digit.  Host software MUST support this more liberal
      syntax.

So, using using number starting subparts inside bidirectional subdomains sould be allowed, otherwise we sholud treat any number-started domain name (for example http://37signals.com/) as invalid

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/496)

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -130,7 +130,7 @@ fn passes_bidi(label: &str, is_bidi_domain: bool) -> bool {
 
     match first_char_class {
         // LTR label
-        BidiClass::L => {
+        BidiClass::L | BidiClass::EN => {
             // Rule 5
             loop {
                 match chars.next() {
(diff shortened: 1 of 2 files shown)


rtf-const:
It's my fault I wrongly interpreted RFC

SimonSapin:
Please note that this library implements the specification at https://url.spec.whatwg.org/ (and specifically https://url.spec.whatwg.org/#concept-domain-to-ascii, which is defined based on https://www.unicode.org/reports/tr46/#ToASCII).

PR #497, opened 2019-05-07
bidi rules

rtf-const:
This is a pull request for issue #489

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/497)

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -115,7 +115,7 @@ fn map_char(codepoint: char, flags: Flags, output: &mut String, errors: &mut Vec
 }
 
 // http://tools.ietf.org/html/rfc5893#section-2
-fn passes_bidi(label: &str, is_bidi_domain: bool) -> bool {
+fn passes_bidi(label: &str, is_bidi_domain: bool, after_rtl_label: &mut bool) -> bool {
     // Rule 0: Bidi Rules apply to Bidi Domain Names: a name with at least one RTL label.  A label
     // is RTL if it contains at least one character of bidi class R, AL or AN.
     if !is_bidi_domain {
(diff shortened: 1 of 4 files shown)


idna/tests/uts46.rs
```diff
@@ -37,6 +40,13 @@ pub fn collect_tests<F: FnMut(String, TestFn)>(add_test: &mut F) {
         let to_ascii = pieces.remove(0);
         let nv8 = if pieces.len() > 0 { pieces.remove(0) } else { "" };
 
+        for skip in SKIP_TEST {
+            if original == *skip {
```
vorner: Can this be done with `SKIP_TEST.contains(original)`?

idna/tests/uts46.rs
```diff
@@ -10,6 +10,9 @@ use std::char;
 use idna::uts46;
 use test::TestFn;
 
+const SKIP_TEST: &'static [&'static str] = &["0A.\\u05D0", "0a.\\u05D0", "0a.xn--4db"
+    ,"1.걾6.𐱁\\u06D0", "1.걾6.𐱁\\u06D0","1.xn--6-945e.xn--glb1794k"];
```
vorner: Probably explaining why these are skipped in a comment would be nice ‒ maybe linking to the issue?

idna/src/uts46.rs
```diff
@@ -223,6 +225,36 @@ fn passes_bidi(label: &str, is_bidi_domain: bool) -> bool {
...
+                match last {
+                    Some(c) => {
+                        if !c.is_ascii() {
+                            return false;
+                        }
+                        last = chars.next();
+                    }
+                    _ => { break; }
+                }
+            }
```
vorner: This loop doesn't really check if it is truly a LDH label. I think it would accept something like `3_4` as well (`_` is ascii, but neither letter, digit or hyphen). Does something else check that in the existing code?

hsivonen:
`main` branch is now different enough that this PR doesn't apply. However, it doesn't make sense to rewrite the PR at this time, since the issue is still pending in Unicode Technical Committee. Closing the PR and leaving the corresponding issue open.

PR #502, opened 2019-06-07
Explicit unnamed member example

bestouff:
Be a bit more explicit in the use of the `serde` attribute.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/502)


jdm:
@bors-servo r+
Thanks!

PR #508, opened 2019-07-13
Fix warnings and increase MSRV for 2.0

### LINKED ISSUE #455: Use of std::ascii::AsciiExt causes warnings
std::ascii::AsciiExt has been deprecated since 1.26.0 and travis build shows warnings about them at compiling. See https://travis-ci.org/servo/rust-url/jobs/400867523 for example.

I think we should replace them with inherent method calls or at least suppress the warnings.

est31:
This includes some warning fixes as well as a MSRV increase from 1.17.0 to 1.24.0 (which is the oldest release the replacements for the deprecated functionalities are available). The first commit should be least controversial as it doesn't require any MSRV increase. Please say whether MSRV increases are a problem, if they are I can put the first commit into a separate PR.

Fixes #455

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/508)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 
 jobs:
   include:
-  - rust: 1.17.0
+  - rust: 1.24.0
     install:
     # --precise requires Cargo.lock to already exist
     - cargo update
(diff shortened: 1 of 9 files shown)


est31:
Hmm I think I missed some AsciiExt occurences, don't approve yet...

est31:
Ok, all AsciiExt gone now. It can be reviewed!

est31:
Ohhh this is the wrong branch, sorry

PR #509, opened 2019-07-13
Remove the deprecated log_syntax_violation function for 2.0

est31:
I've also removed the `OldFn` variant as that function was the only place that variant was constructed. Does it make sense to rename the `NewFn` variant to something else? Idk.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/509)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,20 +211,6 @@ impl<'a> ParseOptions<'a> {
         self
     }
 
-    /// Call the provided function or closure on non-fatal parse errors, passing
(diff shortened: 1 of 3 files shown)


est31:
Hmm it seems that the `ViolationFn` enum was introduced in #433 , before there was an `Option`. Should I bring back the old state of using an `Option`? Should I rename the `ViolationFn` variants instead?

est31:
Wrong branch here, too :)

PR #510, opened 2019-07-13
Fix warnings and increase MSRV for 2.0 

### LINKED ISSUE #455: Use of std::ascii::AsciiExt causes warnings

est31:
This includes some warning fixes as well as a MSRV increase from 1.17.0 to 1.24.0 (which is the oldest release the replacements for the deprecated functionalities are available). The first commit should be least controversial as it doesn't require any MSRV increase. Please say whether MSRV increases are a problem, if they are I can put the first commit into a separate PR.

Fixes #455

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/510)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 
 jobs:
   include:
-  - rust: 1.17.0
+  - rust: 1.24.0
     install:
     # --precise requires Cargo.lock to already exist
     - cargo update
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -11,8 +11,7 @@ jobs:
 
     - cargo update -p unicode-normalization --precise 0.1.5
 
-    # data-url uses pub(crate) which is unstable in 1.17
-    script: cargo test --all-features -p url -p idna -p percent-encoding -p url_serde
+    script: cargo test --all-features -p url -p idna -p percent-encoding -p url_serde -p data-url
```
SimonSapin: I think `-p` arguments can be replaced with `--all`?

PR #511, opened 2019-07-13
Remove the deprecated log_syntax_violation function for 2.0

est31:
I've also removed the `OldFn` variant as that function was the only place that variant was constructed. 
Does that make sense or should I replace the entire enum with `Option`'s everywhere like it was before #433?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/511)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,20 +211,6 @@ impl<'a> ParseOptions<'a> {
         self
     }
 
-    /// Call the provided function or closure on non-fatal parse errors, passing
-    /// a static string description.  This method is deprecated in favor of
-    /// `syntax_violation_callback` and is implemented as an adaptor for the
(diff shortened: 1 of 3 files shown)


dekellum:
If you replaced the enum with an Option, you'd still need to wrap in a new-type (generic over the lifetime) to implement the same specific methods for the type, right? Looks good to me, the way you have it.

SimonSapin:
Merged in the 2.0 branch, with an `Option`: 48f3bed3e773b8d9da10f1bb99094575afda4c43.

Thanks!

dekellum:
…or move those methods to `Parser` and don't bother using them from `Input`. That works too, I see, and is also more compact.

est31:
Merge commit was: 90e65b26d4404c97ccfc7597e676070596fa1c9d

PR #516, opened 2019-07-15
Change the closure taken by `with_default_port` to return Option instead

tmccombs:
of Result.

`Result` seems like overkill for the closure, especially now that the
`?` operator has been stabilized for `Option`.

Related to #299

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/516)


tmccombs:
Looks like that function was removed

Issue #480, opened 2019-01-26
Consider adding a bunch of helper methods to the Url type


nox:
I would like to add methods for concepts such as [whether the URL includes credentials](https://url.spec.whatwg.org/#include-credentials).

PR #520, opened 2019-07-18
Add url.includes_credentials() and url.is_special().

o0Ignition0o:
As part of #480 this pull request adds includes_credentials and is_special to url, which are defined [in the Miscellaneous section](https://url.spec.whatwg.org/#url-miscellaneous).

There are probably other cool helper methods to add, but I don't know which ones would make sense.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -474,7 +474,7 @@ impl Url {
                     assert_eq!(host_str, h.to_string())
                 }
                 HostInternal::Domain => {
-                    if SchemeType::from(self.scheme()).is_special() {
+                    if self.is_special() {
                         assert!(!host_str.is_empty())
                     }
                 }
@@ -1566,16 +1566,16 @@ impl Url {
         }
 
         if let Some(host) = host {
-            if host == "" && SchemeType::from(self.scheme()).is_special() {
+            if host == "" && self.is_special() {
                 return Err(ParseError::EmptyHost);
             }
-            if SchemeType::from(self.scheme()).is_special() {
+            if self.is_special() {
                 self.set_host_internal(Host::parse(host)?, None)
             } else {
                 self.set_host_internal(Host::parse_opaque(host)?, None)
             }
         } else if self.has_host() {
-            if SchemeType::from(self.scheme()).is_special() {
+            if self.is_special() {
                 return Err(ParseError::EmptyHost);
             }
             debug_assert!(self.byte_at(self.scheme_end) == b':');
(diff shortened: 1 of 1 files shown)


nox:
A new method for https://url.spec.whatwg.org/#cannot-have-a-username-password-port could be used in `set_username`, `set_password` and `set_port`.

o0Ignition0o:
I decided to refactor the method contents to explicitly match the specification: 
`A URL cannot have a username/password/port if its host is null or the empty string, its cannot-be-a-base-URL flag is set, or its scheme is "file". `
I can of course revert the contents if needed.

o0Ignition0o:
Rebased to master :)

src/lib.rs
```diff
@@ -2103,6 +2102,68 @@ impl Url {
...
+    /// let url = Url::parse("unix:/run/foo.socket")?;
+    /// assert!(!url.is_special());
+    ///
+    /// let url = Url::parse("data:text/plain,Stuff")?;
+    /// assert!(!url.is_special());
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
+    pub fn is_special(&self) -> bool {
```
djc, line 2130: As I understand it, "special" is really about the scheme (which matches the docs), whereas a `Url::is_special()` method seems to be about the whole URL. Should this be renamed to `is_special_scheme()` or `is_scheme_special()` or `has_special_scheme()`?

src/lib.rs
```diff
@@ -2103,6 +2102,68 @@ impl Url {
...
+    /// assert!(!url.includes_credentials());
+    ///
+    /// let url = Url::parse("https://@www.my_site.com")?;
+    /// assert!(!url.includes_credentials());
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
+    pub fn includes_credentials(&self) -> bool {
+        self.username() != "" || self.password().unwrap_or(&"") != ""
```
djc, line 2161: I'm confused about this code, is it even possible to have a password without username?

src/lib.rs
```diff
@@ -2361,6 +2358,12 @@ impl Url {
 
     // Private helper methods:
 
+    fn cannot_have_username_password_port(&self) -> bool {
+        self.host().unwrap_or(Host::Domain("")) == Host::Domain("")
```
djc, line 2362: It would be nice to link https://url.spec.whatwg.org/#url-miscellaneous in the docstring for this method.

djc (commented):
Thanks for working on these helper functions! I have some some questions about the details, but the direction looks great.

PR #523, opened 2019-07-20
Update port on scheme change + host parsing rules to the host setter + hash parsing rules

o0Ignition0o:
On scheme change, we forgot to update the port if it's the new default one.

This will help the test : 

```

        {
            "comment": "Port is set to null if it is the default for new scheme.",
            "href": "http://foo.com:443/",
            "new_value": "https",
            "expected": {
                "href": "https://foo.com/",
                "protocol": "https:",
                "port": ""
            }
        }
```
pass later, once we're done with the forward slash appending when there's no path... which is quite a challenge for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/523)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1954,6 +1954,13 @@ impl Url {
 
         parser.serialization.push_str(self.slice(old_scheme_end..));
         self.serialization = parser.serialization;
+
+        // Update the port so it can be removed
+        // If it is the scheme's default
+        // We don't mind it silently failing
+        // If there was no port in the first place
+        let _ = self.set_port(self.port());
+
         Ok(())
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1569,10 +1569,25 @@ impl Url {
             if host == "" && SchemeType::from(self.scheme()).is_special() {
                 return Err(ParseError::EmptyHost);
             }
+            let mut host_substr = host;
+            // Otherwise, if c is U+003A (:) and the [] flag is unset, then
+            if !host.starts_with('[') || !host.ends_with(']') {
+                match host.find(':') {
+                    Some(0) => {
+                        // If buffer is the empty string, validation error, return failure.
+                        return Err(ParseError::InvalidDomainCharacter);
+                    }
+                    // Let host be the result of host parsing buffer
+                    Some(colon_index) => {
+                        host_substr = &host[..colon_index];
+                    }
+                    None => {}
+                }
+            }
             if SchemeType::from(self.scheme()).is_special() {
-                self.set_host_internal(Host::parse(host)?, None)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1954,6 +1954,13 @@ impl Url {
 
         parser.serialization.push_str(self.slice(old_scheme_end..));
         self.serialization = parser.serialization;
+
+        // Update the port so it ca be removed
```
Eijebong: can* :)

o0Ignition0o: Oh right, I'll blame the keyboard for this nit :p 
Fixing it as soon as I get back home !

o0Ignition0o:
The second commit applies host parsing rules to the host setter. The ':' handling was missing.

This should let the following host example pass: 

```
        {
            "comment": "Port number is unchanged if not specified",
            "href": "http://example.net:8080",
            "new_value": "example.com:",
            "expected": {
                "href": "http://example.com:8080/",
                "host": "example.com:8080",
                "hostname": "example.com",
                "port": "8080"
            }
        },
```
The remaining failing tests seem to mostly involve the forward slash bug.
There's one javascript remaining test I'm hoping to handle though :)

o0Ignition0o:
The last commit edits the hash setting and getting rules to match the [dom url hash specification](https://url.spec.whatwg.org/#dom-url-hash)

This allows the following setter test to pass:
```
        {
            "href": "javascript:alert(1)",
            "new_value": "castle",
            "expected": {
                "href": "javascript:alert(1)#castle",
                "hash": "#castle"
            }
        }
```
I would love to solve the forward slash bug, but I will probably not have enough time to tackle it for now :/

o0Ignition0o:
[a port setter test](https://github.com/web-platform-tests/wpt/blob/master/url/resources/setters_tests.json#L1362)
 seems to be inconsistent with a [protocol setter test](https://github.com/web-platform-tests/wpt/blob/master/url/resources/setters_tests.json#L230)
There doesn't seem to be any consistency in WPT wrt the "/" when there is no path. 
The [specification](https://url.spec.whatwg.org/#url-serializing) seems to dismiss '/' at the end if there is no path: 
" Otherwise, then for each string in url’s path, append U+002F (/) followed by the string to output."

@nox I don't really know how to reliably handle it, is there anything I am missing I should know ?

o0Ignition0o:
Ok the culprit seems to be https://github.com/web-platform-tests/wpt/commit/37a9631f087bec53ecd0b2d6b19251f93a066095#diff-6ef48fcfe9e518dd93d4fe48a795d3ac

@SimonSapin I'll rollback until just before this commit, and push things progressively, working my way step by step

o0Ignition0o:
Subset of of #537

Issue #498, opened 2019-05-08
Option to reject legacy inet_aton notation for IPv4 addresses


zackw:
The current rev of the URL Standard specifies that the various legacy notations accepted by `inet_aton` for IPv4 addresses are validation errors, but rust-url silently accepts them (they're not even diagnosed via the syntax violation callback).  I would like to have an option to reject these.

Examples of undesirable constructs:

```
# all are treated as equivalent to http://127.0.0.1/
http://127.0.1/
http://127.1/
http://0177.0.1/
http://0177.0.0.1/
http://0177.1/
http://0x7f.0.1/
http://0x7f.0.0.1/
http://0x7f.1/
http://2130706433/
http://0x7f000001/
http://017700000001/

# equivalent to http://127.0.0.8/ , not http://127.0.0.10/
http://0127.0.0.010/
```

See https://url.spec.whatwg.org/#concept-ipv4-parser (pay close attention to where it, and its subroutine https://url.spec.whatwg.org/#ipv4-number-parser , set the "validationErrorFlag") for gory details.

Previous discussion in #116.

dekellum:
I wasn't aware of this legacy edge case (thanks!), but I agree with your motivation.   Seems exploitable for misuse. For example servers might have good reason to want to reject provided URLs to localhost.  Do common IPv4 resolvers actually accept these

If the violation callback was extended for this case, you would have the option of promoting the violation to your own error or other form of rejection?  That would be a light-weight and (I think, reasonably) backward-compatible way to achieve it, while still making the implementation available in this crate. I'd use that!

zackw:
That would work for me except that in #462 it’s proposed to remove the
syntax violation callback.

dekellum:
Indeed! I've suggested keeping it. :)

SimonSapin:
Aligning with the spec in terms of what inputs log a syntax violation sounds good.

zackw:
@dekellum re
> Do common IPv4 resolvers actually accept and resolve these examples to local ports?

The [POSIX spec for `getaddrinfo`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getaddrinfo.html) says that "address strings using Internet standard dot notation as specified in [inet_addr](https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_addr.html) are valid", and that does include all of the unusual syntaxes I listed.

dekellum:
[...]

PR #525, opened 2019-07-22
Add unused variants to ParseError and SyntaxViolation enums

dekellum:
Per discussion in https://github.com/servo/rust-url/issues/498#issuecomment-513299432 , this adds a hidden unused variant to each of `ParseError` and `SyntaxViolation` enums so as to make future additions non-breaking changes.   This is the long standing manual way (e.g. originally used by`std::io::ErrorKind`), given lack of a stable feature and acceptable rust MSRV with rust-lang/rust#44109.

Also these types already implemented `Display` but having been burnt with `Debug` vs `Display` versions of the `fmt` function, I also disambiguated paths in these `Display` implementations.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/525)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -49,11 +49,17 @@ pub type ParseResult<T> = Result<T, ParseError>;
 macro_rules! simple_enum_error {
     ($($name: ident => $description: expr,)+) => {
         /// Errors that can occur during parsing.
+        ///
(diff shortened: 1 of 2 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #526, opened 2019-07-22
Remove redundant field names in struct init

dekellum:
Besides the obvious minimization, these changes avoid lints of the following form from clippy:

``` txt
    warning: redundant field names in struct initialization
      --> src/form_urlencoded.rs:30:13
       |
    30 |     Parse { input: input }
       |             ^^^^^^^^^^^^ help: replace it with: `input`
       |
       = help: for further information visit
    https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/526)

### DIFF
--- a/src/form_urlencoded.rs
+++ b/src/form_urlencoded.rs
@@ -27,7 +27,7 @@ use std::str;
 /// converted to `[("#first", "%try%")]`.
 #[inline]
 pub fn parse(input: &[u8]) -> Parse {
-    Parse { input: input }
+    Parse { input }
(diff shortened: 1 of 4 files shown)


SimonSapin:
This code probably dates from before this syntax existed in the language :)

I generally don’t use Clippy and being "clippy-clean" is not at all a goal, but this particular change looks good to me. Thanks.

@bors-servo r+

Issue #532, opened 2019-07-30
v2.0 upgrade path


benesch:
First of all, thanks for maintaining this crate! Maintaining an open source project can be thankless work, and especially so when it's a foundational crate like this one.

I recently embarked on upgrading reqwest to url v2.0 (https://github.com/seanmonstar/reqwest/pull/583), and ran into some roadblocks. These roadblocks would have been greatly eased if this crate maintained a changelog or release notes. (If it does, and I've just missed it, apologies! But it's awfully hard to find in that case.) It's very hard to figure out what's changed when upgrading across backwards-incompatible changes without a changelog.

The commit messages don't do a great job of explaining rationale, either. Here's a quick example: https://github.com/servo/rust-url/commit/fe74a60bd0636c5e5da920674b9bbffc22f3c384. Why were the public exports of percent_encoding removed? It's no longer clear to me how to ensure that the version of `percent_encoding` I'm using matches the version used by `url`. I poked at the PR description (https://github.com/servo/rust-url/pull/517), too, and it's no more helpful in explaining the rationale. There's a brief mention of the work item in the 2.0 tracking issue (https://github.com/servo/rust-url/issues/463), but again doesn't present rationale.

The [upgrade guide for 0.1 -> 1.0](https://github.com/servo/rust-url/blob/master/UPGRADING.md), by contrast, is fantastic! I'd like to put one together for 1.0 -> 2.0, but I'll need some help to do so.

Some other issues that could use explaining:

  * percent_encoding no longer exports useful character classes. I had to dig through closed issues to find the rationale for this: https://github.com/servo/rust-url/issues/529;
  * the removal of ToSocketAddr, which is also explained in a closed issue: https://github.com/servo/rust-url/issues/530.

[...]

PR #533, opened 2019-07-30
Restore implementation for ToSocketAddrs

benesch:
I thought of this while writing up #532. 

url::Url's ToSocketAddrs implementation was dropped because of its large
API surface. Restore it with a simpler implementation that does not
introduce as much API surface. The upside is that the HostAndPort and
SocketAddrs helper types are no longer necessary. The downside is that
if the URL specifies an IP address, the ToSocketAddrs implementation has
to parse the IP address from a string again.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/533)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,11 +123,13 @@ use std::cmp;
 use std::error::Error;
 use std::fmt::{self, Write};
 use std::hash;
+use std::io;
 use std::mem;
-use std::net::IpAddr;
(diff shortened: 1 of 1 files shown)


SimonSapin:
Alright, #532 and this PR make a convincing case that this functionality should be present. I had some comments on API details, and writing another PR felt easier. Closing in favor of https://github.com/servo/rust-url/pull/535, please comment there about its proposed API. Thanks anyway for submitting this PR!

Issue #290, opened 2017-03-10
Align with latest of web-platform-tests


annevk:
Please correct me if I'm wrong, but we've made numerous changes to the URL Standard lately and also added tons of new tests (both for the API setters and the URL parser), and I've the impression that these changes haven't made their way to rust-url yet.

It might be good to investigate that, especially with rust-url getting more usage in Firefox over time.

SimonSapin:
You’re not wrong. I’ve had the same worry and filed #163… more than a year ago.

annevk:
Ah okay, I guess you can close this as a duplicate then. https://github.com/jsdom/whatwg-url indicates which commit they are aligned with and which version of the test suite they are aligned with.

nox:
So I locally updated the tests and…:

```
running 713 tests
............thread '"#x" @ base "sc://ñ"' panicked at '
"sc://%C3%B1/#x"
!= expected.href
"sc://%C3%B1#x"
for URL "sc://%C3%B1/#x"
', tests/data.rs:69:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
.........F...........thread '"/..//localhost//pig" @ base "file://lion/"' panicked at '
"file:////localhost//pig"
!= expected.href
"file://lion/localhost//pig"
for URL "file:////localhost//pig"
', tests/data.rs:69:5
...thread '"////" @ base "sc://x/"' panicked at '
"sc:///"
!= expected.href
"sc:////"
for URL "sc:///"
', tests/data.rs:69:5
thread '"/////mouse" @ base "file:///elephant"' panicked at '
"file://///mouse"
!= expected.href
"file:///mouse"
for URL "file://///mouse"
', tests/data.rs:69:5
..thread '"////x/" @ base "sc://x/"' panicked at '
"sc://x/"
!= expected.href
"sc:////x/"
for URL "sc://x/"
', tests/data.rs:69:5.
..F...Fthread '"//localhost//pig" @ base "file://lion/"' panicked at '
"file:////pig"
!= expected.href
"file:///pig"
for URL "file:////pig"
', tests/data.rs:69:5
FF.thread '"/?chai" @ base "file://tea/"' panicked at '
"file:///?chai"
!= expected.href
"file://tea/?chai"
for URL "file:///?chai"
', tests/data.rs:69:5
...F..thread '"/c:/foo/bar" @ base "file:///c:/baz/qux"' panicked at '
"file:///c:/c:/foo/bar"
!= expected.href
[...]

Issue #163, opened 2016-01-28
Track spec changes somehow


SimonSapin:
1. Pick a commit of http://url.spec.whatwg.org/
2. Go through all of it, fix rust-url where we differ or file issues. This will be slow and boring and needs to be done very carefully.
3. Keep track of that which commit of the spec we’re up-to-date with, modulo filed issues
4. Regularly, go through new commits on the spec repository and update code or file new issues for spec changes

gerv:
2. needs to involve writing tests, presumably.

SimonSapin:
We run a copy of the tests at https://github.com/w3c/web-platform-tests/tree/master/url. If some start failing after we update them, that’s one way to find out that we probably need some code change. But ideally we’d still go through the spec’s text in case some aspect is not covered by the test suite. In this case, yes, we should write new tests and contribute them upstream.

SimonSapin:
(Well, not all of https://github.com/w3c/web-platform-tests/tree/master/url since we don’t have a DOM in this repo’s unit tests, but the subset based on JSON data.)

djc:
I've updated the JSON to URL Standard's current master in #623.

PR #534, opened 2019-07-30
Add a changelog and an upgrade guide for v2.0.

### LINKED ISSUE #532: v2.0 upgrade path

benesch:
See #532 for context. Having this available would have been a huge help when upgrading to url v2.0, and I wanted to save others the pain. I left a few TODOs where I'll need someone else to explain the impact of a change.

Fix #532.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/534)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,4 +1,113 @@
-# Guide to upgrading from url 0.x to 1.x
+# Upgrade guide
+
+This guide contains steps for upgrading crates in this project between major
+versions. Only the most common issues are covered here. For full details of
+changes, see the [changelog](CHANGELOG.md).
+
+## Upgrading from url 1.x to 2.x
+
+* The minimum supported Rust version is now v1.33.0. Verify that you can bump
+  your library or application to the same MSRV.
+
+* `Url` no longer implements `std::net::ToSocketAddrs`. You will instead need to
+  explicitly convert your `Url` to a type that implements `ToSocketAddrs`.
+
+    Before upgrading:
+
+    ```rust
+    let url = Url::parse("http://github.com:80").unwrap();
+    let stream = TcpStream::connect(url).unwrap();
+    ```
+
+    After upgrading:
+
+    ```rust
+    let url = Url::parse("http://github.com:80").unwrap();
+    let port = url.port_or_known_default().unwrap();
+    let addrs;
+    let addr;
+    let addrs = match url.host().unwrap() {
+        url::Host::Domain(domain) => {
+            addrs = (domain, port).to_socket_addrs().unwrap();
+            addrs.as_slice()
+        }
+        url::Host::Ipv4(ip) => {
+            addr = (ip, port).into();
+            std::slice::from_ref(&addr)
+        }
+        url::Host::Ipv6(ip) => {
+            addr = (ip, port).into();
+            std::slice::from_ref(&addr)
+        }
+    };
+    let stream = TcpStream::connect(addrs).unwrap();
+    ```
+
+* `url_serde` is no longer required to use `Url` with Serde 1.x. Remove
+  references to `url_serde` and enable the `serde` feature instead.
+
+     ```toml
+     # Cargo.toml
+     [dependencies]
+     url = { version = "2.0", features = ["serde"] }
+     ```
+
+* The `idna` and `percent_export` crates are no longer exported by the `url`
+  crate. Depend on those crates directly instead. See below for additional
+  breaking changes in the percent-export package.
+
+    Before upgrading:
+
+    ```rust
+    use url::percent_encoding::percent_decode;
+    ```
+
+    After upgrading:
+
+    ```rust
+    use percent_encoding::percent_decode;
+    ```
+
+## Upgrading from percent-encoding 1.x to 2.x
+
+* Prepackaged encoding sets, like `QUERY_ENCODE_SET` and
+  `PATH_SEGMENT_ENCODE_SET`, are no longer provided. You will need to read You
+  will need to read the specifications relevant to your domain and construct
+  your own encoding sets by using the `percent_encoding::AsciiSet` builder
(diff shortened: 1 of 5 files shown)


SimonSapin:
Thank you for taking the time to write all this up. I agree that a section for 2.x in `UPGRADING.md` will be good to have. However:

> All notable changes to the […] package will be documented in this file.

Until this project is has a long-term maintainer who wants to commit to maintain changelogs, that statement is simply not true. As such, I would prefer not to have `CHANGELOG.md` files.

UPGRADING.md
```diff
@@ -1,4 +1,113 @@
...
+versions. Only the most common issues are covered here. For full details of
+changes, see the [changelog](CHANGELOG.md).
+
+## Upgrading from url 1.x to 2.x
+
+* The minimum supported Rust version is now v1.33.0. Verify that you can bump
+  your library or application to the same MSRV.
+
+* `Url` no longer implements `std::net::ToSocketAddrs`. You will instead need to
+  explicitly convert your `Url` to a type that implements `ToSocketAddrs`.
```
SimonSapin: Assuming some form of https://github.com/servo/rust-url/pull/535 lands, this should reference it instead.

UPGRADING.md
```diff
@@ -1,4 +1,113 @@
...
+    /// https://url.spec.whatwg.org/#path-percent-encode-set
+    const PATH_ENCODE_SET: &AsciiSet = &FRAGMENT_ENCODE_SET
+        .add(b'#').add(b'?').add(b'{').add(b'}');
+
+    const PATH_SEGMENT_ENCODE_SET: &AsciiSet = &PATH_ENCODE_SET.add(b'/').add(b'%');
+
+    percent_encoding::utf8_percent_encode(value, PATH_SEGMENT_ENCODE_SET);
+    ```
+
+  You can use the [private encoding sets in the URL parser][encoding-sets] as
```
SimonSapin: I would prefer not to recommend code that is known to be out of date compared to the specification it implements. See https://github.com/servo/rust-url/issues/163 and https://github.com/servo/rust-url/issues/290.

UPGRADING.md
```diff
@@ -1,4 +1,113 @@
...
+    ```rust
+    /// https://url.spec.whatwg.org/#fragment-percent-encode-set
+    const FRAGMENT_ENCODE_SET: &AsciiSet = &percent_encoding::CONTROLS
+        .add(b' ').add(b'"').add(b'<').add(b'>').add(b'`');
+
+    /// https://url.spec.whatwg.org/#path-percent-encode-set
+    const PATH_ENCODE_SET: &AsciiSet = &FRAGMENT_ENCODE_SET
+        .add(b'#').add(b'?').add(b'{').add(b'}');
+
+    const PATH_SEGMENT_ENCODE_SET: &AsciiSet = &PATH_ENCODE_SET.add(b'/').add(b'%');
```
SimonSapin: Since `PATH_SEGMENT_ENCODE_SET` is not directly in a specification, let’s remove it and `PATH_ENCODE_SET` here and have the example use `FRAGMENT_ENCODE_SET`. Let’s also call it `FRAGMENT_PERCENT_ENCODE_SET` like in the spec, or just `FRAGMENT`.

UPGRADING.md
benesch: Well, the whole crate is out of date, isn't it? 😄 But sure, I see your point. Removed.

SimonSapin:
> I find it pretty worrying that this crate doesn't have a maintainer!

Welcome to open-source ¯\_(ツ)_/¯

SimonSapin:
Moved  to #536

PR #535, opened 2019-07-31
Restore DNS resolution functionality from url 1.x

SimonSapin:
This is an alternative to https://github.com/servo/rust-url/pull/533.

I feel that an inherent method is more appropriate than an impl of the `ToSocketAddrs` trait: it is not the URL as a whole that is converted, only parts of it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/535)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "2.0.0"
+version = "2.1.0"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
--- a/idna/tests/punycode.rs
+++ b/idna/tests/punycode.rs
@@ -7,8 +7,8 @@
 // except according to those terms.
 
 use idna::punycode::{decode, encode_str};
-use serde_json::Value;
 use serde_json::map::Map;
+use serde_json::Value;
 use std::str::FromStr;
 use test::TestFn;
 
--- a/idna/tests/tests.rs
+++ b/idna/tests/tests.rs
@@ -1,6 +1,6 @@
 extern crate idna;
-extern crate serde_json;
 extern crate rustc_test as test;
+extern crate serde_json;
 
 mod punycode;
 mod uts46;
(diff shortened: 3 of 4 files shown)


SimonSapin:
r=nox per IRC discussion.

@benesch what do you think of this API?

benesch:
> I feel that an inherent method is more appropriate than an impl of the ToSocketAddrs trait: it is not the URL as a whole that is converted, only parts of it.

Perhaps you have a more nuanced understanding of what it means to implement `ToSocketAddrs` than I do. The Rust docs say ToSocketAddrs is "a trait for objects which can be converted or resolved to one or more SocketAddr values", which seems applicable enough to `Url`. To be honest, I'm not sure I see how a method named `socket_addrs` on `Url` is better at conveying that only part of the URL is used in the conversion.

I'd also like to point out that, in #533, implementing `ToSocketAddrs` on `Url` itself isn't strictly required. We could just restore a slightly tweaked `with_default_port` method instead:

```diff
+    pub fn with_default_port<F>(&self, f: F) -> io::Result<(&str, u16)>
+    where
+        F: FnOnce(&Url) -> Result<u16, ()>,
+    {
+        Ok((self
+                .host_str()
+                .ok_or(())
+                .or_else(|()| io_error("URL has no host"))?,
+            self
+                .port_or_known_default()
+                .ok_or(())
+                .or_else(|()| f(self))
+                .or_else(|()| io_error("URL has no port number"))?,
+        ))
+    }
+
```

As I mentioned in #533, this approach has the definite defect that it may cause an already-parsed IP address to be reparsed, a defect which is not present in this PR. (Although the extra vec allocation might mean that it's a wash.) But it has the upside that a lot of code written against 1.0 that used `with_default_port` would just work with this implementation with no code changes.

But! I'm happy enough with the approach in this PR! I'm quite glad I was able to convince you to restore this functionality at all. :)

SimonSapin:
It’s a bit hand-wavy, I interpret “objects which can be converted or resolved to” as the object *is* something that can be converted, whereas I view URLs as *having* components that can be converted. They also have other components that are not relevant to that conversion.

@bors-servo r=nox

SimonSapin:
This is in https://crates.io/crates/url/2.1.0

PR #536, opened 2019-08-01
Add an upgrade guide for v2.0.

### LINKED ISSUE #532: v2.0 upgrade path

benesch:
Extracted from #534, with review feedback addressed. This variant omits the changelogs in an effort to introduce less of a maintenance burden.

Fix #532.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/536)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,4 +1,108 @@
-# Guide to upgrading from url 0.x to 1.x
+# Upgrade guide
+
+This guide contains steps for upgrading crates in this project between major
+versions.
+
+## Upgrading from url 1.x to 2.1+
+
+* The minimum supported Rust version is now v1.33.0. Verify that you can bump
+  your library or application to the same MSRV.
+
+* `Url` no longer implements `std::net::ToSocketAddrs`. You will instead need to
+  explicitly call `socket_addrs` to convert your `Url` to a type that implements
+  `ToSocketAddrs`.
+
+  Note that v2.0 removed support for `std::net::ToSocketAddrs` with no
+  replacement; the `socket_addrs` method was not added until v2.1.
+
+    Before upgrading:
+
+    ```rust
+    let url = Url::parse("http://github.com:80").unwrap();
+    let stream = TcpStream::connect(url).unwrap();
+    ```
+
+    After upgrading:
+
+    ```rust
+    let url = Url::parse("http://github.com:80").unwrap();
+    let addrs = url.socket_addrs(|| None).unwrap();
+    let stream = TcpStream::connect(addrs).unwrap();
+    ```
+
+    Before upgrading:
+
+    ```rust
+    let url = Url::parse("socks5://localhost").unwrap();
+    let stream = TcpStream::connect(url.with_default_port(|url| match url.scheme() {
+        "socks5" => Ok(1080),
+        _ => Err(()),
+    })).unwrap();
+    ```
+
+    After upgrading:
+
+    ```rust
+    let url = Url::parse("http://github.com:80").unwrap();
+    let stream = TcpStream::connect(url.socket_addrs(|| match url.scheme() {
+        "socks5" => Some(1080),
+        _ => Err(()),
+    })).unwrap();
+    ```
+
+* `url_serde` is no longer required to use `Url` with Serde 1.x. Remove
+  references to `url_serde` and enable the `serde` feature instead.
+
+     ```toml
+     # Cargo.toml
+     [dependencies]
+     url = { version = "2.0", features = ["serde"] }
+     ```
+
+* The `idna` and `percent_export` crates are no longer exported by the `url`
+  crate. Depend on those crates directly instead. See below for additional
+  breaking changes in the percent-export package.
+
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -1,4 +1,108 @@
...
+    After upgrading:
+
+    ```rust
+    use percent_encoding::percent_decode;
+    ```
+
+## Upgrading from percent-encoding 1.x to 2.x
+
+* Prepackaged encoding sets, like `QUERY_ENCODE_SET` and
+  `PATH_SEGMENT_ENCODE_SET`, are no longer provided. You will need to read You
```
o0Ignition0o:
```suggestion
  `PATH_SEGMENT_ENCODE_SET`, are no longer provided. You
```

benesch: Thanks!

benesch:
@SimonSapin, I believe this is ready to be merged.

UPGRADING.md
```diff
@@ -1,4 +1,116 @@
...
+    const FRAGMENT_ENCODE_SET: &AsciiSet = &percent_encoding::CONTROLS
+        .add(b' ').add(b'"').add(b'<').add(b'>').add(b'`');
+
+    /// https://url.spec.whatwg.org/#path-percent-encode-set
+    const PATH_ENCODE_SET: &AsciiSet = &FRAGMENT_ENCODE_SET
+        .add(b'#').add(b'?').add(b'{').add(b'}');
+
+    const PATH_SEGMENT_ENCODE_SET: &AsciiSet = &PATH_ENCODE_SET.add(b'/').add(b'%');
+
+    percent_encoding::utf8_percent_encode(value, PATH_SEGMENT_ENCODE_SET);
```
SimonSapin: I think https://github.com/servo/rust-url/pull/534#discussion_r309230834 is still relevant.

UPGRADING.md
```diff
@@ -1,4 +1,116 @@
-# Guide to upgrading from url 0.x to 1.x
+# Upgrade guide
+
+This guide contains steps for upgrading crates in this project between major
+versions. Only the most common issues are covered here. For full details of
+changes, see the [changelog](CHANGELOG.md).
```
SimonSapin: Since there is no `CHANGELOG.md` let’s remove these two sentences.

benesch: Done.

UPGRADING.md
benesch: Fixed. I used `QUERY` instead of `FRAGMENT` because `FRAGMENT_ENCODE_SET` didn't actually exist in the old percent_encoding, while `QUERY_ENCODE_SET` did.

SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #537, opened 2019-08-02
Pass "new" tests 

o0Ignition0o:
current status : 

```
test result: ok. 63 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Please note I tried my best to follow the spec, but it sometimes (often?) felt like I was just hacking my way until tests passed.

I know the PR is huge, I would love to split it (and then squash it) if it makes sense.

Mentioning @nox , @SimonSapin and @valenting because they're the contributors I've seen the most in the commit logs :)

Please let me know if there's anything I can do to improve the PR, and thank you for your time!
 

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/537)

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -24,9 +24,13 @@ pub(crate) enum HostInternal {
     Ipv6(Ipv6Addr),
 }
 
-impl<S> From<Host<S>> for HostInternal {
+impl<S> From<Host<S>> for HostInternal
+where
+    S: ToString,
+{
     fn from(host: Host<S>) -> HostInternal {
         match host {
+            Host::Domain(ref s) if s.to_string().is_empty() => HostInternal::None,
             Host::Domain(_) => HostInternal::Domain,
             Host::Ipv4(address) => HostInternal::Ipv4(address),
             Host::Ipv6(address) => HostInternal::Ipv6(address),
--- a/src/host.rs
+++ b/src/host.rs
@@ -24,13 +24,10 @@ pub(crate) enum HostInternal {
     Ipv6(Ipv6Addr),
 }
 
-impl<S> From<Host<S>> for HostInternal
-where
-    S: ToString,
-{
-    fn from(host: Host<S>) -> HostInternal {
+impl From<Host<String>> for HostInternal {
+    fn from(host: Host<String>) -> HostInternal {
         match host {
-            Host::Domain(ref s) if s.to_string().is_empty() => HostInternal::None,
+            Host::Domain(ref s) if s.is_empty() => HostInternal::None,
             Host::Domain(_) => HostInternal::Domain,
             Host::Ipv4(address) => HostInternal::Ipv4(address),
             Host::Ipv6(address) => HostInternal::Ipv6(address),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1403,12 +1403,8 @@ impl Url {
                 }
                 parser.parse_cannot_be_a_base_path(parser::Input::new(path));
             } else {
-                let path_start = parser.serialization.len();
                 let mut has_host = true; // FIXME
                 parser.parse_path_start(scheme_type, &mut has_host, parser::Input::new(path));
-                if scheme_type.is_file() {
-                    parser::trim_path(&mut parser.serialization, path_start);
-                }
             }
         });
         self.restore_after_path(old_after_path_pos, &after_path);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2058,12 +2058,12 @@ impl Url {
         let new_scheme_type = SchemeType::from(&parser.serialization);
         let old_scheme_type = SchemeType::from(self.scheme());
         // If url’s scheme is a special scheme and buffer is not a special scheme, then return.
-        if new_scheme_type.is_special() && !old_scheme_type.is_special() ||
+        if (new_scheme_type.is_special() && !old_scheme_type.is_special()) ||
             // If url’s scheme is not a special scheme and buffer is a special scheme, then return.
-            !new_scheme_type.is_special() && old_scheme_type.is_special() ||
+            (!new_scheme_type.is_special() && old_scheme_type.is_special()) ||
             // If url includes credentials or has a non-null port, and buffer is "file", then return.
             // If url’s scheme is "file" and its host is an empty host or null, then return.
-            new_scheme_type.is_file() && self.has_authority()
+            (new_scheme_type.is_file() && self.has_authority())
         {
             return Err(());
         }
@@ -2095,8 +2095,8 @@ impl Url {
 
         // Update the port so it can be removed
         // If it is the scheme's default
-        // We don't mind it silently failing
-        // If there was no port in the first place
+        // we don't mind it silently failing
+        // if there was no port in the first place
         let previous_port = self.port();
         let _ = self.set_port(previous_port);
 
@@ -2575,7 +2575,7 @@ fn file_url_segments_to_pathbuf(
     }
     // A windows drive letter must end with a slash.
     if bytes.len() > 2 {
-        if matches!(bytes[bytes.len() -2], b'a'..=b'z' | b'A'..=b'Z')
+        if matches!(bytes[bytes.len() - 2], b'a'..=b'z' | b'A'..=b'Z')
             && matches!(bytes[bytes.len() - 1], b':' | b'|')
         {
             bytes.push(b'/');
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -456,13 +456,15 @@ impl Url {
 
         if self.slice(self.scheme_end + 1..).starts_with("//") {
             // URL with authority
-            match self.byte_at(self.username_end) {
-                b':' => {
-                    assert!(self.host_start >= self.username_end + 2);
-                    assert_eq!(self.byte_at(self.host_start - 1), b'@');
+            if self.username_end != self.serialization.len() as u32 {
+                match self.byte_at(self.username_end) {
+                    b':' => {
+                        assert!(self.host_start >= self.username_end + 2);
+                        assert_eq!(self.byte_at(self.host_start - 1), b'@');
+                    }
+                    b'@' => assert!(self.host_start == self.username_end + 1),
+                    _ => assert_eq!(self.username_end, self.scheme_end + 3),
                 }
-                b'@' => assert!(self.host_start == self.username_end + 1),
-                _ => assert_eq!(self.username_end, self.scheme_end + 3),
             }
             assert!(self.host_start >= self.username_end);
             assert!(self.host_end >= self.host_start);
@@ -490,7 +492,10 @@ impl Url {
                     Some(port_str.parse::<u16>().expect("Couldn't parse port?"))
                 );
             }
-            assert_eq!(self.byte_at(self.path_start), b'/');
+            assert!(
+                self.path_start as usize == self.serialization.len()
+                    || matches!(self.byte_at(self.path_start), b'/' | b'#' | b'?')
+            );
         } else {
             // Anarchist URL (no authority)
             assert_eq!(self.username_end, self.scheme_end + 1);
@@ -501,11 +506,11 @@ impl Url {
             assert_eq!(self.path_start, self.scheme_end + 1);
         }
         if let Some(start) = self.query_start {
-            assert!(start > self.path_start);
+            assert!(start >= self.path_start);
             assert_eq!(self.byte_at(start), b'?');
         }
         if let Some(start) = self.fragment_start {
-            assert!(start > self.path_start);
+            assert!(start >= self.path_start);
             assert_eq!(self.byte_at(start), b'#');
         }
         if let (Some(query_start), Some(fragment_start)) = (self.query_start, self.fragment_start) {
@@ -745,7 +750,10 @@ impl Url {
     pub fn password(&self) -> Option<&str> {
         // This ':' is not the one marking a port number since a host can not be empty.
         // (Except for file: URLs, which do not have port numbers.)
-        if self.has_authority() && self.byte_at(self.username_end) == b':' {
+        if self.has_authority()
+            && self.username_end != self.serialization.len() as u32
+            && self.byte_at(self.username_end) == b':'
+        {
             debug_assert!(self.byte_at(self.host_start - 1) == b'@');
             Some(self.slice(self.username_end + 1..self.host_start - 1))
         } else {
@@ -1226,7 +1234,7 @@ impl Url {
         if let Some(input) = fragment {
             self.fragment_start = Some(to_u32(self.serialization.len()).unwrap());
             self.serialization.push('#');
-            self.mutate(|parser| parser.parse_fragment(parser::Input::new(input)))
+            self.mutate(|parser| parser.parse_fragment(parser::Input::no_trim(input)))
         } else {
             self.fragment_start = None
         }
@@ -1284,7 +1292,12 @@ impl Url {
             let scheme_type = SchemeType::from(self.scheme());
             let scheme_end = self.scheme_end;
             self.mutate(|parser| {
-                parser.parse_query(scheme_type, scheme_end, parser::Input::new(input))
+                let vfn = parser.violation_fn;
+                parser.parse_query(
+                    scheme_type,
+                    scheme_end,
+                    parser::Input::trim_tab_and_newlines(input, vfn),
+                )
             });
         }
 
@@ -1390,8 +1403,12 @@ impl Url {
                 }
                 parser.parse_cannot_be_a_base_path(parser::Input::new(path));
             } else {
+                let path_start = parser.serialization.len();
                 let mut has_host = true; // FIXME
                 parser.parse_path_start(scheme_type, &mut has_host, parser::Input::new(path));
+                if scheme_type.is_file() {
+                    parser::trim_path(&mut parser.serialization, path_start);
+                }
             }
         });
         self.restore_after_path(old_after_path_pos, &after_path);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2101,7 +2101,8 @@ impl Url {
         // If it is the scheme's default
         // We don't mind it silently failing
         // If there was no port in the first place
-        let _ = self.set_port(self.port());
+        let previous_port = self.port();
+        let _ = self.set_port(previous_port);
 
         Ok(())
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -690,7 +690,7 @@ impl Url {
     /// ```
     #[inline]
     pub fn cannot_be_a_base(&self) -> bool {
-        !self.slice(self.path_start..).starts_with('/')
+        !self.slice(self.scheme_end + 1..).starts_with('/')
     }
 
     /// Return the username for this URL (typically the empty string)
@@ -1642,10 +1642,25 @@ impl Url {
             if host == "" && SchemeType::from(self.scheme()).is_special() {
                 return Err(ParseError::EmptyHost);
             }
+            let mut host_substr = host;
+            // Otherwise, if c is U+003A (:) and the [] flag is unset, then
+            if !host.starts_with('[') || !host.ends_with(']') {
+                match host.find(':') {
+                    Some(0) => {
+                        // If buffer is the empty string, validation error, return failure.
+                        return Err(ParseError::InvalidDomainCharacter);
+                    }
+                    // Let host be the result of host parsing buffer
+                    Some(colon_index) => {
+                        host_substr = &host[..colon_index];
+                    }
+                    None => {}
+                }
+            }
             if SchemeType::from(self.scheme()).is_special() {
-                self.set_host_internal(Host::parse(host)?, None)
+                self.set_host_internal(Host::parse(host_substr)?, None);
             } else {
-                self.set_host_internal(Host::parse_opaque(host)?, None)
+                self.set_host_internal(Host::parse_opaque(host_substr)?, None);
             }
         } else if self.has_host() {
             if SchemeType::from(self.scheme()).is_special() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2508,6 +2508,7 @@ fn path_to_file_url_segments_windows(
     }
     let mut components = path.components();
 
+    let host_start = serialization.len() + 1;
     let host_end;
     let host_internal;
     match components.next() {
@@ -2534,14 +2535,23 @@ fn path_to_file_url_segments_windows(
         _ => return Err(()),
     }
 
+    let mut path_only_has_prefix = true;
     for component in components {
         if component == Component::RootDir {
             continue;
         }
+        path_only_has_prefix = false;
         // FIXME: somehow work with non-unicode?
         let component = component.as_os_str().to_str().ok_or(())?;
         serialization.push('/');
         serialization.extend(percent_encode(component.as_bytes(), PATH_SEGMENT));
+    }
+    // A windows drive letter must end with a slash.
+    if serialization.len() > host_start
+        && parser::is_windows_drive_letter(&serialization[host_start..])
+        && path_only_has_prefix
+    {
+        serialization.push('/');
     }
     Ok((host_end, host_internal))
 }
@@ -2566,6 +2576,14 @@ fn file_url_segments_to_pathbuf(
     for segment in segments {
         bytes.push(b'/');
         bytes.extend(percent_decode(segment.as_bytes()));
+    }
+    // A windows drive letter must end with a slash.
+    if bytes.len() > 2 {
+        if matches!(bytes[bytes.len() -2], b'a'..=b'z' | b'A'..=b'Z')
+            && matches!(bytes[bytes.len() - 1], b':' | b'|')
+        {
+            bytes.push(b'/');
+        }
     }
     let os_str = OsStr::from_bytes(&bytes);
     let path = PathBuf::from(os_str);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1663,8 +1663,13 @@ impl Url {
                 self.set_host_internal(Host::parse_opaque(host_substr)?, None);
             }
         } else if self.has_host() {
-            if SchemeType::from(self.scheme()).is_special() {
+            let scheme_type = SchemeType::from(self.scheme());
+            if scheme_type.is_special() {
                 return Err(ParseError::EmptyHost);
+            } else {
+                if self.serialization.len() == self.path_start as usize {
+                    self.serialization.push('/');
+                }
             }
             debug_assert!(self.byte_at(self.scheme_end) == b':');
             debug_assert!(self.byte_at(self.path_start) == b'/');
@@ -1967,14 +1972,28 @@ impl Url {
     ///
     /// # fn run() -> Result<(), ParseError> {
     /// let mut url = Url::parse("https://example.net")?;
-    /// let result = url.set_scheme("foo");
-    /// assert_eq!(url.as_str(), "foo://example.net/");
+    /// let result = url.set_scheme("http");
+    /// assert_eq!(url.as_str(), "http://example.net/");
     /// assert!(result.is_ok());
     /// # Ok(())
     /// # }
     /// # run().unwrap();
     /// ```
+    /// Change the URL’s scheme from `foo` to `bar`:
     ///
+    /// ```
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let mut url = Url::parse("foo://example.net")?;
+    /// let result = url.set_scheme("bar");
+    /// assert_eq!(url.as_str(), "bar://example.net");
+    /// assert!(result.is_ok());
+    /// # Ok(())
+    /// # }
+    /// # run().unwrap();
+    /// ```
     ///
     /// Cannot change URL’s scheme from `https` to `foõ`:
     ///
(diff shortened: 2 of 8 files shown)


o0Ignition0o:
The goal would be to merge this branch into the tests one, before merging both I suppose.
I can of course also make a pull request directly to master if need be

o0Ignition0o:
Rebased into master, and fixed the merge issues.
I think a lot of us are on PTO right now, please let me know when you're back :)

nox:
> The goal would be to merge this branch into the tests one, before merging both I suppose.
> I can of course also make a pull request directly to master if need be

I don't understand this part.

o0Ignition0o:
> I don't understand this part.
My bad, I finally changed the target branch to master. the sentence is not relevant anymore

nox:
Could you maybe squash some of the commits together to make the tryouts commits and whatnot disappear, to help with review?

o0Ignition0o:
Sure, would you rather have one huge commit, or several smaller one ?
I know the PR is pretty big, it's probably a PITA to review, I'm willing to do anything I can to ease the review process, I just don't know how to make it easier to read.

nox:
Small ones are ok as long as they are self-contained, for example the commits for setter fixes are ok, but the ones with commented-out assertions aren't.

o0Ignition0o:
I've narrowed it down to 5-6 commits, not sure I can do more though :/

nox:
That's great already! Thanks a lot for that.

src/host.rs
```diff
@@ -24,9 +24,13 @@ pub(crate) enum HostInternal {
...
 }
 
-impl<S> From<Host<S>> for HostInternal {
+impl<S> From<Host<S>> for HostInternal
+where
+    S: ToString,
+{
     fn from(host: Host<S>) -> HostInternal {
         match host {
+            Host::Domain(ref s) if s.to_string().is_empty() => HostInternal::None,
```
nox: What exactly is this doing? This allocates a new string every time. Do we really need the generics here? Couldn't it be `From<Host<String>>` or `From<Host<&'_ str>>`?

src/parser.rs
```diff
@@ -79,6 +79,7 @@ macro_rules! simple_enum_error {
 
 simple_enum_error! {
     EmptyHost => "empty host",
+    InvalidAuthority => "invalid authority",
```
nox: Breaking change. :/

src/quirks.rs
```diff
@@ -110,14 +112,23 @@ pub fn set_host(url: &mut Url, new_host: &str) -> Result<(), ()> {
     let opt_port;
     {
         let scheme = url.scheme();
-        let result = Parser::parse_host(Input::new(new_host), SchemeType::from(scheme));
+        let scheme_type = SchemeType::from(scheme);
+        let result = if scheme_type == SchemeType::File {
+            Parser::get_file_host(input)
+        } else {
+            Parser::parse_host(input, scheme_type)
```
nox: Why not do what `get_file_host` does directly in `parse_host`, checking in that latter function directly whether `scheme_type` is `SchemeType::File`?

src/quirks.rs
```diff
@@ -154,8 +165,30 @@ pub fn set_hostname(url: &mut Url, new_hostname: &str) -> Result<(), ()> {
     if url.cannot_be_a_base() {
         return Err(());
     }
-    let result = Parser::parse_host(Input::new(new_hostname), SchemeType::from(url.scheme()));
+    // Host parsing rules are strict,
+    // We don't want to trim the input
```
nox: Where is the spec for all of this? Is this what the parser spec does when given the hostname state override or whatever?

src/host.rs
```diff
@@ -24,9 +24,52 @@ pub(crate) enum HostInternal {
     Ipv6(Ipv6Addr),
 }
 
-impl<S> From<Host<S>> for HostInternal {
+#[cfg(feature = "serde")]
```
nox: Why is this back?

src/lib.rs
```diff
@@ -490,7 +492,12 @@ impl Url {
                     Some(port_str.parse::<u16>().expect("Couldn't parse port?"))
                 );
             }
-            assert_eq!(self.byte_at(self.path_start), b'/');
+            assert!(
+                self.path_start as usize == self.serialization.len()
+                    || self.byte_at(self.path_start) == b'/'
+                    || self.byte_at(self.path_start) == b'#'
+                    || self.byte_at(self.path_start) == b'?'
+            );
```
nox: Use `matches!` instead of multiple `==`.

src/lib.rs
```diff
@@ -1390,8 +1404,12 @@ impl Url {
                 }
                 parser.parse_cannot_be_a_base_path(parser::Input::new(path));
             } else {
+                let path_start = parser.serialization.len();
                 let mut has_host = true; // FIXME
                 parser.parse_path_start(scheme_type, &mut has_host, parser::Input::new(path));
+                if scheme_type.is_file() {
+                    parser::trim_path(&mut parser.serialization, path_start);
+                }
```
nox: Where is the spec for this?

src/parser.rs
```diff
@@ -201,6 +201,19 @@ impl<'i> Input<'i> {
...
     }
 
+    pub fn no_trim(input: &'i str) -> Self {
+        Input {
+            chars: input.chars(),
+        }
+    }
+
+    pub fn trim_tab_and_newlines(original_input: &'i str) -> Self {
+        let input = original_input.trim_matches(ascii_tab_or_new_line);
```
nox: Mmmh, when trimming those, are they no longer syntax violations?

src/parser.rs
```diff
@@ -1109,35 +1186,59 @@ impl<'a> Parser<'a> {
...
-                        && is_windows_drive_letter(&self.serialization[path_start + 1..])
-                    {
-                        if self.serialization.ends_with('|') {
-                            self.serialization.pop();
-                            self.serialization.push(':');
+                    // If url’s scheme is "file", url’s path is empty, and buffer is a Windows drive letter, then
+                    if scheme_type.is_file() && is_windows_drive_letter(segment_before_slash) {
+                        //  Replace the second code point in buffer with U+003A (:).
+                        if let Some(c) = segment_before_slash.chars().nth(0) {
+                            let mut drive_letter = "".to_string();
```
nox: Don't allocate a string for that, just push directly to `self.serialization`.

src/parser.rs
```diff
@@ -1147,16 +1248,48 @@ impl<'a> Parser<'a> {
...
+                && !path_starts_with_windows_drive_letter(&serialization[segment_before_start..])
+        } else {
+            false
+        }
+    }
+
+    /// https://url.spec.whatwg.org/#shorten-a-urls-path
+    fn shorten_path(&mut self, scheme_type: SchemeType, path_start: usize) {
+        // If path is empty, then return.
+        if self.serialization.len() <= path_start {
```
nox: I don't understand how the comment relates to the condition.

src/parser.rs
```diff
@@ -1147,16 +1248,48 @@ impl<'a> Parser<'a> {
...
+            return;
+        }
+        // If url’s scheme is "file", path’s size is 1, and path[0] is a normalized Windows drive letter, then return.
+        let segments: Vec<&str> = self.serialization[path_start..]
```
nox: Seems bad to me that we allocate a vector all the time here.

src/host.rs
o0Ignition0o: We actually don't need the generics here.
The main goal of this change is to make sure HostInternal::None is returned when an empty domain is given.

nox: How can it be empty?

o0Ignition0o: Here's a list of the failing tests that occur if I don't add this check: 

```
failures:
    "file://hi/x".host = "" 
    "file://hi/x".hostname = "" 
    "file://y/".host = "loc%41lhost" 
    "file://y/".hostname = "loc%41lhost" 
```
Maybe the fix should occur somewhere else, if after the host parsing ?

```
........................................................................................................................................................................thread '"file://hi/x".hostname = "" ' panicked at 'called `Result::unwrap()` on an `Err` value: "!( !host_str.is_empty() ) for URL \"file:///x\""', src/libcore/result.rs.:999:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
..thread '"file://hi/x".host = "" ' panicked at 'called `Result::unwrap()` on an `Err` value: "!( !host_str.is_empty() ) for URL \"file:///x\""', src/libcore/result.rs:999:5
.......F.F...thread '"file://y/".host = "loc%41lhost" ' panicked at 'called `Result::unwrap()` on an `Err` value: "!( !host_str.is_empty() ) for URL \"file:///\""thread '', "file://y/".hostname = "loc%41lhost" src/libcore/result.rs' panicked at ':called `Result::unwrap()` on an `Err` value: "!( !host_str.is_empty() ) for URL \"file:///\""999', :src/libcore/result.rs5:
999:5
..F.F............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

src/parser.rs
```diff
@@ -217,7 +218,7 @@ impl<'i> Input<'i> {
     pub fn with_log(original_input: &'i str, vfn: Option<&dyn Fn(SyntaxViolation)>) -> Self {
         let input = original_input.trim_matches(c0_control_or_space);
         if let Some(vfn) = vfn {
-            if input.len() < original_input.len() {
+            if input.len() != original_input.len() {
```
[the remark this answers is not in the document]
o0Ignition0o: It is not, I probably got confused while working with the trimming functions. reverting this change.

src/parser.rs
```diff
@@ -1018,10 +1048,41 @@ impl<'a> Parser<'a> {
         Ok((host, input))
     }
 
-    pub(crate) fn parse_file_host<'i>(
+    pub fn get_file_host<'i>(input: Input<'i>) -> ParseResult<(Host<String>, Input)> {
```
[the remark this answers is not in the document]
o0Ignition0o: It indeed doesn't have to. 
This comment also helped me remember I'm probably the reason why [the host can be empty](https://github.com/servo/rust-url/pull/537#discussion_r315246952).

[Host::parse](https://github.com/servo/rust-url/blob/master/src/host.rs#L75) is not aware of the Scheme, and cannot "clear" localhost if the SchemeType is file.
Now the problem is these functions don't deal with HostInternal but with Host variants, so I'm not able to leverage HostInternal::None. 

However you being confused means I should refactor it. I'm not sure how / which way would be the most elegant though.

src/lib.rs
o0Ignition0o: I had to dig quite a bit to figure out how to pass the path tests.
All I could find was [this discussion](https://github.com/whatwg/url/issues/232).

Here's a list of test that fail when the path isn't trimmed:

```
running 713 tests
..........................................................................................................................................................thread '"file:///unicorn".pathname = "//\\/" File URLs and (back)slashes' panicked at 'assertion failed: `(left == right)`
  left: `"file://////"`,
 right: `"file:///"`.', tests/data.rs:192:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
thread '"file:///unicorn".pathname = "//monkey/..//" File URLs and (back)slashes' panicked at 'assertion failed: `(left == right)`
  left: `"file://///"`,
 right: `"file:///"`.', tests/data.rs:192:5
..........FF.......thread '"file://monkey/".pathname = "\\\\" File URLs and (back)slashes' panicked at 'assertion failed: `(left == right)`
  left: `"file://monkey//"`,
 right: `"file://monkey/"`', .tests/data.rs:192:5
.F.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
failures:

failures:
    "file:///unicorn".pathname = "//\\/" File URLs and (back)slashes
    "file:///unicorn".pathname = "//monkey/..//" File URLs and (back)slashes
    "file://monkey/".pathname = "\\\\" File URLs and (back)slashes
```

src/parser.rs
o0Ignition0o: Oh my bad, they are !

src/parser.rs
o0Ignition0o: There are now three possibilities: 
- The scheme is special and the path is empty: "https://servo.org" ( path start is right after the buffer, self.serialization.len() )
- The scheme is not special and the path is empty: "tests://servo.org/" ( path start is the last slash position, self.serialization.len() -1 )
- The path is not empty: "https://servo.org/screenshot.png"  ( path start is the last slash position, which is much less than self.serialization.len() )

A URLs path is currently empty if ` self.serialization.len() < path_start `.
A URLs path only contains a root slash if ` self.serialization.len() == path_start `.

o0Ignition0o: If the comment makes sense to you here, it probably belongs to the code.
Let me know and I'll add it.
If not I need to figure out a more explicit way to convey that.

src/parser.rs
o0Ignition0o: It is, I'll try to figure something else out :/

src/host.rs
o0Ignition0o: https://github.com/servo/rust-url/commit/171e65e5c982480fc81b428d44228f153435d0a8 I didnt notice that, sorry ^^

src/quirks.rs
o0Ignition0o: Here are the two tests that fail when a trimmed input is passed : 

```
running 713 tests
......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................thread '"sc://x/".host = " " ' panicked at 'assertion failed: `(left == right)`
  left: `"sc:///"`,
 right: `"sc://x/"`', tests/data.rs:190:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
.........thread '"sc://x/".host = "\u{0}" Non-special scheme' panicked at 'assertion failed: `(left == right)`
  left: `"sc:///"`,
 right: `"sc://x/"`', tests/data.rs:190:5
...........F..F...........................................................
failures:

failures:
    "sc://x/".host = " " 
    "sc://x/".host = "\u{0}" Non-special scheme

test result: FAILED. 711 passed; 2 failed; 0 ignored; 0 measured
```
Analyzing one of them Is what lead me to change it : 
```
        {
            "comment": "Non-special scheme",
            "href": "sc://x/",
            "new_value": "\u0000",
            "expected": {
                "href": "sc://x/",
                "host": "x",
                "hostname": "x"
            }
        },
```
I think I read the correct section, but misimplemented it: https://url.spec.whatwg.org/#hostname-state

I think the relevant rule here is this one:
> Otherwise, if state override is given, buffer is the empty string, and either url includes credentials or url’s port is non-null, validation error, return.

The problem is I don't think there's a way for me to figure out state_override is given :/

src/lib.rs
```diff
@@ -456,13 +456,15 @@ impl Url {
 
         if self.slice(self.scheme_end + 1..).starts_with("//") {
             // URL with authority
-            match self.byte_at(self.username_end) {
-                b':' => {
-                    assert!(self.host_start >= self.username_end + 2);
-                    assert_eq!(self.byte_at(self.host_start - 1), b'@');
+            if self.username_end < self.serialization.len() as u32 {
```
nox: Why would `self.username_end` be greater than `self.serialization`?

src/host.rs
nox: Where is the `<From<Host<S>>>::from` call done in those?

src/lib.rs
```diff
@@ -745,7 +750,10 @@ impl Url {
     pub fn password(&self) -> Option<&str> {
         // This ':' is not the one marking a port number since a host can not be empty.
         // (Except for file: URLs, which do not have port numbers.)
-        if self.has_authority() && self.byte_at(self.username_end) == b':' {
+        if self.has_authority()
+            && self.username_end < self.serialization.len() as u32
```
nox: Why would `self.username_end` be greater than `self.serialization`?

src/lib.rs
```diff
@@ -1284,7 +1292,12 @@ impl Url {
             let scheme_type = SchemeType::from(self.scheme());
             let scheme_end = self.scheme_end;
             self.mutate(|parser| {
-                parser.parse_query(scheme_type, scheme_end, parser::Input::new(input))
+                let vfn = parser.violation_fn;
+                parser.parse_query(
+                    scheme_type,
+                    scheme_end,
+                    parser::Input::trim_tab_and_newlines(input, vfn),
```
nox: Where is the spec for the trimming when setting a query?

src/lib.rs
nox: That's not a spec issue, please link to a spec commit.

If I understand correctly, you trim the path only after it was built and not trimmed? Couldn't it just not be pushing empty segments on repeated slashes in the first place?

src/parser.rs
nox: I don't understand anymore what this code is supposed to be doing anymore. Why are we not trimming `c0_control_or_space` anymore? This method looks oddly like `with_log` now.

src/parser.rs
```diff
@@ -283,6 +307,17 @@ impl Pattern for char {
     }
 }
 
+impl Pattern for String {
```
nox: Isn't `Pattern for &'_ str` enough? Why do we need this impl?

src/parser.rs
```diff
@@ -440,7 +475,11 @@ impl<'a> Parser<'a> {
                         .collect::<String>()
                         != "//"
                 });
-                self.after_double_slash(remaining, scheme_type, scheme_end)
+                if let Some(after_prefix) = input.split_prefix("/".repeat(slashes_count as usize)) {
```
nox: That allocates a new `String` for nothing, please change this.

src/parser.rs
```diff
@@ -556,16 +597,28 @@ impl<'a> Parser<'a> {
...
+                    input_after_first_char
+                };
+
+                let remaining =
+                    self.parse_path(SchemeType::File, &mut false, host_end, parse_path_input);
+
+                let host_start = host_start as u32;
+
+                // TODO: Handle authority
+                trim_path(&mut self.serialization, host_end);
```
nox: Why do we need to trim here?

src/parser.rs
```diff
@@ -556,16 +597,28 @@ impl<'a> Parser<'a> {
...
+                } else {
+                    input_after_first_char
+                };
+
+                let remaining =
+                    self.parse_path(SchemeType::File, &mut false, host_end, parse_path_input);
+
+                let host_start = host_start as u32;
+
+                // TODO: Handle authority
```
nox: What is this todo?

src/parser.rs
```diff
@@ -1040,21 +1112,36 @@ impl<'a> Parser<'a> {
...
-        match input.split_first() {
-            (Some('/'), remaining) => input = remaining,
-            (Some('\\'), remaining) => {
-                if scheme_type.is_special() {
+        let path_start = self.serialization.len();
+        let (maybe_c, remaining) = input.split_first();
+        // If url is special, then:
+        if scheme_type.is_special() {
+            if let Some(c) = maybe_c {
+                if c == '\\' {
```
nox: You can just do `if maybe_c == Some('\\')`.

src/parser.rs
```diff
@@ -1109,35 +1197,57 @@ impl<'a> Parser<'a> {
                     }
                 }
             }
-            match &self.serialization[segment_start..] {
+
+            let segment_before_slash = if ends_with_slash {
+                &self.serialization[segment_start..self.serialization.len() - 1]
+            } else {
+                &self.serialization[segment_start..self.serialization.len()]
+            };
```
nox: Why are we pushing the slash earlier now if we need to ignore it later?

src/parser.rs
```diff
@@ -1109,35 +1197,57 @@ impl<'a> Parser<'a> {
...
+                    // and then if neither c is U+002F (/), nor url is special and c is U+005C (\), append the empty string to url’s path.
+                    if ends_with_slash && !self.serialization.ends_with("/") {
+                        self.serialization.push('/');
                     }
                 }
+                // Otherwise, if buffer is a single-dot path segment and if neither c is U+002F (/),
+                // nor url is special and c is U+005C (\), append the empty string to url’s path.
                 "." | "%2e" | "%2E" => {
                     self.serialization.truncate(segment_start);
+                    if ends_with_slash && !self.serialization.ends_with("/") {
```
nox: Why is `ends_with_slash` true if `self.serialization.ends_with("/")` is false?

src/parser.rs
nox: So when is `path_start` ever greater than `self.serialization.len()`?

src/parser.rs
```diff
@@ -1318,6 +1460,19 @@ impl<'a> Parser<'a> {
     }
 }
 
+// Trim path start forward slashes when no authority is present
+// https://github.com/whatwg/url/issues/232
+pub fn trim_path(serialization: &mut String, path_start: usize) {
+    let path = serialization.split_off(path_start);
+    if path.starts_with("/") {
+        let mut trimmed_path = "/".to_string();
```
nox: Unneeded string allocation.

src/parser.rs
```diff
@@ -1318,6 +1460,19 @@ impl<'a> Parser<'a> {
     }
 }
 
+// Trim path start forward slashes when no authority is present
+// https://github.com/whatwg/url/issues/232
```
nox: I'm pretty sure I have some code somewhere that fixes the same tests as the ones in that WHATWG issue with less shenanigans, I'll try to find it again.

src/parser.rs
```diff
@@ -1380,12 +1541,34 @@ fn is_windows_drive_letter(segment: &str) -> bool {
...
+fn path_starts_with_windows_drive_letter(s: &str) -> bool {
+    s.len() > 3
+        && matches!(s.as_bytes()[0], b'/' | b'\\' | b'?' | b'#')
+        && starts_with_windows_drive_letter(&s[1..])
+}
+
 fn starts_with_windows_drive_letter(s: &str) -> bool {
-    ascii_alpha(s.as_bytes()[0] as char) && matches!(s.as_bytes()[1], b':' | b'|')
+    ascii_alpha(s.as_bytes()[0] as char)
+        && matches!(s.as_bytes()[1], b':' | b'|')
```
nox: This can panic if `s` is too short.

src/parser.rs
```diff
@@ -1380,12 +1541,34 @@ fn is_windows_drive_letter(segment: &str) -> bool {
...
+        && (s.len() == 2 || matches!(s.as_bytes()[2], b'/' | b'\\' | b'?' | b'#'))
 }
 
+/// https://url.spec.whatwg.org/#start-with-a-windows-drive-letter
 fn starts_with_windows_drive_letter_segment(input: &Input) -> bool {
```
nox: This method should probably be calling `starts_with_windows_drive_letter`.

src/parser.rs
o0Ignition0o: As defined [in the basic url parser spec](https://url.spec.whatwg.org/#concept-basic-url-parser), removing C0 control or space from input only occurs when If url is not given.

Removing tabs and newlines occurs either way though.

This means we need to use different `parser::Input` s.

I think It would be more explicit if with_log / trim_tab_and_newlines mentionned that, ie 
parser::Input::with_base_url / parser::Input::no_base_url or something, or having a factory that would decide on the best parser to build depending on having a base_url.

Setters would use the former, and the default parser would use the latter.
What do you think ?

src/parser.rs
o0Ignition0o: I think you're correct, `if self.serialization.len() == path_start` is enough.

src/parser.rs
o0Ignition0o: That was my intention at first, but I couldnt figure out a way to go from input to a &str, and I decided to back off.
It also means starts_with_windows_drive_letter is only called once, so I'd rather remove it :/

src/parser.rs
o0Ignition0o: Yeah let's pretend it never happened. :see_no_evil:

src/parser.rs
o0Ignition0o: Yeah It's because of the useless shenanigans I pulled with the after_prefix. Just removed it.

src/parser.rs
o0Ignition0o: The RFC https://www.ietf.org/rfc/rfc3986.txt claims 
> When authority is present, the path must
   either be empty or begin with a slash ("/") character.  When
   authority is not present, the path cannot begin with two slash
   characters ("//").  These restrictions result in five different ABNF
   rules for a path (Section 3.3), only one of which will match any
   given URI reference.

I think that's what I marked as todo.
I think I can remove it since the trimming only occurs in the parse_file paths.

The trimming function will probably vanish anyway, because we can probably just not add slashes in the first place.

src/parser.rs
o0Ignition0o: There's probably a much wiser way to do it, which I still need to figure out, but I'm making sure I push the root slash if there's one. I must admit it creates much more convoluted code afterwards. I ll try to refactor it.

src/lib.rs
o0Ignition0o: I have just found https://bugzilla.mozilla.org/show_bug.cgi?id=1351603, but I still cannot find any spec commit :/

valenting: I believe this was the spec bug for the tests: https://github.com/whatwg/url/pull/278

o0Ignition0o:
```
If url’s scheme is "file" and c is the EOF code point, U+003F (?), or U+0023 (#), then while url’s path’s size is greater than 1 and url’s path[0] is the empty string, validation error, remove the first item from url’s path.
``` 
is probably the culprit ! thanks for the hint @valenting :D 

There's probably a way to avoid pushing the empty segments, which I ll work on now that I see the spec :)

src/parser.rs
```diff
@@ -1142,54 +1227,118 @@ impl<'a> Parser<'a> {
...
             if !ends_with_slash {
                 break;
             }
         }
+        if scheme_type.is_file() {
+            // while url’s path’s size is greater than 1
+            // and url’s path[0] is the empty string,
+            // validation error, remove the first item from url’s path.
+            //TODO: log violation
+            trim_path(&mut self.serialization, path_start);
```
o0Ignition0o: @nox @valenting as stated per the specification, we're supposed to log a violation here.

On one hand there doesn't seem to be any relevant enum variant [available yet](https://github.com/servo/rust-url/blob/master/src/parser.rs#L136)
OTOH I don't know if we want to introduce a breaking change.

What do you think ?

src/host.rs
o0Ignition0o: I m not really sure this is the question you're asking me, but the conversion I'm trying to work on is [host.into()](https://github.com/servo/rust-url/blob/master/src/lib.rs#L1677)

src/parser.rs
o0Ignition0o: Found it ! I'm pushing it as a fixup, will do a rebase once we're both safisfied with most of the fixes

src/parser.rs
o0Ignition0o: There was indeed no need for an if else ^^'

o0Ignition0o:
Hey everyone, is there anything I can do to ease the reviewing process ?
Could anyone let me know how to go on with this pr, and where to focus first to help land it someday ? 
Thanks again for your patience and your time <3

o0Ignition0o:
Ok It seems like the build failed because smallvec updated their mem::uninitialized() code in https://github.com/servo/rust-smallvec/pull/162, which fixed https://github.com/servo/rust-smallvec/issues/126.

Given the nature of the commit I really think it was actually a good idea, but it blows up our rust 1.33 stable support (std::mem::MaybeUninit was introduced in Rust 1.36)

I'm not knowledgeable enough to know if we could pin smallvec to an earlier commit (thus keeping possible unsoundness but maintaining our rust 1.33 stable support), or switch our minimum supported version to 1.36 (this merge would probably introduce major changes to rust-url/break some behaviour and backward compatibility anyway)

@valenting @nox @SimonSapin you're probably more able to make an informed decision than I am, any thought ?

o0Ignition0o:
| I think you can also have an empty username but a non empty password. Do you mind adding a test for that if we don't have one already?

Given the username/password separator is a `:`, If I try to parse a URL with an empty username and a password it will error out trying to parse the password as a port.

I did however try this sequence: 

```

#[test]
fn test_empty_username_and_password() {
    use url::quirks;
    let mut base: Url = "moz://foo:bar@servo/baz".parse().unwrap();
    base.set_username("");
    assert_eq!(base.as_str(), "moz://:bar@servo/baz");
    base.set_host(None);
    assert_eq!(base.as_str(), "moz:/baz");
}
```

And the test did pass, although moz:/baz is definitely not a correct URL.

I'll try to investigate and come up with something asap

o0Ignition0o:
I have added the test to the suite, and bumped rustc to version 1.36.0 in travis (in case we go that way)
I can of course rollback the change if we want to keep rust 1.33 onwards.

valenting:
> Edit: as shown [here](https://github.com/servo/rust-url/blob/master/src/lib.rs#L1570) it seems to be the accepted behaviour...

Actually, I think that is wrong. IMO it should be `moz:///baz`

This seems to agree with me:
https://jsdom.github.io/whatwg-url/
Note: you can open the console on that page and run 
``` js
url = new whatwgURL.URL("moz://foo:bar@servo/baz");
```
to test the setters. 

But in any case, given the comment, I think we can leave it like this for now, and open another bug for that.

o0Ignition0o:
> This seems to agree with me: jsdom.github.io/whatwg-url

Wow I didn't know about this tool, pretty cool!

> Actually, I meant this should fail if we have a non empty password, like it does for non empty usernames.

Oh you mean If we try to set_hostname on a URL that has an empty username _and_ a non empty password?
Ok I'll add a test for it as well ! :)

valenting:
> Oh you mean If we try to set_hostname on a URL that has an empty username _and_ a non empty password?

Right, so if there's either a username or password we should fail setting the hostname.
`moz://foo@servo/baz`.set_hostname("") should fail
`moz://:pass@servo/baz`.set_hostname("") should fail
`moz://servo/baz`.set_hostname("") should work and for now it's OK that it parses to `moz:/baz`

o0Ignition0o:
Thanks for the thorough response and the examples.
I'm going to write them as is in the unit tests, and make sure I use results in the process :)

SimonSapin:
>  I can of course rollback the change if we want to keep rust 1.33 onwards.

What change do you mean to roll back? As far as I can tell breakage on 1.33 happens because of https://github.com/unicode-rs/unicode-normalization/issues/50 which is outside of our control.

I’ve reopened https://github.com/servo/rust-url/issues/566 to discuss this specifically, since this thread is already long.

o0Ignition0o:
As discussed [here](https://github.com/servo/rust-url/issues/566#issuecomment-563301401) I will put back the 1.33 version in .travis.yml so we can open an other PR (and figure out what we actually want to do).

o0Ignition0o:
> > Oh you mean If we try to set_hostname on a URL that has an empty username _and_ a non empty password?
> 
> Right, so if there's either a username or password we should fail setting the hostname.
> `moz://foo@servo/baz`.set_hostname("") should fail
> `moz://:pass@servo/baz`.set_hostname("") should fail
> `moz://servo/baz`.set_hostname("") should work and for now it's OK that it parses to `moz:/baz`

It seems like I got a bit confused between Url::set_host(&mut self, host: Option<&str>) and quirks::set_hostname(url: &mut Url, new_hostname: &str).

It turns out [set_hostname behaves as expected](https://github.com/servo/rust-url/pull/537/files#diff-43b0229bccbdb5d0f2ee7e0e56735654R52) :)
I have added quirks tests as defined in your comment to make sure it stays that way :)

o0Ignition0o:
@valenting Just rebased ! 🤞

PR #550, opened 2019-09-28
Make EncodingOverride Send + Sync so that form_urlencoded::Serializer is as well

udoprog:
Without this, `form_urlencoded::Serializer` _usually_ can't cross `await` points in async code, making it harder to use.

Unfortunately I think this is a major change, so please advice.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/550)

### DIFF
--- a/src/query_encoding.rs
+++ b/src/query_encoding.rs
@@ -8,7 +8,7 @@
 
 use std::borrow::Cow;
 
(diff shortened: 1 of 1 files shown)


nox:
This is indeed a breaking change and we just released `url` 2.

djc:
Changing the title to note that these changes are semver-incompatible (in my understanding). Tracking all proposed semver-incompatible changes in #627.

PR #555, opened 2019-10-15
WIP: Fix hostname parsing, limit to ASCII / IDNA

### LINKED ISSUE #554: I think host parsing is incorrect, might not follow spec, use ASCII / IDNA only
A few days ago I discovered that some string are parsed successfully as `Url`, even though I'm quite certain those are invalid, such as `"http://\""`. (see https://github.com/seanmonstar/reqwest/issues/668 via #552)

### The problem
I located the problem to be with hostname parsing. Any hostname characters are accepted at this moment (except for a small blacklist). I believe hostnames must be limited to ASCII or IDNA encodable characters, so that'd be an error.

To fix this, I changed the following line to use `domain_to_ascii_strict` instead.
This strict variant fails on any character that is not encodable as ASCII/IDNA:

https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/src/host.rs#L83

Changing this does solve the issue I've linked before: https://github.com/seanmonstar/reqwest/issues/668

### The catch
But there's a catch. This change makes a few included unit tests fail. I'm
wondering whether these unit tests are conform the [URL specification][url-spec] that is explicitly linked in the README of this crate. If these are not, I'd argue they should be fixed or removed.

#### Failing tests
* First, the 'Domains with empty labels' fails from the test data set:
  https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/tests/urltestdata.json#L3802-L3847
  This is probably related to the following:

* Second, the leading dots (host) unit test fails:
  https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/tests/unit.rs#L421-L429  
  Issue #166 and #337 are related.

* Third, the host unit test fails:
  https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/tests/unit.rs#L203-L237  
  Only line [219](https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/tests/unit.rs#L219) and [234](https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/tests/unit.rs#L234) fail. Has to do with dots again.

I don't know whether these tests are conform the URL specification. Sadly, I don't have enough time right now to study this.

### Questions
Now I might be totally wrong on this, that's why I'm asking the following to `url` maintainers. Hopefully you can help me out.

- Is my assumption correct, is `"http://\""` incorrect?
- Should `Url::parse(...)` fail, for input not conform the [URL specification][url-spec]?
- Should hostnames be limited to ASCII and IDNA characters?
- Are the linked unit tests faulty, based on the above?

Based on the questions above, what would be the proper steps to take, if there are any?
Maybe a special case should be implemented for handling `.` and `..` domains if a base is known (this would partially fix unit tests)?

I've prepared a branch/PR with this change. It doesn't contain much changes because the questions need answering first. It currently obviously fails due to these unit tests. Here it is: #555

[url-spec]: https://url.spec.whatwg.org/

timvisee:
Fixes #554

This is a work in progress, please refer to the linked issue for more information.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -80,7 +80,7 @@ impl Host<String> {
             return parse_ipv6addr(&input[1..input.len() - 1]).map(Host::Ipv6);
         }
         let domain = percent_decode(input.as_bytes()).decode_utf8_lossy();
-        let domain = idna::domain_to_ascii(&domain)?;
+        let domain = idna::domain_to_ascii_strict(&domain)?;
         if domain
             .find(|c| {
                 matches!(


djc:
There were some changes in the URL Standard and accompanying tests that I just merged into master. These might interact with the issues you resolved in #554, in particular the changes in https://github.com/servo/rust-url/commit/6b05499d3f42f3522fb549838b80148a4cbdce23. Since the issue you filed was closed, I think it makes sense to close this too at this point, but please let me know if you still think there's an issue!

PR #560, opened 2019-11-03
Unsafe audit

JoshMcguigan:
Hello,

I just did a quick unsafe audit, directed here from https://github.com/rust-secure-code/safety-dance/issues/51.

I was able to remove the unsafe block from `decode_utf8_lossy`. Also, I removed the duplicate definition of `decode_utf8_lossy` by allowing `src/form_urlencoded.rs` to import it from `percent_encoding/lib.rs` rather than `src/query_encoding.rs`. 

All other unsafe blocks look appropriate to me, as they ensure we don't waste CPU cycles checking the validity of bytes that we already have confirmed in some way as safe to convert to UTF-8, but I added comments to justify why they are safe.

### DIFF
--- a/idna/src/punycode.rs
+++ b/idna/src/punycode.rs
@@ -143,11 +143,10 @@ pub fn encode_str(input: &str) -> Option<String> {
 /// 63 encoded bytes, the DNS limit on domain name labels.
 pub fn encode(input: &[char]) -> Option<String> {
     // Handle "basic" (ASCII) code points. They are encoded as-is.
-    let output_bytes = input
+    let mut output : String = input
         .iter()
-        .filter_map(|&c| if c.is_ascii() { Some(c as u8) } else { None })
+        .filter(|&c| c.is_ascii())
         .collect();
-    let mut output = unsafe { String::from_utf8_unchecked(output_bytes) };
     let basic_length = output.len() as u32;
     if basic_length > 0 {
         output.push_str("-")
--- a/form_urlencoded/src/lib.rs
+++ b/form_urlencoded/src/lib.rs
@@ -150,6 +150,10 @@ impl<'a> Iterator for ByteSerialize<'a> {
                 None => (self.bytes, &[][..]),
             };
             self.bytes = remaining;
+            // This unsafe is appropriate because we have already checked these
+            // bytes in byte_serialized_unchanged, which checks for a subset
+            // of UTF-8. So we know these bytes are valid UTF-8, and doing
+            // another UTF-8 check would be wasteful.
             Some(unsafe { str::from_utf8_unchecked(unchanged_slice) })
         } else {
             None
--- a/form_urlencoded/src/query_encoding.rs
+++ b/form_urlencoded/src/query_encoding.rs
@@ -18,18 +18,34 @@ pub(crate) fn encode<'a>(encoding_override: EncodingOverride, input: &'a str) ->
 }
 
 pub(crate) fn decode_utf8_lossy(input: Cow<[u8]>) -> Cow<str> {
+    // Note: This function is duplicated in `percent_encoding/lib.rs`.
     match input {
         Cow::Borrowed(bytes) => String::from_utf8_lossy(bytes),
         Cow::Owned(bytes) => {
-            let raw_utf8: *const [u8];
             match String::from_utf8_lossy(&bytes) {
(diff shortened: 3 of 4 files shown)


SimonSapin:
Thanks for doing this audit!

It looks like the new `decode_utf8_lossy` spends more CPU cycles: if the first UTF-8 error is near the end of a large string, almost all of the string will be scanned twice. Once in `String::from_utf8` and again from the start in `String::from_utf8_lossy`.

Also, `percent-encoding` is a crate that can be used separately, and that has a public API governed by SemVer. `decode_utf8_lossy` is short enough that I’d prefer duplicating it rather than making it part of the stable API.

idna/src/punycode.rs
```diff
@@ -147,6 +147,10 @@ pub fn encode(input: &[char]) -> Option<String> {
         .iter()
         .filter_map(|&c| if c.is_ascii() { Some(c as u8) } else { None })
         .collect();
+    // This is an appropriate use of unsafe because the bytes are confirmed
+    // as a subset of UTF-8 above. This logic removes non-ascii bytes rather
+    // than replacing them with some placeholder character, so it would not
+    // be correct to use from_utf8_lossy here.
     let mut output = unsafe { String::from_utf8_unchecked(output_bytes) };
```
Shnatsel: This `unsafe` block is not actually required - you can collect chars directly to String instead of creating an intermediate vector above:

```rust
    let output: String = input
        .iter()
        .filter(|&c| c.is_ascii())
        .collect();
```

It also results in a considerably shorter assembly than the current unsafe version.

JoshMcguigan: This is a great suggestion, thanks for pointing it out. I've updated it as shown.

JoshMcguigan:
@SimonSapin Thanks for the review.

> It looks like the new decode_utf8_lossy spends more CPU cycles: if the first UTF-8 error is near the end of a large string, almost all of the string will be scanned twice. Once in String::from_utf8 and again from the start in String::from_utf8_lossy.

Good catch, I was thinking only about allocations. I've rolled this back, but added some small comments/tweaks that I think improve readability. Let me know what you think.

> Also, percent-encoding is a crate that can be used separately, and that has a public API governed by SemVer. decode_utf8_lossy is short enough that I’d prefer duplicating it rather than making it part of the stable API.

This makes sense. For now, I've rolled back the de-duplication and added a note so if those functions are modified in the future they can be changed in both places. It seems the ideal case would be for this functionality to live in the standard library. I'm thinking the function signature of `String::from_utf8_lossy` could look like:

```rust
impl String {
    fn from_utf8_lossy(bytes: T) -> Cow<str>
        where T: LossyConvertibleBytes;
}
```

Where `LossyConvertibleBytes` would be implemented for `&[u8]`, `[u8]`, and `Cow<[u8]>`. Do you see any issues with this? I know this discussion would ultimately need to happen on the Rust repo, but I just wanted to sanity check it here first.

SimonSapin:
Regarding the standard library’s `from_utf8_lossy`: rather than a new trait, maybe `Into<Cow<'a [u8]>>` would work? Either way, feel free to pursue this separately but this repo won’t be able to rely on it any time soon :)

JoshMcguigan:
Is there any further work you'd like to see here before merging?

Shnatsel:
There seem to be no outstanding concerns on this PR. Is anything still blocking the merge?

src/form_urlencoded.rs
```diff
@@ -142,6 +142,10 @@ impl<'a> Iterator for ByteSerialize<'a> {
                 None => (self.bytes, &[][..]),
             };
             self.bytes = remaining;
+            // This unsafe is appropriate because we have already checked these
+            // bytes in byte_serialized_unchaged, which checks for a subset
```
jplatte, line 146:
```suggestion
            // bytes in byte_serialized_unchanged, which checks for a subset
```

JoshMcguigan: Good catch, I've fixed this.

JoshMcguigan:
I've rebased and updated this based on the latest changes. Let me know if this needs anything else.

djc:
I've applied some minor formatting tweaks to your commit and pushed it as 3b587ea32f87b40157ad1aa686e5b0bc1b38a5c4. Thanks for all the work!

Issue #557, opened 2019-10-19
WASM file size


MartinKavik:
Hi,

when I include this library to the [Seed](https://github.com/David-OConnor/seed) app and then compile and optimize it => the output WASM file size changes from **466** to **745** KB (+**279** KB).
Do you know how to reduce it? (feature flags, dependency changes, etc.)

Thank you!

SimonSapin:
The Unicode tables used for IDNA come to mind, but there is no plan to make those optional.

https://github.com/rustwasm/twiggy is made to investigate exactly this kind of question. Let us know what you find.

MartinKavik:
Thanks!

> Let us know what you find.

<details>
<summary>WASM without rust-url</summary>

```
$ twiggy top -n 30 package_bg.wasm
 Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────
         51594 ┊    13.36% ┊ "function names" subsection
         48499 ┊    12.56% ┊ data[0]
         12522 ┊     3.24% ┊ todomvc::view::h97225a52d3c16381
          9025 ┊     2.34% ┊ seed::dom_entity_names::styles::style_names::St::as_str::h034572cb759631f0
          8152 ┊     2.11% ┊ <seed::events::Ev as alloc::string::ToString>::to_string::hcc1c264f86668900
          7330 ┊     1.90% ┊ seed::dom_types::At::as_str::h7851cbcb93205c52
          6673 ┊     1.73% ┊ todomvc::todo_item::h1df45a0a040d8b33
          6311 ┊     1.63% ┊ seed::patch::patch_els::h957d52591338c6a5
          5900 ┊     1.53% ┊ <&mut serde_json::de::Deserializer<R> as serde::de::Deserializer>::deserialize_struct::h59632fa41a2529e5
          4124 ┊     1.07% ┊ core::num::flt2dec::strategy::dragon::format_shortest::hc3fdbdc1b58d0107
          3958 ┊     1.03% ┊ core::num::flt2dec::strategy::dragon::mul_pow10::he029ff2298318da8
          3803 ┊     0.99% ┊ seed::vdom::App<Ms,Mdl,ElC,GMs>::rerender_vdom::hffd3fa7a285240be
          3587 ┊     0.93% ┊ seed::dom_types::Tag::as_str::hd5afd3749469c804
[...]

PR #562, opened 2019-11-07
[idna] Bump unicode-normalization to 0.1.9

ashthespy:
With https://github.com/unicode-rs/unicode-normalization/pull/37 landing, this should help with compilation times and sizes (https://github.com/servo/rust-url/issues/557)

### DIFF
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -24,5 +24,5 @@ serde_json = "1.0"
 
 [dependencies]
 unicode-bidi = "0.3"
-unicode-normalization = "0.1.5"
+unicode-normalization = "0.1.9"
 matches = "0.1"


SimonSapin:
As far as I understand this PR is not necessary. A dependency declaration of `unicode-normalization = "0.1.5"` in Cargo means “Any 0.1.x version, where x >= 5”. So when building a project that doesn’t have a `Cargo.lock` file, Cargo will already pick the most recent matching version. If `Cargo.lock` already exists and specifies an older version, you can upgrade with `cargo update -p unicode-normalization` without changing rust-url’s `Cargo.toml`.

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

ashthespy:
Your right off course, sorry for the noise!

PR #563, opened 2019-11-08
Make percent-encoding no_std-compatible

dflemstr:
This makes the library no_std-compatible in a pretty straightforward manner by disabling methods that can't possibly work in a no_std environment (like Cow and dynamic string allocation)

As a second step, this could be made alloc-enabled but that is not needed for my use-case.


djc:
This seems like a good step forward, thanks!

Issue #568, opened 2019-12-05
feature: impl TryFrom<&str> for Url


yoshuawuyts:
Having `impl TryFrom<&str> for Url` would be fantastic because it would allow passing both `Url` and `&str` to the same API through an `impl TryInto<Url>` bound.

## Example

```rust
fn make_request(impl TryInto<Url>) -> Result<(), ParseError>;
```

```rust
// Pass a literal directly.
let mut req = make_request("https://example.com")?;

// Create a url, manipulate it, then pass it.
let url = Url::parse("https://example.com")?;
validate_url(&url)?;
let mut req = make_request(url)?;
```

## Caveats

The only caveat here is that `TryFrom` and `TryInto` were stabilized in Rust 1.34, and the minimum guaranteed Rust version for `url` is 1.33, which means this would likely would require a major version bump.

---

Thanks!

tmccombs:
> which means this would likely would require a major version bump

Or add a feature to enable this (and possibly other features that require a more recent version of rust)

yoshuawuyts:
@tmccombs oh brilliant; didn't think of that. Issued #569 with an implementation. Hope it's good!

_edit: Oops, misclicked and accidentally closed this issue._

yoshuawuyts:
Ah since https://github.com/servo/rust-url/issues/566 landed the minimum supported Rust version is now 1.36, which means the flag would no longer serve any purpose.

Issue #566, opened 2019-11-25
Minimum supported Rust version v.s. SemVer


rjloura:
When running clippy on rust 1.35 I encounter the following error:
```
error[E0658]: use of unstable library feature 'alloc': this library is unlikely to be stabilized in its current form or name (see issue #27783)
  --> /home/rui/.cargo/registry/src/github.com-1ecc6299db9ec823/smallvec-1.0.0/lib.rs:38:1
   |
38 | extern crate alloc;
   | ^^^^^^^^^^^^^^^^^^^

```

For my project the url crate is pulled in by gotham, but this occurs when I run clippy on `rust-url` by itself.  The tree is:
```
$ cargo tree
url v2.1.0 (/home/rui/git/rust-url)
├── idna v0.2.0 (/home/rui/git/rust-url/idna)
│   ├── matches v0.1.8
│   ├── unicode-bidi v0.3.4
│   │   └── matches v0.1.8 (*)
│   └── unicode-normalization v0.1.11
│       └── smallvec v1.0.0 <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
├── matches v0.1.8 (*)
└── percent-encoding v2.1.0 (/home/rui/git/rust-url/percent_encoding)
SNIP!
```

What I don't understand is how this tree was built.  `idna` is currently at version `0.2.0` and it specifies `unicode-normalization` version of `0.1.5` (https://github.com/servo/rust-url/blob/master/idna/Cargo.toml).

The only thing I did notice was that unicode-normalization was bumped to it's new version 3 days ago and that brought in the `smallvec v1.0.0` dependency. (https://github.com/unicode-rs/unicode-normalization/blame/master/Cargo.toml)

For this project I need to stay on rust 1.35.  I am not sure why the dependency tree changed without the `rust-url` version changing.

SimonSapin:
It looks like https://github.com/unicode-rs/unicode-normalization/pull/46 increased our minimum Rust version from under us. The last PR that we merged succeeded with 1.33:

https://github.com/servo/rust-url/blob/7d2c9d6ceb3307a3fad4c4bcb7f5b07b5e24c755/.travis.yml#L6

SimonSapin:
> For this project I need to stay on rust 1.35.

You may be able to do this by adding a dependency from your crate `unicode-normalization`, with a requirement for some older version number.

rjloura:
> > For this project I need to stay on rust 1.35.
> 
[...]

PR #569, opened 2019-12-05
impl TryFrom<'a str> for Url

yoshuawuyts:
Closes https://github.com/servo/rust-url/issues/568. Implements `TryFrom<'a str> for Url` gated behind the `try_from` feature. This should make it possible to use this conversion without issuing a breaking change. Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,6 @@ harness = false
 [lib]
 test = false
 
-[features]
-try_from = []
-
 [dev-dependencies]
 rustc-test = "0.3"
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,9 @@ harness = false
 [lib]
 test = false
 
+[features]
+try_from = []
+
 [dev-dependencies]
 rustc-test = "0.3"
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
It seems the build is failing for reasons unrelated to this patch. Only on ARM 64, and it seems something in smallvec isn't right.

yoshuawuyts:
ping @tmccombs do you maybe have any pointers on how to progress this? From the looks of the log it seems that `smallvec` now requires Rust `1.36` for `MaybeUninit` support, while `url` is trying to guarantee it builds on `1.33`.

__edit:__ it seems https://github.com/servo/rust-url/issues/566 was filed and resolved.

yoshuawuyts:
Updated to remove the feature flag since the minimum required Rust version is now 1.36, and the minimum required version for this feature is 1.34 :tada:

yoshuawuyts:
@valenting if you have time, could you perhaps take a look at this PR? Landing this would be a great help for several of our libraries.

yoshuawuyts:
@valenting would it be possible to publish a new release in the near future that includes this change? We're excited to use it!

PR #570, opened 2019-12-08
Implement `From<ParseError>` for `std::io::Error`

nvzqz:
Enables using `?` postfix operator and `try!` macro with `ParseError` when the enclosing scope expects an `io::Error`.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -8,6 +8,7 @@
 
 use std::error::Error;
 use std::fmt::{self, Formatter, Write};
+use std::io;
 use std::str;
 
 use host::{Host, HostInternal};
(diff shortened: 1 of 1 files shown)


dekellum:
I suspect you are only wanting this for convenience, but having done some work with the error types here and elsewhere, I think what you are asking to merge is, unfortunately, a bit of an anti-pattern and can cause confusion due to its implicit nature. 

IMO, its generally better do an explicit `map_err` for this conversion, if that is really what you want.

SimonSapin:
I think a more common pattern is to define a local error type that implement `From<…>` with the various error types from libraries you use.

djc:
As discussed in previous comments, I don't think this would be a sensible change. Thanks for trying to improve the library; let me know if you have further issues or concerns.

PR #571, opened 2019-12-09
Dummy PR

SimonSapin:


SimonSapin:
CI [fails on Rust 1.33](https://travis-ci.com/servo/rust-url/jobs/265209754#L242) in a PR with no significant change, because of https://github.com/unicode-rs/unicode-normalization/issues/50.

PR #573, opened 2020-01-02
Increment MSVR to 1.36.0

SimonSapin:
… because `unicode-normalization` did.

Fixes https://github.com/servo/rust-url/issues/566
Closes https://github.com/servo/rust-url/pull/572


valenting:
Thank you!

PR #574, opened 2020-01-04
Remove deprecated Error::description

AnderEnder:
`Error::description` has been documented as soft-deprecated since 1.27.0 (17 months ago). It is going to be hard-deprecated soon.

This PR:
- Moves an implementation of `Error::description` into internal `ParseError` type implementation similar to `SyntaxViolation`

Related PR: https://github.com/rust-lang/rust/pull/66919

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -62,7 +62,7 @@ macro_rules! simple_enum_error {
             __FutureProof,
         }
 
-        impl Error for ParseError {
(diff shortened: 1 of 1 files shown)


djc:
I've fixed this in a slightly different way in c9a7003e2936be83ed4e461d2423b22282e2637c. Thanks!

PR #584, opened 2020-01-31
decode size hint: Fix lower boundary

chrysn:
The size of a decoded URI can shrink only in fixed steps of 3 byte ("%xx") to 1 byte, thus the resulting size is always rounded up (2 bytes can not shrink any further), which is most easily implemented in integer division by adding `divisor - 1`.

---

Admittedly, this is not expected to make much difference; primarily it makes the implementation precise. Feel free to reject this if the overhead is not worth it. (Given this stems from a random crev crate review and not a practical problem).


djc:
Makes sense, thanks for the fix!

PR #547, opened 2019-09-16
form-urlencoded: Allow to encode key parameters without a values

Argentumbolo:
According to [#546](https://github.com/servo/rust-url/issues/546#issue-493793309)
Add methods:
    - Serializer<UrlQuery>::append_key_only(&mut self, name: &str) -> &mut Self;
    - Serializer<UrlQuery>::extend_keys_only<I, K>(&mut self, iter: I) -> &mut Self;
in order to allow query parameters without a values like:
    - https://example.org/exchange?date=20190915&json
    - https://example.org/exchange?date=20190915&xml
    - https://example.org/exchange?date=20190915&bin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-url/547)

PR #591, opened 2020-05-13
Parse standalone query-string keys as None

jayaddison:
This set of changes allows query-strings that contain standalone query-string keys (i.e. keys without attached `=` characters) to parse to the `None` Option value, indicating an absence-of-value.

The intent here is to distinguish between the empty string and absence-of-value when serializing and deserializing from query-string (and more generally, form data) contents.

This pull request is a companion pull request to https://github.com/servo/rust-url/pull/547 (no prior relationship there; this was discovered while searching for similar issues on GitHub).

I'm relatively new to Rust, and I'm certain there are many things about these changes that could be improved; I'd be very glad for any feedback.


src/form_urlencoded.rs
```diff
@@ -276,21 +276,29 @@ impl<'a, T: Target> Serializer<'a, T> {
     pub fn extend_pairs<I, K, V>(&mut self, iter: I) -> &mut Self
     where
         I: IntoIterator,
-        I::Item: Borrow<(K, V)>,
+        I::Item: Borrow<(K, Option<V>)>,
```
jayaddison, line 279: NB: Despite my potentially-misleading previous commit message, this interface is *not* currently backwards-compatible.  It feels like it should be possible to maintain a compatible API, but I'm not yet certain about the best order of steps to achieve that from here.

Observations include:

- That the underlying [name, value iterator](https://github.com/servo/rust-url/blob/28314752bf138a87bd8151f610d385629dec5710/src/form_urlencoded.rs#L46-L55) could be refactor to return a stream of tokens (see a [similar proposed refactor in cpython](https://github.com/python/cpython/pull/19949/files#diff-01126e4d609357e876a7fd7baa62ce18R748-R779))

- That it may be possible to define [shared behaviour](https://doc.rust-lang.org/book/ch10-02-traits.html#traits-defining-shared-behavior) on a common base parser trait, and then allow call-sites to explicitly opt-in to override and enable parsing of standalone keys

djc:
Before going into the code, I feel this is conceptually on shaky grounds. The [relevant bits](https://url.spec.whatwg.org/#urlencoded-parsing) of the URL standard seem to be saying that leaving out the `=` makes the value equivalent to the empty string, so your changes make sense within the Rust type system, they introduce a distinction that doesn't make sense in the context of the applicable standard. On top of that, it feels like the changes here would make the interface harder to use for most use cases in order to support what's arguably a niche use case.

As such, I'm inclined not to accept this PR, but I would be happy to hear your feedback.

jayaddison:
Thanks @djc - I expect I'd have a similar reaction if I were in your shoes as a maintainer :)

As you infer, adherence to the spec is always wise and that's likely where any distinction between empty-string and 'none-like' behaviour should indeed really be defined; that'd certainly make an approach like this more justifiable and compatible.

There's been some discussion around a similar proposal (referred to as concise query-string rendering) and whether that could also apply to `x-www-form-urlencoded`) in https://github.com/whatwg/url/issues/469 -- although that currently looks stalled (I could perhaps have approached the conversation more carefully).

Current standards aside -- although there's some small developer overhead here, I do genuinely believe that there's a large opportunity here to distinguish null and empty-string values in query strings and browser-submitted form data, not just in `rust-url` but across various platforms (JS, Python, Java/.NET, ...) and that encoding these values distinctly could improve safety and communication of those values a lot.  That said I don't know whether I have the individual energy to fight those battles upstream, so.. yep, we may just place this on pause or close it for now.

djc:
Reading through the linked WHATWG issue, I definitely don't see this happening at the standard level. Let's close this -- sorry for all the time you've spent on this only to get to this disappointing result, it looks like you did quite a bit of work.

jayaddison:
All good; it's been a learning experience - a good excuse to try out some Rust dev if nothing else :)

PR #593, opened 2020-05-20
Fix socket_addrs upgrade notes

slonopotamus:


slonopotamus:
No, it is even more broken. It should talk about `default_port` instead.

PR #604, opened 2020-06-15
Make crate about 50% lighter

Keats:
According to cargo-diet:

> Saved 54% or 250.1 KB in 9 files (of 466.0 KB and 22 files in entire crate)


tmccombs:
how does "include" work with workspaces? is this excluding the other creates like idna, percent_encoding, and data_url?

djc:
I think it would maybe make more sense to flatten the repository, moving the `url` crate into a `url` directory in the repository root.

djc:
I merged bf9bbf2f8cadf2792949d9e798c39c7221e29a07, which is a rebased version of your change. Thanks!

PR #607, opened 2020-06-16
Split form_urlencoded into its own crate

### LINKED ISSUE #605: Separate form_urlencoded into its own crate
Hi! I'm trying to trim [Ruma](https://github.com/ruma/ruma)'s dependencies, and noticed that the url crate, of which we only use the `form_urlencoded` module (which AFAICT doesn't require any external dependencies) is the only reason for the inclusion of `idna`, `unicode-bidi`, `unicode-normalization` and `smallvec` in our dep graph. Would a PR separating it out into its own crate like `percent-encoding` be considered?

jplatte:
Resolves #605.

I wasn't sure what version to give the new crate, so I went with 1.0.0.


valenting:
It all looks good to me but I asked @SimonSapin to also take a quick look.

form_urlencoded/src/lib.rs
```diff
@@ -13,11 +13,17 @@
...
+#[macro_use]
+extern crate matches;
+
 use percent_encoding::{percent_decode, percent_encode_byte};
-use query_encoding::{self, decode_utf8_lossy, EncodingOverride};
+use query_encoding::EncodingOverride;
 use std::borrow::{Borrow, Cow};
 use std::str;
 
+pub mod query_encoding;
```
SimonSapin, line 25: This module was previously private and made public here. Is this intentional? If so, why?

form_urlencoded/src/lib.rs
```diff
@@ -345,3 +327,20 @@ fn append_pair(
 fn append_encoded(s: &str, string: &mut String, encoding: EncodingOverride) {
     string.extend(byte_serialize(&query_encoding::encode(encoding, s.into())))
 }
+
+fn decode_utf8_lossy(input: Cow<[u8]>) -> Cow<str> {
```
SimonSapin, line 331: Why move this from the encoding module?

form_urlencoded/src/lib.rs
jplatte: Yes, this is intentional. This functionality is used both in form_urlencoded and in url, so I decided to make it a public module of the new crate. I could also duplicate it, like the decode_utf8_lossy function that was already duplicated between percent_encoding and url (now form_urlencoded).

form_urlencoded/src/lib.rs
jplatte: I think this is just a fragment of me thinking that this was the only functionality form_urlencoded uses from that module and moving it out, to then move query_encoding back into the url crate.

jplatte:
@SimonSapin What should I do with the `query_encoding` module and `decode_utf8_lossy`?

SimonSapin:
`EncodingOverride` is the one item that needs to be shared. I’d rather have it reexported from the root of the crate(s) than adding a new module to the public API. As to `encode`, duplicating it is kinda unfortunate but it’s so short it could even be manually inlined.

jplatte:
> `EncodingOverride` is the one item that needs to be shared. I’d rather have it reexported from the root of the crate(s) than adding a new module to the public API.

Done.

> As to `encode`, duplicating it is kinda unfortunate but it’s so short it could even be manually inlined.

I manually inlined into `url`s crate root. Should I inline it into its now only call site in `form_urlencoded` too, then?

form_urlencoded/src/query_encoding.rs
```diff
@@ -10,14 +10,14 @@ use std::borrow::Cow;
...
 
-pub(crate) fn encode<'a>(encoding_override: EncodingOverride, input: &'a str) -> Cow<'a, [u8]> {
+pub fn encode<'a>(encoding_override: EncodingOverride, input: &'a str) -> Cow<'a, [u8]> {
     if let Some(o) = encoding_override {
         return o(input);
     }
     input.as_bytes().into()
 }
 
-pub(crate) fn decode_utf8_lossy(input: Cow<[u8]>) -> Cow<str> {
```
SimonSapin, line 20: The net effect is the same since the module is private, but could you revert this to `pub(crate)`? Thanks!

jplatte: I'm guessing you mean both of these. Done.

SimonSapin:
Looks good, thanks!

SimonSapin:
https://crates.io/crates/form_urlencoded is up

PR #617, opened 2020-07-08
Set up GitHub Actions CI

djc:
I guess this needs someone to enable GitHub Actions for this repo. Also doesn't test WASM support yet.


SimonSapin:
Why?

djc:
Supports all the platforms in a single provider. In my experience, quite a bit faster than Travis. Travis seems to get very little maintenance.

PR #622, opened 2020-08-08
Add nicer Display implementation for idna::Errors

clarfon:
This offers a custom string for each error variant and separates them by commas.

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -480,12 +480,32 @@ impl Config {
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 enum Error {
     PunycodeError,
+
+    // https://unicode.org/reports/tr46/#Validity_Criteria
     ValidityCriteria,
     DissallowedByStd3AsciiRules,
     DissallowedMappedInStd3,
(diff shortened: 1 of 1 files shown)


idna/src/uts46.rs
```diff
@@ -499,6 +519,12 @@ impl StdError for Errors {}
 
 impl fmt::Display for Errors {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self)
+        let (first, rest) = self.0.split_first().expect("should be at least one error");
+        f.write_str(first.as_str())?;
+        for x in rest {
+            f.write_str(", ")?;
+            f.write_str(x.as_str())?;
```
djc, line 526:
```suggestion
        for (i, err) in &self.0 {
            if i > 0 {
                f.write_str(", ")?;
            }
            f.write_str(x.as_str())?;
        }
```

clarfon:
Fixed the build and made the changes you suggested. :)

djc:
This seems like an improvement, thanks! I'll probably tweak it a bit from here.

PR #623, opened 2020-08-10
Update to latest WHATWG URL standard

djc:


djc:
I changed my commits to remove tests that failed in the early commits, such that each commit passes the test suite. Of course the final commit still updates the full test suite to match the WHATWG tests. I've then merged this -- closing.

PR #629, opened 2020-08-24
Fix non-idempotent serialization for ./ and ../ paths (fixes #601)

### LINKED ISSUE #601: URLs containing no host and a path starting with ./ or ../ are not serialized idempotently
Parsing a url like `scheme:/.//path` internally removes the `./` in the serialization ([per the spec](https://url.spec.whatwg.org/#path-state)), so parsing the serialized URL of `scheme://path` yields a URL that is parsed differently than the original one (ie. path is treated as a host now, instead of part of the URL's path).

Cases to consider:
* `scheme:/.//path`
* `scheme:/..//path`
* `scheme:/./..//path` (any leading mixed collection of `./` and `../`)

My reading of the spec is that this might be a shortcoming that needs to be addressed there, rather than just an implementation bug. I can envision a couple possible solutions:
* don't truncate `./` or `../` if it's the first component of a path, avoiding the issue entirely
* truncate `/` when the previous segment is empty (resulting in `scheme:/path`)

djc:


valenting:
I think I didn't handle the code path for the setters :thinking:

annevk:
In the standard this ended up impacting the URL serializer exclusively. It would surprise me a bit of all the parser changes here were needed.

valenting:
> In the standard this ended up impacting the URL serializer exclusively. It would surprise me a bit of all the parser changes here were needed.

rust-url builds the serialization as it parses the input.
For speed mostly. This way serializing is just returning a reference to the serialization which is just one string - no allocations.

zealousidealroll (approved):

PR #635, opened 2020-08-25
idna: prevent allocations (mainly) in Punycode decoding

djc:


djc:
@valenting can you review the API introduced in 5c91e06? (Feel free to look at the other stuff too, but it's a bit gnarly and only amounts to a modest performance improvement for more complex invocations of `to_unicode()`.)

valenting:
> @valenting can you review the API introduced in [5c91e06](https://github.com/servo/rust-url/commit/5c91e0629b0d05b223609a4bcc07f9947fec3626)? (Feel free to look at the other stuff too, but it's a bit gnarly and only amounts to a modest performance improvement for more complex invocations of `to_unicode()`.)

I think it's generally OK. As a side note, do we use Idna.normalized for anything? Or can it be just a local buffer?

djc:
What do you mean by "local buffer"? `processing()` needs a local buffer. Keeping it in `Idna` amortizes the cost of the allocation over all calls to `Idna::to_ascii()` and `Idna::to_unicode()`.

valenting:
> What do you mean by "local buffer"? `processing()` needs a local buffer. Keeping it in `Idna` amortizes the cost of the allocation over all calls to `Idna::to_ascii()` and `Idna::to_unicode()`.

In the sense that it doesn't seem to be used outside of `processing()`. But maybe I'm missing something.

djc:
No, that is correct. I just though amortizing the cost of allocating that buffer would be better for performance than allocating it locally for every call to `processing()`.

djc:
Newer version in #645.

PR #636, opened 2020-08-25
idna: switch to stack-allocated simple Errors struct

djc:
Presumably preserving the amount of errors of each type is less important
than avoiding the allocations resulting from the use of Vec.

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -82,7 +82,7 @@ fn find_char(codepoint: char) -> &'static Mapping {
         .unwrap()
 }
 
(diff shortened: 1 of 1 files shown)


idna/src/uts46.rs
```diff
@@ -550,20 +512,40 @@ fn is_bidi_domain(s: &str) -> bool {
 ///
 /// This is opaque for now, only indicating the presence of at least one error.
```
valenting, line 513: Is this comment still accurate?

valenting (approved):
Looks good! One minor question.

idna/src/uts46.rs
djc: I've tweaked the comment.

PR #637, opened 2020-08-25
Show components in Url's Debug implementation

### LINKED ISSUE #596: Url's Debug trait should show parsed components
[Per the Debug trait docs](https://doc.rust-lang.org/std/fmt/trait.Debug.html):

> Debug should format the output in a programmer-facing, debugging context.
> Generally speaking, you should just derive a Debug implementation.

In Url's case, the straightforward `#[derive(Debug)]` wouldn't be that helpful since it would show the indexes into the backing string. [Right now `Url`'s Debug implementation emits its unparsed input String](https://github.com/servo/rust-url/blob/v2.1.1/src/lib.rs#L2339). But I think it would be really useful to instead give a Debug implementation that showed the parsed components as fields. For instance:

```
println!("{:?}", url::Url::parse("http://[::1]:80/file#anchor")?)

...

Url { scheme: "http", host: Some(Ipv6(::1)), port: None, path: "/file", query: None, fragment: Some("anchor") }
```

Here's a patch that implements that behavior. I'm happy to submit a PR, but I wanted to gather feedback on the general idea first.

```patch
diff --git a/src/lib.rs b/src/lib.rs
index 8aff08a..dd48969 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2346,7 +2346,15 @@ impl fmt::Display for Url {
 impl fmt::Debug for Url {
     #[inline]
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Debug::fmt(&self.serialization, formatter)
+        formatter.debug_struct("Url")
+            .field("scheme", &self.scheme())
+            .field("host", &self.host())
+            .field("port", &self.port())
+            .field("path", &self.path())
+            .field("query", &self.query())
+            .field("fragment", &self.fragment())
+            .finish();
+        Ok(())
     }
 }
```

I've also been discussing a similar issue on the `http` repo: https://github.com/hyperium/http/issues/423.

jsha:
Fixes #596


djc:
Thanks! Would you mind fixing up the formatting and clippy issues?

djc:
(Do you use Rust/this project for your work? I thought LetsEncrypt mostly uses Go?)

jsha:
Done, thanks! And yep, Let's Encrypt is mainly Go. We've been thinking about spinning up expertise in Rust for a new project, and so I've been learning Rust and poking around at various libraries.

djc:
Thanks!

I'd like to hear more about the new project -- I might be able to help out a little.

PR #638, opened 2020-08-28
url: add TryFrom<&[u8]> for Url and TryFrom tests

acfoltzer:
Per [discussion on Matrix](https://matrix.to/#/!wurHQemAfsdBxUYHeU:mozilla.org/$MLu4rFb7uBI54AT4-UwKZSyEMXELJRM7Rqm1dpGZ6kM?via=mozilla.org&via=matrix.org).

I didn't realize when first asking that `TryFrom<&str>` is already implemented, but hasn't been released. So, I ended up just adding the impl for byte slices, an error variant for invalid UTF-8 inputs, and tests for the impls.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -2339,6 +2339,15 @@ impl<'a> TryFrom<&'a str> for Url {
     }
 }
 
+impl<'a> TryFrom<&'a [u8]> for Url {
+    type Error = ParseError;
+
+    fn try_from(bytes: &'a [u8]) -> Result<Self, Self::Error> {
+        let s = std::str::from_utf8(bytes).map_err(|_| ParseError::InvalidUtf8)?;
+        Url::parse(s)
+    }
+}
+
 /// Display the serialization of this URL.
 impl fmt::Display for Url {
     #[inline]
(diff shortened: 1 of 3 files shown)


acfoltzer:
I believe CI is failing because a credential is not available from my fork.

SimonSapin:
I feel this should not be implemented. URL parsing is defined over Unicode input. `&[u8]` and `&str` are not interchangeable representations of "the same thing", Unicode and bytes are separate "domains" of data and the choice of strict UTF-8 decoding for converting between them should be explicit to users.

(Unfortunately the Matrix link doesn’t seem to work from here.)

acfoltzer:
@SimonSapin I definitely agree with you on these being separate domains of data, but I believe that's reflected by the use of `TryFrom` (rather than `From`) and the addition of the `InvalidUtf8` error variant.

There wasn't much discussion on Matrix really:
image

djc:
FWIW, like @SimonSapin I think this is a bad fit.

acfoltzer:
No worries, again it's just a convenience and not a dealbreaker for our adoption.

PR #643, opened 2020-10-04
remove dependency on `matches` crate - `matches!` is now in Rust stable

utkarshkukreti:
`matches!` was stabilized in Rust 1.42.0: https://doc.rust-lang.org/stable/std/macro.matches.html.

### DIFF
--- a/data-url/Cargo.toml
+++ b/data-url/Cargo.toml
@@ -8,9 +8,6 @@ license = "MIT OR Apache-2.0"
 edition = "2018"
 autotests = false
 
-[dependencies]
-matches = "0.1"
-
 [dev-dependencies]
 rustc-test = "0.3"
 serde = {version = "1.0", features = ["derive"]}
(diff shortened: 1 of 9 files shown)


utkarshkukreti:
Just realized the CI tests with Rust 1.36.0. Feel free to close the PR if you still want to support < 1.42.0! Or if you'd like I can bump it to 1.42.0.

SimonSapin:
This dependency is very small, removing it does not seem worth incrementing the minimum Rust version to me.

utkarshkukreti:
> This dependency is very small, removing it does not seem worth incrementing the minimum Rust version to me.

Makes sense. Sorry I did not notice the Rust version in the workflow before I sent this PR. The dependency can be removed later then, when the MSRV has to be increased for some other reason.

SimonSapin:
No worries! And yeah that sounds like a good plan.

PR #645, opened 2020-10-19
More efficient IDNA APIs

djc:
This is the latest version, which makes some improvements to the error handling in the idna crate.


djc:
(Failure of the audit CI job is fixed in #646.)

djc:
Something like this has been merged.

Issue #639, opened 2020-08-30
`url_serde` asking for unrelated trait bounds


danjenson:
Using the `serde` and `url_serde` crates, I get errors suggesting I need to fulfill unrelated trait bounds:

```
   Compiling ser v0.1.0 (/data/scratch/ser)
error[E0277]: the trait bound `for<'a> url_serde::Ser<'a, url::Url>: a::_::_serde::Serialize` is not satisfied
  --> src/b.rs:4:10
   |
4  | #[derive(Serialize, Deserialize)]
   |          ^^^^^^^^^ the trait `for<'a> a::_::_serde::Serialize` is not implemented for `url_serde::Ser<'a, url::Url>`
   | 
  ::: /home/danj/.cargo/registry/src/github.com-1ecc6299db9ec823/url_serde-0.2.0/src/lib.rs:77:46
   |
77 |     where S: Serializer, for<'a> Ser<'a, T>: Serialize
   |                                              --------- required by this bound in `url_serde::serialize`
   |
   = help: the following implementations were found:
             <url_serde::Ser<'a, std::option::Option<url::Url>> as a::_::_serde::Serialize>
             <url_serde::Ser<'a, url::Url> as a::_::_serde::Serialize>
             <url_serde::Ser<'a, url::host::Host<String>> as a::_::_serde::Serialize>
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `url_serde::De<url::Url>: a::_::_serde::Deserialize<'_>` is not satisfied
   --> src/b.rs:4:21
    |
4   | #[derive(Serialize, Deserialize)]
    |                     ^^^^^^^^^^^ the trait `a::_::_serde::Deserialize<'_>` is not implemented for `url_serde::De<url::Url>`
    | 
   ::: /home/danj/.cargo/registry/src/github.com-1ecc6299db9ec823/url_serde-0.2.0/src/lib.rs:158:40
    |
158 |     where D: Deserializer<'de>, De<T>: Deserialize<'de>
    |                                        ---------------- required by this bound in `url_serde::deserialize`
    |
    = help: the following implementations were found:
              <url_serde::De<std::option::Option<url::Url>> as a::_::_serde::Deserialize<'de>>
              <url_serde::De<url::Url> as a::_::_serde::Deserialize<'de>>
              <url_serde::De<url::host::Host> as a::_::_serde::Deserialize<'de>>
[...]

PR #647, opened 2020-11-06
Add serde feature hint

nkconnor:
As far as I can tell this information is only tucked away in UPGRADING.md, and not mentioned in https://docs.rs/url_serde/0.2.0/url_serde/

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -73,6 +73,9 @@ assert!(data_url.fragment() == Some(""));
 # run().unwrap();
 ```
 
+## Serde
+
+Add `serde` feature to complement `url::Url` with the `Deserialize` and `Serialize`.
 
 # Base URL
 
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

url/src/lib.rs
```diff
@@ -73,6 +73,9 @@ assert!(data_url.fragment() == Some(""));
 # run().unwrap();
 ```
 
+## Serde
+
+Add `serde` feature to complement `url::Url` with the `Deserialize` and `Serialize`.
```
djc, line 78: Nit: I would find this clearer as "Enable the `serde` feature to include `Deserialize` and `Serialize` implementations for `url::Url`."

djc (commented):
Thanks for adding this! Small suggestion about the wording. (This is an unofficial review.)

url/src/lib.rs
nkconnor: Agreed, this is much better, thanks!

djc (approved):

tmccombs (approved):

nkconnor:
I think this may also close #639

PR #650, opened 2020-11-18
mention feature serde in docs

kosta:
This is a very important, but really well hidden feature (if don't exactly know where to look).

Not anymore :)

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -103,6 +103,18 @@ assert_eq!(css_url.as_str(), "http://servo.github.io/rust-url/main.css");
 # Ok(())
 # }
 # run().unwrap();
+```
+
+# Feature: `serde`
+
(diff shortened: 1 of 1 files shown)


kosta:
Could you maybe mention this snippet from UPGRADING.md at these two locations as well? When googling for ["rust url serde"](https://www.google.com/search?q=rust+url+serde), the `url_serde` crate is still the first google hit.

https://crates.io/crates/url_serde
https://docs.rs/url_serde/0.2.0/url_serde/

I can branch the last commit with url_serde and perform the doc change myself, but you would need to merge and publish it.

    `url_serde` is no longer required to use `Url` with Serde 1.x. Remove references to `url_serde` and enable the `serde` feature instead.

     ```toml
     # Cargo.toml
     [dependencies]
     url = { version = "2.0", features = ["serde"] }
     ```

PR #651, opened 2020-11-21
Generalize PartialEq impl for Host

mbrubeck:
This allows `Host<String>` to be compared with a `Host<&str>`, for example.

### DIFF
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -38,7 +38,7 @@ impl From<Host<String>> for HostInternal {
 
 /// The host name of an URL.
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
-#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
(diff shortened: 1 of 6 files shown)


mbrubeck:
I included commits to fix existing CI failures on the master branch.  I can drop these or separate them into different PR(s) if desired.

url/src/host.rs
```diff
@@ -172,6 +172,20 @@ impl<S: AsRef<str>> fmt::Display for Host<S> {
     }
 }
 
+impl<S, T> PartialEq<Host<T>> for Host<S>
+where
+    S: PartialEq<T>,
+{
+    fn eq(&self, other: &Host<T>) -> bool {
+        match (self, other) {
+            (Host::Domain(a), Host::Domain(b)) => *a == *b,
```
djc, line 181: (Unofficial review:) this looks good to me, although tiny nit: I think we can do without the `*` operators here, and that might look slightly nicer IMO.

djc (approved):

url/src/host.rs
mbrubeck: Done.

PR #653, opened 2020-11-27
idna: optimize punycode handling, add amortized allocation API, improve error reporting

djc:
Benchmarks before:

```
test to_ascii_merged         ... bench:       2,102 ns/iter (+/- 309)
test to_ascii_puny_label     ... bench:       1,214 ns/iter (+/- 220)
test to_ascii_simple         ... bench:         246 ns/iter (+/- 34)
test to_unicode_ascii        ... bench:          74 ns/iter (+/- 14)
test to_unicode_merged_label ... bench:       2,127 ns/iter (+/- 400)
test to_unicode_puny_label   ... bench:       1,328 ns/iter (+/- 353)
```

Benchmarks after:

```
test to_ascii_merged         ... bench:       1,937 ns/iter (+/- 497)
test to_ascii_puny_label     ... bench:       1,147 ns/iter (+/- 254)
test to_ascii_simple         ... bench:         247 ns/iter (+/- 74)
test to_unicode_ascii        ... bench:          78 ns/iter (+/- 14)
test to_unicode_merged_label ... bench:       1,848 ns/iter (+/- 239)
test to_unicode_puny_label   ... bench:       1,158 ns/iter (+/- 312)
```

Across a few runs, I see performance improvements like this:

to_ascii_merged: 4%
to_ascii_puny_label: 12%
to_ascii_simple: 3%
to_unicode_ascii: -5% (but on a much smaller base)
to_unicode_merged_label: 14%
to_unicode_puny_label: 13%

This pretty much comes from the first four commits (each commit is logically separate and can be reviewed as such, although some of the first commits don't pay off in performance until the fourth one). The fifth commit adds a separate API that allows reusing allocations -- any performance benefits of this aren't reflected in the benchmarks because they only do one operation per benchmark. The final three commits instead improve error handling (although they don't change the API).

All of this comes at the cost of +300 SLOC in the idna crate.

I'm using this for my work project, so would like to get something like this merged.

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -326,7 +326,12 @@ fn is_valid(label: &str, config: Config) -> bool {
 
 /// http://www.unicode.org/reports/tr46/#Processing
 #[allow(clippy::manual_strip)] // introduced in 1.45, MSRV is 1.36
-fn processing(domain: &str, config: Config) -> (String, Errors) {
+fn processing(
+    domain: &str,
+    config: Config,
+    normalized: &mut String,
+    output: &mut String,
+) -> Errors {
     // Weed out the simple cases: only allow all lowercase ASCII characters and digits where none
     // of the labels start with PUNYCODE_PREFIX and labels don't start or end with hyphen.
     let (mut prev, mut simple, mut puny_prefix) = ('?', !domain.is_empty(), 0);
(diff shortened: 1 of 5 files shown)


djc:
Also happy to split this in separate PRs if that helps.

idna/src/uts46.rs
```diff
@@ -279,11 +279,11 @@ fn passes_bidi(label: &str, is_bidi_domain: bool) -> bool {
 /// V1 (NFC) and V8 (Bidi) are checked inside `processing()` to prevent doing duplicate work.
 ///
 /// http://www.unicode.org/reports/tr46/#Validity_Criteria
-fn is_valid(label: &str, config: Config) -> bool {
+fn is_valid(label: &str, config: Config, errors: &mut Errors) {
```
valenting, line 282: since we're not returning anything, maybe we should call it `check_validity`?

valenting:
Great work on this one. I'll merge it ASAP.

idna/src/uts46.rs
djc: Fair enough, fixed that.

PR #659, opened 2020-12-10
Include username and password in Url's Debug implementation

e00E:
While looking into improving fuzzing I noticed that these two fields are missing which can make it confusing when two urls aren't equal but their debug is equal.


djc (approved):
Ah, that explains some surprising things I saw recently. Thanks!

e00E:
CI seems to be broken on coverage. Don't think it's related to this PR.

valenting (changes requested):
I think those fields should go in the appropriate position - after the scheme?

e00E:
I have no strong opinion on the order. Moved them to your suggested position.

PR #660, opened 2020-12-10
Update fuzz crate 

e00E:
Following the instructions from the fuzzing book the fuzz crate has been
recreated.
I renamed the file because it makes it slightly more clear what it does.
Putting the second parse and the assert on different lines gives better
panic messages because it makes it easier to distinguish whether the
unwrapping or the assert failed.

To run the fuzzer:

```
cargo +nightly install cargo-fuzz
cd url
cargo +nightly fuzz run parse_roundtrip
```


djc:
Can you change your commits to (a) leave the username/password commit out of this PR, and (b) tell git about the rename of the fuzzing module? That would make it easier to confidently review this.

e00E:
Removed the debug pr and removed the fuzz target rename too.

djc (approved):
LGTM, thanks!

valenting:
I think this target is being currently fuzzed.
See https://github.com/google/oss-fuzz/pull/3948

@DavidKorczynski does this change look OK to you?

e00E:
If this fuzz target is supposed to be run already then something is going wrong with that process seeing as me manually running it found a failing test case in a couple of seconds.

DavidKorczynski:
The current fuzzer is being run by OSS-Fuzz and I can see there is also an assert bug listed on OSS-Fuzz. So I guess the question is whether the changes updated by @e00E are interesting to have - I think that's up to the maintainers to decide? 

To be clear though, the fuzzer will continue to be run by OSS-Fuzz if these gets merged in.

PR #665, opened 2021-01-11
small documentation inaccuracy

Johann150:
Fixed a small inaccuracy in the docs for [`url::Url::host`](https://docs.rs/url/2.2.0/url/struct.Url.html#method.host). It did only say that domains would be IDNA encoded when in fact as documented at [`url::Host::Domain`](https://docs.rs/url/2.2.0/url/enum.Host.html#variant.Domain), for URLs with a "non-special" scheme, domains will be percent encoded.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -842,7 +842,8 @@ impl Url {
     }
 
     /// Return the parsed representation of the host for this URL.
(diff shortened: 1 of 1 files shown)


djc:
Unofficial review: LGTM. Thanks for fixing this!

Issue #666, opened 2021-01-23
API for converting a base URL and another URL into a relative one


sdroege:
This currently does not seem to exist in `url` and would basically be the inverse of `Url::join`. I have a implementation for this in my code but if there's interest I could create a PR to add it here. Opinions?

Code with some tests follows.

```rust
/// Extension trait for [`url::Url`]
pub trait UrlExt {
    /// Returns `url` relative to `self` if `self` is a base of it.
    fn make_relative(&self, url: &url::Url) -> Option<String>;
}

impl UrlExt for url::Url {
    fn make_relative(&self, url: &url::Url) -> Option<String> {
        if self.cannot_be_a_base() {
            return None;
        }

        // Scheme, host and port need to be the same
        if self.scheme() != url.scheme() || self.host() != url.host() || self.port() != url.port() {
            return None;
        }

        // We ignore username/password at this point

        // The path has to be transformed
        let mut relative = String::new();

        // Extract the filename of both URIs, these need to be handled separately
        fn extract_path_filename(s: &str) -> (&str, &str) {
            let last_slash_idx = s.rfind('/').unwrap_or(0);
            let (path, filename) = s.split_at(last_slash_idx);
            if filename.is_empty() {
                (path, "")
            } else {
                (path, &filename[1..])
            }
        }

        let (base_path, base_filename) = extract_path_filename(self.path());
        let (url_path, url_filename) = extract_path_filename(url.path());

        let mut base_path = base_path.split('/').peekable();
        let mut url_path = url_path.split('/').peekable();

        // Skip over the common prefix
        while base_path.peek().is_some() && base_path.peek() == url_path.peek() {
            base_path.next();
            url_path.next();
        }

        // Add `..` segments for the remainder of the base path
        for base_path_segment in base_path {
            // Skip empty last segments
            if base_path_segment.is_empty() {
                break;
            }

            if !relative.is_empty() {
                relative.push('/');
            }

            relative.push_str("..");
        }

        // Append the remainder of the other URI
        for url_path_segment in url_path {
            if !relative.is_empty() {
                relative.push('/');
            }

            relative.push_str(url_path_segment);
        }

        if base_filename != url_filename {
            if url_filename.is_empty() {
                relative.push('/');
            } else {
                if !relative.is_empty() {
                    relative.push('/');
                }
                relative.push_str(url_filename);
            }
        }

        // Query and fragment are only taken from the other URI
        if let Some(query) = url.query() {
            relative.push('?');
[...]

PR #668, opened 2021-01-28
Add Url::make_relative() as the inverse of Url::join()

sdroege:
This creates a relative URL from a base URL and another absolute URL.

Fixes https://github.com/servo/rust-url/issues/666

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -320,6 +320,8 @@ impl Url {
 
     /// Parse a string as an URL, with this URL as the base URL.
     ///
+    /// The inverse of this is [`make_relative`].
+    ///
     /// Note: a trailing slash is significant.
     /// Without it, the last path component is considered to be a “file” name
     /// to be removed to get at the “directory” that is used as the base:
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -349,11 +351,144 @@ impl Url {
...
+    ///
+    /// ```rust
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
+    /// # fn run() -> Result<(), ParseError> {
+    /// let base = Url::parse("https://example.net/a/b.html")?;
+    /// let url = Url::parse("https://example.net/a/c.png")?;
+    /// let relative = base.make_relative(&url);
+    /// assert_eq!(relative.as_deref(), Some("c.png"));
```
sdroege, line 374:
```
---- src/lib.rs - Url::make_relative (line 366) stdout ----
error[E0599]: no method named `as_deref` found for type `std::option::Option<std::string::String>` in the current scope
```

What's the minimum supported Rust version here? `Option::as_deref()` is stable since 1.40

sdroege:
The clippy failures are all unrelated to this PR. I can submit another PR for them if you agree that all those should be fixed, or let me know which of the lints you'd like to disable (personally I'm not a fan of `clippy::manual-range-contains`...)

url/src/lib.rs
djc: CI runs with 1.36.0.

sdroege: Do you want to keep that, and if so, which of the many patterns to solve this would you prefer? `relative.as_ref().map(|s| s.as_str())`? :)

djc:
Yup, sorry about the clippy failures. Another PR would be nice, though a separate commit in this PR is also fine. I think we should fix all of them except the ones about the unit errors, because those would break backwards compatibility.

url/src/lib.rs
djc: This one seems fine.

sdroege: Done, thanks!

sdroege:
clippy fixes are in https://github.com/servo/rust-url/pull/669

sdroege:
Rebased on top of #669

valenting:
Thanks for this PR. It looks really nice.
Would you mind adding a couple of tests for the negative? ie? returning None?

sdroege:
Sure, will add a few tests for that later or tomorrow! Thanks for the review

sdroege:
Added tests for the 4 possible error cases.

sdroege:
And rebased for the merge conflicts. @valenting anything else missing here?

PR #674, opened 2021-02-08
url: add the authority method

unleashed:
This is a revisit of #15 to add a method to be able to get the authority on its own.

Much has changed in the meantime, and the method can now be implemented without needing an additional struct.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -792,6 +792,8 @@ impl Url {
     /// URLs that do *not* are either path-only like `unix:/run/foo.socket`
     /// or cannot-be-a-base like `data:text/plain,Stuff`.
     ///
+    /// See also the `authority` method.
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -677,6 +679,46 @@ impl Url {
         self.slice(self.scheme_end..).starts_with("://")
     }
 
+    /// Return the authority of this URL as an ASCII string.
+    ///
+    /// Non-ASCII domains are punycode-encoded per IDNA if this is the host
+    /// of a special URL, or percent encoded for non-special URLs.
+    /// IPv6 addresses are given between `[` and `]` brackets.
```
djc, line 686: Are you sure that this documentation is actually how it works? Would be good to add a unit test that covers some of these cases, with at least one IDNA-encoded domain, one percent-encoded host and an IPv6 host.

djc (commented):
Thanks for working on this, I think it makes sense to add this. Before we merge it, I'd look to see a little more test coverage to ensure the current documentation matches the implementation.

url/src/lib.rs
unleashed: Thanks for thr review, @djc.

Yes, I am sure - the documentation is taken from methods that already state this and are implemented in a similar fashion by taking an internal slice. I've added another commit with a few more doc tests to exercise what is being said here and also added a note on ports for special URLs, which are omitted from the authority if they match the well known scheme port.

djc: Thanks for adding the tests!

I'd prefer to have this as unit tests not doctests, otherwise the example code gets a bit too bulky for my taste.

unleashed: Agreed, it's a bit too much for docs. Moved those to unit tests leaving just a few examples and also took the opportunity to add unit tests for `has_authority`.

djc (approved):

url/src/lib.rs
djc: Thanks!

unleashed:
Rebased to fix conflicts.

SimonAdameit:
Hi, I'd be interested in having this. What's still required to get it merged?

unleashed:
I don't know of any specific roadblock other than having a maintainer review and approve it.

pehala:
I would like to revive this PR, is there anything required other than rebasing to the current main?

valenting:
Rebase on the latest changes for the CI to run.
Then we can probably merge this.

unleashed:
@pehala @valenting rebased on top of latest master.

valenting (approved):
Thanks!

PR #675, opened 2021-02-08
Use assert!()'s built-in formatting mechanics

djc:
As suggested by clippy. Apparently this becomes a hard error in the 2021 edition.


djc:
(The warning about the `errors` field not being read is a false positive, tracked [here](https://github.com/rust-lang/rust/issues/81626).)

PR #676, opened 2021-02-16
Implement support for reporting errors on invalid IDNA2008 characters 

djc:

PR #677, opened 2021-02-16
Optimize IDNA tables

djc:
Performance changes over 3 runs of the benchmarks:

to_ascii_merged: -5.2% (new: ~1922ns)
to_ascii_puny_label: -3.7% (new: ~1122ns)
to_ascii_simple: +2.0% (new: ~260ns, I think this is not as bad as it looks due to an outlier)
to_unicode_ascii: -1.2% (new: ~79ns)
to_unicode_merged_label: -8.8% (new: ~1858ns, I think this is not as good as it looks due to an outlier)
to_unicode_puny_label: -1.4% (new: ~1133ns)

It also appears to shave ~2.3% off the binary size for the rlib and rmeta files.


djc:
(The linting failures are due to clippy warnings that are fixed as part of #676.)

PR #678, opened 2021-02-17
idna: bump version to 0.2.2

djc:
Would be nice to get this published!


djc:
(Should get #680 in first.)

PR #680, opened 2021-02-17
Update `unicode-normalization` to `0.1.17`

### LINKED ISSUE #679: `idna` declares incompatible dependencies
https://github.com/servo/rust-url/commit/e579c03677b864d54de5c080f6b44d61736fbd11 introduced this import:
```rust
use unicode_normalization::{is_nfc, UnicodeNormalization};
```
yet the version of `unicode-normalization` `idna` says it uses, `0.1.5`, doesn't have `is_nfc` (which seems to be introduced in the next patch version, `0.1.6`). The version isn't pinned, so cargo takes the newest compatible version and that doesn't matter... _unless_ someone has an old `Cargo.lock` file. Then `idna` doesn't compile.

I was updating an old project of mine and it took me like an hour to figure out I not only need to update the versions in the `Cargo.toml` file, but also remove `Cargo.lock` :sweat_smile:

mrobakowski:
Fixes #679

### DIFF
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -26,7 +26,7 @@ serde_json = "1.0"
 
 [dependencies]
 unicode-bidi = "0.3"
-unicode-normalization = "0.1.5"
+unicode-normalization = "0.1.17"
 matches = "0.1"
 
 [[bench]]


djc (approved):
LGTM, thanks!

PR #686, opened 2021-02-22
url: deprecate into_string for Into<String> (fixes #685)

### LINKED ISSUE #685: Replace into_string with an implementation of Into<String>?
It seems that having `Into<String>` for `Url` is quite natural and can be used in more contexts. Thoughts?

If this idea sounds good, I can work on a PR.

dtantsur:


url/src/lib.rs
```diff
@@ -410,6 +410,8 @@ impl Url {
     ///
     /// # Examples
     ///
+    /// Using this deprecated method:
+    ///
```
djc, line 414: Let's remove the example usage of deprecated code.

dtantsur:
@djc updated. It seems that the coverage failure is infrastructure related (no valid token).

djc (approved):

djc:
LGTM, thanks!

PR #691, opened 2021-03-10
Use collections from the alloc crate in percent_encoding

grego:
Use collections from `alloc` in the `percent_encoding` crate, thus making it fully `no_std` compatible.

~~Additionally implements `From<PercendEncode>>` and `From<PercendDecode>>` for `beef::Cow`, using an optional `beef` feature gate.~~
Edit: ditched the beef implementation as it didn't pass all of the checks.

### DIFF
--- a/percent_encoding/Cargo.toml
+++ b/percent_encoding/Cargo.toml
@@ -6,6 +6,3 @@ description = "Percent encoding and decoding"
 repository = "https://github.com/servo/rust-url/"
 license = "MIT/Apache-2.0"
 edition = "2018"
-
-[dependencies]
-beef = { version = "0.5", optional = true }
--- a/percent_encoding/Cargo.toml
+++ b/percent_encoding/Cargo.toml
@@ -7,6 +7,5 @@ repository = "https://github.com/servo/rust-url/"
 license = "MIT/Apache-2.0"
 edition = "2018"
 
-[features]
-default = ["std"]
-std = []
+[dependencies]
+beef = { version = "0.5", optional = true }
--- a/percent_encoding/Cargo.toml
+++ b/percent_encoding/Cargo.toml
@@ -6,3 +6,7 @@ description = "Percent encoding and decoding"
 repository = "https://github.com/servo/rust-url/"
 license = "MIT/Apache-2.0"
 edition = "2018"
+
+[features]
+default = ["alloc"]
+alloc = []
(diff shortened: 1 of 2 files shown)


djc:
Some downstream users might not have alloc available, right? I think this makes their lives worse?

grego:
Fair enough. Added all the usage of `alloc` types behind an optional `alloc` feature gate.

grego:
Could you, please, publish a version with this PR? I need it for a project.

tesaguri:
Technically speaking, this is a breaking change if a downstream user specifies `default-features = false`, though I don't think it would be a problem in practice since `pecent-encoding` has had no feature flags before. I just thought it worth mentioning.

djc:
I wouldn't be surprised if there are people using default-features = false anyway out of habit.

tesaguri:
> I wouldn't be surprised if there are people using default-features = false anyway out of habit.

I've done a quick search for packages in [rust-lang/crates.io-index] whose most recently published version has dependency on `percent-encoding` with default features disabled.

[rust-lang/crates.io-index]: <https://github.com/rust-lang/crates.io-index>

```command
$ curl -fL -o crates.io-index.zip https://github.com/rust-lang/crates.io-index/archive/refs/heads/master.zip
$ stat -f '%Sm' -t '%FT%T%z' crates.io-index.zip
2021-12-08T17:43:38+0900
$ unzip crates.io-index.zip
$ find crates.io-index-master -type f ! -name config.json | xargs -n 253 tail -q -n 1 | jq -c 'select(.deps | any([.name == "percent-encoding", .default_features == false] | all)) | {name, vers}'
```

Result:

```json
{"name":"instagram-web-api","vers":"0.1.1"}
{"name":"twilight-lavalink","vers":"0.8.0"}
{"name":"twilight-http","vers":"0.8.0"}
{"name":"mendes","vers":"0.0.60"}
{"name":"atlassian-app-auth","vers":"1.0.1"}
{"name":"gcs-rsync","vers":"0.1.10"}
{"name":"cloud-storage-rs","vers":"0.4.3"}
{"name":"cloud-storage","vers":"0.11.0-rc.1"}
```

So there are indeed downstream crates with `default-features = false`.

I didn't build them with this change applied so I don't know if this breaks the above crates (I'm too nervous to build random packages in an unsandboxed environment and too lazy to set up a Crater-like sandboxed environment or manually review each package).

PR #698, opened 2021-04-10
fix: implement std::error::Error for data-url

### LINKED ISSUE #697: std::error::Error is not implemented for Errors in data-url crate
The `data-url` crate does not implement `std::error::Error` (and thus no human readable formatting) for the following errors in that crate:

- `DecodeError`
- `DataUrlError`
- `MimeParsingError`

Is there any reason for this? If not I would be glad to submit a PR.

lucacasonato:
This commit implements std::error::Error for errors in the `data-url`
crate.

Closes #697

### DIFF
--- a/data-url/src/forgiving_base64.rs
+++ b/data-url/src/forgiving_base64.rs
@@ -1,9 +1,25 @@
 //! <https://infra.spec.whatwg.org/#forgiving-base64-decode>
 
 use alloc::vec::Vec;
+use core::fmt;
 
 #[derive(Debug)]
 pub struct InvalidBase64(InvalidBase64Details);
+
+impl fmt::Display for InvalidBase64 {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.0 {
+            InvalidBase64Details::UnexpectedSymbol(code_point) => {
+                write!(f, "symbol with codepoint {} not expected", code_point)
+            }
+            InvalidBase64Details::AlphabetSymbolAfterPadding => {
+                write!(f, "alphabet symbol present after padding")
+            }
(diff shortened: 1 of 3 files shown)


data-url/src/forgiving_base64.rs
```diff
@@ -17,6 +32,18 @@ pub enum DecodeError<E> {
     WriteError(E),
 }
 
+impl<E> std::fmt::Display for DecodeError<E> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::InvalidBase64(inner) => write!(f, "base64 not valid: {:?}", inner),
```
djc, line 38: Since we still support Rust 1.36, you'll have to write out the full enum name here.

sunsided:
Hey @lucacasonato! I was looking for the exact same feature here. Any updates? 🙂

data-url/src/forgiving_base64.rs
arpankapoor: Could this please be merged now that the MSRV is 1.51?

filips123:
With Servo [resuming its development in 2023](https://servo.org/blog/2023/01/16/servo-2023/), could this PR maybe be merged?

@lucacasonato Also, could you still implement `std::error::Error` on `DecodeError` and other errors where it's not already implemented? Based on [this comment](https://github.com/servo/rust-url/issues/697#issuecomment-817246265), it could be added even without a breaking change. But even if it's still a breaking change, it's a fairly small one, so it probably shouldn't matter if there is `0.2.0` after a few years since the last release.

lucacasonato:
@valenting PTAL - this PR is rebased now.

valenting (approved):

Issue #61, opened 2014-12-16
Changing the  scheme can break things


Manishearth:
If you change the scheme from a non-relative one to a relative one without regenerating the scheme data, the Url object stops working properly. For example, parsing `view-source+http://google.com` and then using `slice()` on `url.scheme` to make it http again does not change the corresponding scheme data.

Perhaps there should be a `set_scheme()` method for setting the scheme safely?

samfoo:
This same issue affects some of the upcoming [HSTS code in servo](https://github.com/samfoo/servo/blob/hsts-preload/components/net/hsts.rs#L124).

If it's not reasonable to mutate the scheme, does it make more sense to make `scheme` private and require a function call to read it? This way you can't create a `Url` instance that has an invalid state.

SimonSapin:
@samfoo `http` and `https` are both "relative" schemes, though. What’s the problem you’re seeing with HSTS? It might not be the same as the original issue here.

samfoo:
A simple test case is probably the easiest way to show one of the problems (that the default port doesn't change). This is specifically the problem with the pull request for HSTS. I'm not sure how much other internal state the `Url` keeps that could potentially cause a problem.

```rust
#[test]
fn test_changing_protocol_should_not_retain_the_old_default_port() {
    let mut http_url = Url::parse("http://mozilla.org").unwrap();
    http_url.scheme = "https".to_string();

    assert_eq!(http_url.port_or_default().unwrap(), 443)
}
```

```
---- hsts::test_changing_protocol_should_not_retain_the_old_default_port stdout ----
	thread 'hsts::test_changing_protocol_should_not_retain_the_old_default_port' panicked at 'assertion failed: `(left == right)` (left: `80`, right: `443`)', /Users/sam/projects/servo/tests/unit/net/hsts.rs:19
```

SimonSapin:
I’m thinking of refactoring the parser and data structure to fix a bunch of things including this.

But in the mean time, you can work around this issue by using `url.defaul_port = Some(443)` when changing the scheme to HTTPS.

samfoo:
```rust
secure_url.relative_scheme_data_mut()
            .map(|scheme_data| {
                scheme_data.default_port = Some(443);
            });
```
Is the best I could come up with, since `default_port` is on the the scheme data, not the `Url`. But that is better than serialising and re-parsing which is the strategy I took up 'til now.

SimonSapin:
[...]

PR #707, opened 2021-04-27
`Url::port_int` implementation

### LINKED ISSUE #706: Method to get port regardless of defaults
At the moment, when using the [`port`](https://docs.rs/url/2.2.1/url/struct.Url.html#method.port) method on an URL that uses a known scheme with none or the default port, `port` will return `None`.
[`port_or_known_default`](https://docs.rs/url/2.2.1/url/struct.Url.html#method.port_or_known_default) will always return the default port for known scheme's unless a non-default port is set in the URL.

Currently there is no way to determine if the input string contained a port or not, for applications which do not want to support any default ports, regardless of scheme's, this is a current limitation.

I propose adding a method `port_int` (like `host_str`, bikeshedding is welcome) that returns whatever port is set in the URL string, regardless of defaults.

daxpedda:
Some questions arose up while I was doing this implementation.

- I implemented this by adding a field `port_int` to `Url`, my initial implementation actually handled it by storing the actual port always in `Url::port` and when using the method `Url::port()` it would filter based on scheme defaults. In addition the `Url::serialization` would always contain the correctly removed default port. This implementation caused a lot of edge cases to arise, so I avoided it by doing it in a separate field.
- Currently when changing a scheme with `Url::set_scheme`, `Url::port_int` is removed if it was the scheme default, see #61. Probably it would be more intuitive that if `Url::port_int` is empty, ergo the user didn't define a port (default or not), then having the new port be the default port for that scheme makes sense. But if the user deliberately defined the default port on the URL, changing the scheme shouldn't change the port.
  ```rust
  // this makes sense
  let mut url = Url::parse("http://example.net").unwrap();
  url.set_scheme("https").unwrap();
  assert_eq!(url.port(), None);
  assert_eq!(url.port_int(), None);
  assert_eq!(url.port_or_known_default(), Some(443));
  
  // this is not intuitive
  let mut url = Url::parse("http://example.net:80").unwrap();
  assert_eq!(url.port(), None);
  assert_eq!(url.port_int(), Some(80));
  assert_eq!(url.port_or_known_default(), Some(80));
  url.set_scheme("https").unwrap();
  assert_eq!(url.port(), None);
  assert_eq!(url.port_int(), None); // why does this vanish, probably expected to stay 80
  assert_eq!(url.port_or_known_default(), Some(443)); // probably expected to stay 80
  
  // even weirder, but you shouldn't use a assigned port to a custom protocol anyway
  let mut url = Url::parse("custom-protocol://example.net:80").unwrap();
  assert_eq!(url.port(), Some(80));
  assert_eq!(url.port_int(), Some(80));
  assert_eq!(url.port_or_known_default(), Some(80));
  url.set_scheme("http").unwrap();
  assert_eq!(url.port(), None);
  assert_eq!(url.port_int(), None);
  assert_eq!(url.port_or_known_default(), Some(80));
  url.set_scheme("https").unwrap();
  assert_eq!(url.port(), None);
  assert_eq!(url.port_int(), None);
  assert_eq!(url.port_or_known_default(), Some(443)); // port changed completely, I guess this is potentially an issue with the current implementation too
  ```
  I don't think this is a big deal, I would argue that this is a pretty bad use-case the url crate might not want to support.
- Comparing two `Url`s with different `Url::port_int` might turn out to be equal, because it compares `Url::serialization`, and default ports for scheme aren't reflected in the serialization.

---

Personally I'm fine with last two points, I consider them edge-cases that are fine for the url crate to ignore.

Fixes #706.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -182,6 +182,7 @@ pub struct Url {
     host_end: u32,
     host: HostInternal,
     port: Option<u16>,
+    port_int: Option<u16>,
     path_start: u32,             // Before initial '/', if any
     query_start: Option<u32>,    // Before '?', unlike Position::QueryStart
     fragment_start: Option<u32>, // Before '#', unlike Position::FragmentStart
@@ -502,10 +503,9 @@ impl Url {
             } else {
                 assert_eq!(self.byte_at(self.host_end), b':');
                 let port_str = self.slice(self.host_end + 1..self.path_start);
-                assert_eq!(
-                    self.port,
-                    Some(port_str.parse::<u16>().expect("Couldn't parse port?"))
-                );
+                let port = port_str.parse::<u16>().expect("Couldn't parse port?");
+                assert_eq!(self.port, Some(port));
+                assert_eq!(self.port_int, Some(port));
             }
             assert!(
                 self.path_start as usize == self.serialization.len()
@@ -518,6 +518,7 @@ impl Url {
             assert_eq!(self.host_end, self.scheme_end + 1);
             assert_eq!(self.host, HostInternal::None);
             assert_eq!(self.port, None);
+            assert_eq!(self.port_int, None);
             assert_eq!(self.path_start, self.scheme_end + 1);
         }
         if let Some(start) = self.query_start {
@@ -545,6 +546,7 @@ impl Url {
                 (self.host_str(), other.host_str()) == (None, Some(""))
         );
         assert_eq!(self.port, other.port);
+        assert!(self.port == None || self.port_int == other.port_int);
         assert_eq!(self.path_start, other.path_start);
         assert_eq!(self.query_start, other.query_start);
         assert_eq!(self.fragment_start, other.fragment_start);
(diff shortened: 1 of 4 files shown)


url/src/lib.rs
```diff
@@ -182,6 +182,7 @@ pub struct Url {
     host_end: u32,
     host: HostInternal,
     port: Option<u16>,
+    port_int: Option<u16>,
```
daxpedda, line 185: Potentially I could make an `enum` like this:
```rust
enum Port {
    None,
    Default(u16),
    Custom(u16),
}
```

url/src/parser.rs
```diff
@@ -947,12 +957,13 @@ impl<'a> Parser<'a> {
         Ok((username_end, remaining))
     }
 
+    #[allow(clippy::type_complexity)]
```
daxpedda, line 960: If use the `enum` from above I could potentially pass it down from here, reducing the complexity.

url/src/quirks.rs
```diff
@@ -141,7 +141,7 @@ pub fn set_host(url: &mut Url, new_host: &str) -> Result<(), ()> {
     if host == Host::Domain("".to_string()) {
         if !username(&url).is_empty() {
             return Err(());
-        } else if let Some(Some(_)) = opt_port {
+        } else if let Some((_, Some(_))) = opt_port {
```
daxpedda, line 144: This might be considered a breaking change? I consider checking for user input regardless of defaults "more" correct, this might have been considered a bug until now.
This can easily be removed.

Issue #609, opened 2020-06-19
no_std support?


cbeck88:
Would you be interested in patches to make the `url` library support a `no_std` build configuration based on the alloc crate instead? I think it should be possible and will make the library more easily portable.

SimonSapin:
The crate relies on several items from `std::net`.

garbageslam:
I see, thanks

garbageslam:
Are there any other blockers to this besides `std::net::{Ipv4Addr, Ipv6Addr}` and the `SocketAddr` thing? I understand that in rust std these have different representations on different OS's, and probably need to be in std. But it seems in principle we should be able to parse IPv4 and IPv6 objects without using platform specific code.

SimonSapin:
In theory we could remove DNS resolution support and use custom types to represent IP addresses. B both would be breaking changes, and upgrading the ecosystem to a semver-incompatible version of the `url` crate takes significant effort.

garbageslam:
Wow, I didn't realize it does DNS resolution support as well -- I guess I don't know the full scope of this library. Thank you!

Plecra:
@garbageslam Do you have specific features from this crate that you're interested in? I'm working on making my redesign of the `uri` crate `no_std`, and it'd be great to know what would be useful for you ^^

garbageslam:
@Plecra hi -- I think the specific features are,
- `FromStr`, with a standards conforming implementation,
- `.as_str()`,
- ways to get and set the scheme, hostname, port, path, query parameters, fragment

pranav-bhatt:
Perhaps progress can be made once [this issue](https://github.com/rust-lang/rfcs/pull/2832) is resolved?

grego:
Could the features that require `std` be placed behind a feature gate?

paolobarbolini:
I see some very interesting progress with `core::net` in 1.77 and `core::error::Error` in 1.81.

mspiegel:
Hi. I have a patch that completes the no_std conversion for the url crate. Some implementation details:
 - The Minimum Supported Rust Version remains unchanged for the "std" feature (1.56). The MSRV for no_std support is 1.81. This allows the crate to use std::* for the MSRV and use a more modern core::* to provide no_std support.
 - Several public methods of the API have been moved behind a "std" feature gate. These are: `url::Url::socket_addrs()`, `url::Url::from_file_path()`, `url::Url::from_directory_path()`, and `url::Url::to_file_path()`.

[...]

PR #717, opened 2021-08-13
`no_std` support for all crates

madsmtm:
Fixes https://github.com/servo/rust-url/issues/609.

~Our dependencies need to support `no_std`~ Done:
- `matches`, support since version `0.1.9`, see https://github.com/SimonSapin/rust-std-candidates/pull/23
- `unicode-bidi`, support since version `0.3.6`, see https://github.com/servo/unicode-bidi/pull/58
- `unicode-normalization`, support since version `0.1.13`, see https://github.com/unicode-rs/unicode-normalization/pull/55

As far as I can see there's a few ways to solve the `std::net` issue noted in the linked issue:
1. Use something like [`no-std-net`](https://github.com/dunmatt/no-std-net) on `no_std` targets. Since that library doesn't have a MSRV, we might need to define that `no_std` support doesn't guarantee a MSRV (until we find a more permanent solution).
2. Change the structure of `Host` to store custom ip address types, and return `std::net` types in functions (that can then be feature gated). Requires a breaking version.
3. Wait until the [RFC](https://github.com/rust-lang/rfcs/pull/2832) is merged, implemented and stabilized, and wait further until they have been around long enough to be in our MSRV.

As far as I can see option 1 is possible now, and it could even (with a little more work on `no-std-net` beforehand) be possible to migrate away from once (if) the RFC is stabilized, all without breaking changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,6 +42,12 @@ jobs:
         with:
           command: test
           args: --all-features
+      - uses: actions-rs/cargo@v1
+        # --no-default-features only available since 1.51.0
+        if: matrix.rust != '1.45.0'
+        with:
+          command: test
+          args: --no-default-features
 
   WASM:
     runs-on: ubuntu-latest
--- a/data-url/src/forgiving_base64.rs
+++ b/data-url/src/forgiving_base64.rs
@@ -1,5 +1,7 @@
 //! <https://infra.spec.whatwg.org/#forgiving-base64-decode>
 
+use alloc::vec::Vec;
+
 #[derive(Debug)]
 pub struct InvalidBase64(InvalidBase64Details);
 
--- a/data-url/src/lib.rs
+++ b/data-url/src/lib.rs
@@ -14,6 +14,12 @@
 //! assert_eq!(body, b"Hello World!");
 //! assert!(fragment.is_none());
 //! ```
+#![no_std]
+
+#[macro_use]
+extern crate alloc;
+
+use alloc::{string::String, vec::Vec};
 
 macro_rules! require {
     ($condition: expr) => {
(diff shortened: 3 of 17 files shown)


MathiasKoch:
Could this PR be revisited, now that all the dependencies are merged?

madsmtm:
I think option 1 described above is feasible: We'll use `no-std-net` until these parts of `net` becomes available in `core` (if they ever do). We can't guarantee an MSRV on `no_std`, because `no-std-net` should be kept up to date with the current standard library, and hence might not compile on older Rust versions.

Currently I'm waiting for `no-std-net` to become a facade to `std`, see https://github.com/dunmatt/no-std-net/pull/13, and for it to update to the implementation from newer Rust versions, see https://github.com/dunmatt/no-std-net/pull/14.

Once these are done, we'll probably need to convince `no-std-net` to not have differences to `std::net` (like their extra `const fn`s). This is required so that if `core::net` ever happens, migration to that is possible without breaking changes.

And finally, we'll need a new release with these changes.

If you can get some traction on the linked PRs, then I'll be glad to give this some more work.

MathiasKoch:
What is the current state of this? It seems to have completely stalled, even though all the blockers seems to have been merged & resolved?

madsmtm:
It has stalled mainly because I haven't received word from the maintainers on neither this nor https://github.com/servo/rust-url/pull/722.

In the meantime, the PR for changing the layout of the network primitives is [almost implemented](https://github.com/rust-lang/rust/pull/78802), after that [RFC 2832](https://github.com/rust-lang/rfcs/pull/2832) should be fairly trivial to accept.

A way forward once we get that is to use something like [`if_rust_version`](https://crates.io/crates/if_rust_version/1.0.0) to optionally import these from `core` instead of `std` (and then, once our MSRV is high enough, remove the dependency on `if_rust_version` again).

madsmtm:
@domenukk rebased this in https://github.com/servo/rust-url/pull/831, given that they seem to want to continue on this work, I'll close this PR (though I'm willing to pick it up again if you'd like my help).

domenukk:
> @domenukk rebased this in #831, given that they seem to want to continue on this work, I'll close this PR (though I'm willing to pick it up again if you'd like my help).

Happy to work together on this in any case

PR #728, opened 2021-11-16
init commit to make idna optional

### LINKED ISSUE #727: [Question] Consider make idna optional?
Background:
I am developing a rust sdk with strong binary size requirement, after introduce url crate, the binary size increased by around 400kb.
I did a quick hack, delete all code reference to idna (4 to 5 places), and compare the final binary size, confirmed idna is the cause. 

```bash
# before 
-rwxr-xr-x  1 xxxxxx staff  1143856 Nov 16 12:12 ../target/release/deps/parse_url-124867e06fe8f448

# after delete idna code
-rwxr-xr-x  1 xxxxxx  staff  591008 Nov 16 12:16 ../target/release/deps/parse_url-8091d2463bfcb288
```

shuoli84:
fix #727

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -25,10 +25,13 @@ bencher = "0.1"
 
 [dependencies]
 form_urlencoded = { version = "1.0.0", path = "../form_urlencoded" }
-idna = { version = "0.2.0", path = "../idna" }
+idna = { version = "0.2.0", path = "../idna", optional = true }
 matches = "0.1"
 percent-encoding = { version = "2.1.0", path = "../percent_encoding" }
 serde = {version = "1.0", optional = true, features = ["derive"]}
+
+[features]
+default = ["idna"]
 
 [[bench]]
 name = "parse_url"
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -170,7 +170,7 @@ impl Host<String> {
     fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
         // without idna feature, we can't verify that xn-- domains correctness
         let domain = domain.to_lowercase();
-        if domain.is_ascii() && !domain.starts_with("xn--") {
+        if domain.is_ascii() && domain.split('.').all(|s| !s.starts_with("xn--")) {
             Ok(domain)
         } else {
             Err(ParseError::InvalidDomainCharacter)
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -82,7 +82,12 @@ impl Host<String> {
             return parse_ipv6addr(&input[1..input.len() - 1]).map(Host::Ipv6);
         }
         let domain = percent_decode(input.as_bytes()).decode_utf8_lossy();
+
+        #[cfg(feature = "idna")]
         let domain = idna::domain_to_ascii(&domain)?;
+        #[cfg(not(feature = "idna"))]
+        let domain = domain.to_string();
+
         if domain.is_empty() {
             return Err(ParseError::EmptyHost);
         }
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -83,10 +83,7 @@ impl Host<String> {
         }
         let domain = percent_decode(input.as_bytes()).decode_utf8_lossy();
 
-        #[cfg(feature = "idna")]
-        let domain = idna::domain_to_ascii(&domain)?;
-        #[cfg(not(feature = "idna"))]
-        let domain = domain.to_string();
+        let domain = Self::domain_to_ascii(&domain)?;
 
         if domain.is_empty() {
             return Err(ParseError::EmptyHost);
@@ -159,6 +156,24 @@ impl Host<String> {
             Ok(Host::Domain(
                 utf8_percent_encode(input, CONTROLS).to_string(),
             ))
+        }
+    }
+
+    /// convert domain with idna
+    #[cfg(feature = "idna")]
+    fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
+        idna::domain_to_ascii(&domain).map_err(Into::into)
+    }
+
+    /// checks domain is ascii
+    #[cfg(not(feature = "idna"))]
+    fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
+        // without idna feature, we can't verify that xn-- domains correctness
+        let domain = domain.to_lowercase();
+        if domain.is_ascii() && !domain.starts_with("xn--") {
+            Ok(domain)
+        } else {
+            Err(ParseError::InvalidDomainCharacter)
         }
     }
 }
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -118,6 +118,16 @@ See [serde documentation](https://serde.rs) for more information.
 ```toml
 url = { version = "2", features = ["serde"] }
 ```
+
+# Feature: `idna`
+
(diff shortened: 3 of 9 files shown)


shuoli84:
To show the differences, run

```sh
cargo build --release --no-default-features --benches
cargo build --release --benches

# then
ls -al ../target/release/deps/parse_url-*
```

shuoli84:
@valenting take a look?

url/src/origin.rs
```diff
@@ -93,7 +94,11 @@ impl Origin {
             Origin::Tuple(ref scheme, ref host, port) => {
                 let host = match *host {
                     Host::Domain(ref domain) => {
+                        #[cfg(feature = "idna")]
                         let (domain, _errors) = domain_to_unicode(domain);
```
madsmtm, line 98: Nit: Do `idna::domain_to_unicode` and remove the `use` above

madsmtm (changes requested):
I think the feature should be documented, akin to [how `regex` does it](https://docs.rs/regex/1.5.4/regex/#crate-features).

Especially important to state the (heavy) disadvantages of opting out!

url/src/origin.rs
shuoli84: Done

shuoli84:
> I think the feature should be documented, akin to [how `regex` does it](https://docs.rs/regex/1.5.4/regex/#crate-features).
> 
> Especially important to state the (heavy) disadvantages of opting out!

@madsmtm I've add a small note on the feature. Not sure how to especially state the disadvantage. If I disable the idna, then all idna url won't parse? Maybe add some examples?

url/src/host.rs
```diff
@@ -82,7 +82,12 @@ impl Host<String> {
             return parse_ipv6addr(&input[1..input.len() - 1]).map(Host::Ipv6);
         }
         let domain = percent_decode(input.as_bytes()).decode_utf8_lossy();
+
+        #[cfg(feature = "idna")]
         let domain = idna::domain_to_ascii(&domain)?;
+        #[cfg(not(feature = "idna"))]
+        let domain = domain.to_string();
```
madsmtm, line 89: Maybe this should panic if `domain` is not ASCII? I think the rest of the code expects `Host::Domain` to hold the normalized domain name, so this way we don't introduce weird incompatibilities.

url/src/origin.rs
```diff
@@ -93,7 +92,11 @@ impl Origin {
             Origin::Tuple(ref scheme, ref host, port) => {
                 let host = match *host {
                     Host::Domain(ref domain) => {
-                        let (domain, _errors) = domain_to_unicode(domain);
+                        #[cfg(feature = "idna")]
+                        let (domain, _errors) = idna::domain_to_unicode(domain);
+                        #[cfg(not(feature = "idna"))]
+                        let domain = domain.clone();
```
madsmtm, line 98: Like with the above comment, this should maybe panic if `domain` would change after `idna::domain_to_unicode` (if we can determine that easily?); otherwise, this function would return two different things based on whether the feature is enabled or not.

url/src/lib.rs
```diff
@@ -118,6 +118,16 @@ See [serde documentation](https://serde.rs) for more information.
 ```toml
 url = { version = "2", features = ["serde"] }
 ```
+
+# Feature: `idna`
+
+You can opt out [idna](https://en.wikipedia.org/wiki/Internationalized_domain_name) support
+to reduce final binary size.
+
+```tomo
```
madsmtm, line 127: Nit
```suggestion
```toml
```

madsmtm (changes requested):
I'm not that knowledged in the actual workings of IDNA, so these are just vague suggestions.

I think it's fine that clients whom know the domains they're going to be accessing, disable the feature. E.g. a resource-constrained device that makes HTTP requests to a server under the developer's control.

But I worry that e.g. some servers will disable the `idna` feature without really thinking further, and then their domain matching code will fail in weird (possibly security-relevant) ways.

url/src/host.rs
valenting: Maybe not panic, but return an error?
Not only if the domain is not ASCII, but also if the domain contains any `xn--` labels - which we can't know if they are correct or no.

url/Cargo.toml
```diff
@@ -25,11 +25,14 @@ bencher = "0.1"
...
 [dependencies]
 form_urlencoded = { version = "1.0.0", path = "../form_urlencoded" }
-idna = { version = "0.2.0", path = "../idna" }
+idna = { version = "0.2.0", path = "../idna", optional = true }
 matches = "0.1"
 percent-encoding = { version = "2.1.0", path = "../percent_encoding" }
 serde = {version = "1.0", optional = true, features = ["derive"]}
 
+[features]
+default = ["idna"]
```
valenting, line 34: I wonder if we should reverse this and name the feature "disable-idna" or "unsafe-disable-idna" ?

url/src/host.rs
shuoli84: I happen to be a backend engineer also, panic is the last thing I expect a base lib to do, especially when it may handle external input. I disable the idna doesn't mean I want my service panic for idna input...

url/src/origin.rs
shuoli84: If idna feature disabled, then the whole fn unicode_serialization can be opt out? I searched spec about unicode_serialization, it is strictly related with idna.

url/src/host.rs
shuoli84: panic should used when the program enter some bad state and not easy to recover. For this case, it has no problem to recover.

url/src/origin.rs
shuoli84: https://html.spec.whatwg.org/multipage/origin.html#unicode-serialisation-of-an-origin now the spec even deletes the section of unicode serialization, just saying "There used to also be a Unicode serialization of an origin. However, it was never widely adopted."

url/src/lib.rs
shuoli84: fixed

shuoli84:
@madsmtm @valenting ping~

shuoli84:
@valenting ping

shuoli84:
Some Lint reports error on some changes not touched by this pr. Do I need to make all green?

url/src/host.rs
```diff
@@ -156,6 +158,24 @@ impl Host<String> {
...
+    fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
+        idna::domain_to_ascii(&domain).map_err(Into::into)
+    }
+
+    /// checks domain is ascii
+    #[cfg(not(feature = "idna"))]
+    fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
+        // without idna feature, we can't verify that xn-- domains correctness
+        let domain = domain.to_lowercase();
+        if domain.is_ascii() && !domain.starts_with("xn--") {
```
valenting, line 173: We also need to deal with sublabels here.
```suggestion
        if domain.is_ascii() && domain.split('.').all(|s| !s.starts_with("xn--")) {
```

valenting (changes requested):
Looks good otherwise. I'll deal with the other lint failures in a separate PR.

url/src/host.rs
shuoli84: good catch

shuoli84:
@valenting updated~

Issue #730, opened 2021-11-23
Support for converting between unix-style `Path` and `Url` on Windows


dsherret:
Often in tests I'll use only unix style paths (ex. `/data/test.txt`) because it works with `Path` on both Windows and Unix.

The problem I have though, is the `Url::from_file_path(path)` and `url.to_file_path()` both error on Windows for these kind of paths and specifiers in this crate even though they can be represented just fine as a `std::path::Path`. For example:

```rust
let path = PathBuf::from("/testing");
// Unix: Ok(_)
// Windows: Err(())
println!("{:?}", Url::from_file_path(&path));

let url = Url::parse("file:///testing").unwrap();
// Unix: Ok("/testing")
// Windows: Err(())
println!("{:?}", url.to_file_path());
```

Is there a reason these aren't supported on Windows? Would you accept a PR that adds support for this?

valenting:
The problem is that unix style paths don't work on Windows as is.
While `/testing` is valid on Linux, on windows you need a drive letter - `C:\testing` or `D:\testing`
I'm not sure how you'd go about converting between these.

dsherret:
I feel like in the spirit of `Path` it might be nice for it to do the conversion and then if someone uses the path with the actual file system then at that point it will error. For example, `PathBuf::from("/testing")` is valid on Windows in Rust (as I guess mostly anything is), but then it errors when the file system itself says it's not a valid path. This is really useful because then these paths can be used with in memory file systems on Windows or for only bothering with unix style paths in tests where the file system is mocked.

valenting:
Now I understand. I completely agree that `file:///testing` should parse on Windows too.
I'm happy to take a PR for it.

andrewbanchich:
That would be great @valenting , thanks! We actually just ran into this as well, but for code used outside of tests.

Hulxv:
I think this issue should be closed

andrewbanchich:
@Hulxv why?

Hulxv:
Didn't it be solved in #732 ?

andrewbanchich:
Nope

https://github.com/servo/rust-url/pull/732#issuecomment-991019365

It would say merged instead of closed

PR #732, opened 2021-12-05
Cross-platform parsing

andrewbanchich:
@valenting I'd like to take #730 but would appreciate some feedback on the implementation details:

1. What should `file:///foo/bar` convert to on Windows? I'm currently converting it to `\foo\bar`, but I'm not sure if that's a good idea since that's a Windows drive-relative path, which isn't what the URL was intended to be interpreted as since the URL is missing the drive letter.
2. I moved this [to a conditional](https://github.com/andrewbanchich/rust-url/blob/39a16680e50c9b092c324751296a6c0ff7152072/url/src/lib.rs#L2852) since `\foo\bar` is not absolute.

```rust
    debug_assert!(
        path.is_absolute(),
        "to_file_path() failed to produce an absolute Path"
    );
```

However, I'm not sure if we want to even keep this assertion anymore.

3. I renamed some variables and used `PathBuf` instead of `String` for clarity.
4. The first `match` arm looks like it's checking for a drive letter, but I wasn't sure what [the second arm](https://github.com/servo/rust-url/blob/master/url/src/lib.rs#L2834) is checking for. Could you clarify that please?
5. I fixed what I think was a typo in [this test](https://github.com/servo/rust-url/blob/master/url/tests/unit.rs#L88). Still working on getting that one to pass, so I commented it out for now.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -2708,15 +2708,14 @@ fn path_to_file_url_segments_windows(
     serialization: &mut String,
 ) -> Result<(u32, HostInternal), ()> {
     use std::path::{Component, Prefix};
-    if !path.is_absolute() {
-        return Err(());
-    }
+
     let mut components = path.components();
 
     let host_start = serialization.len() + 1;
     let host_end;
     let host_internal;
-    match components.next() {
+
+    match dbg!(components.next()) {
         Some(Component::Prefix(ref p)) => match p.kind() {
             Prefix::Disk(letter) | Prefix::VerbatimDisk(letter) => {
                 host_end = to_u32(serialization.len()).unwrap();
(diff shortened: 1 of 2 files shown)


valenting:
>     1. What should `file:///foo/bar` convert to on Windows? I'm currently converting it to `\foo\bar`, but I'm not sure if that's a good idea since that's a Windows drive-relative path, which isn't what the URL was intended to be interpreted as since the URL is missing the drive letter.

Do we actually need the call to succeed? `\foo\bar` is not a valid windows path, so we could just return an error in my opinion.

>     4. The first `match` arm looks like it's checking for a drive letter, but I wasn't sure what [the second arm](https://github.com/servo/rust-url/blob/master/url/src/lib.rs#L2834) is checking for. Could you clarify that please?

It's checking for a percent encoded `:`. %3a or %3A.

>     5. I fixed what I think was a typo in [this test](https://github.com/servo/rust-url/blob/master/url/tests/unit.rs#L88). Still working on getting that one to pass, so I commented it out for now.

I'm having second thoughts about making `Url::from_file_path(Path::new(r"\drive-relative"))` work.
Relative paths should not really be convertible to a Url, since the Url is absolute. This could introduce actual security bugs. I think we should focus on getting `Url::parse("file:///testing")` working on windows.

Alternatively, we could make `file:///testing/test` convert to `/testing/test` on windows too (note no backslashes). Though this might have similar unexpected downsides.

andrewbanchich:
> Do we actually need the call to succeed? \foo\bar is not a valid windows path, so we could just return an error in my opinion.

Isn't that a drive-relative path?

andrewbanchich:
> we should focus on getting Url::parse("file:///testing") working on windows.

I think this has always worked on Windows. I just tried it again and confirmed.

> I'm having second thoughts about making Url::from_file_path(Path::new(r"\drive-relative")) work.

Would the security issue be introduced with `from_file_path` or `to_file_path`? I've updated the PR to only succeed in parsing the URL, so `Url::from_file_path(Path::new(r"\foo\bar"))` would parse to `file:///foo/bar`, but trying to convert that URL to a path again would fail.

url/src/lib.rs
```diff
@@ -2734,30 +2733,48 @@ fn path_to_file_url_segments_windows(
...
         },
-
-        _ => return Err(()),
+        Some(Component::Normal(_)) | Some(Component::ParentDir) => return Err(()),
+        _ => {
+            host_end = to_u32(serialization.len()).unwrap();
+            host_internal = HostInternal::None;
+            // serialization.push('/');
+            // dbg!(&host_end, &serialization);
+            // return Err(());
```
valenting, line 2752: Is the commented text left on purpose?
Could you make sure this branch is covered by a unit test?

url/tests/unit.rs
```diff
@@ -847,6 +860,19 @@ fn test_url_from_file_path() {
...
+#[cfg(windows)]
+#[test]
+fn test_url_from_invalid_file_path() {
+    use std::path::PathBuf;
+    use url::Url;
+
+    assert_eq!(
+        Url::from_file_path(PathBuf::from(r"\foo\bar")).unwrap(),
+        Url::parse("file:///foo/bar").unwrap()
+    );
```
valenting, line 873: Could we add a test that to_file_path fails for this and the `\\server` case?

valenting (changes requested):

url/tests/unit.rs
```diff
@@ -84,8 +84,16 @@ fn new_file_paths() {
...
-        assert_eq!(Url::from_file_path(Path::new(r"\\ucn\")), Err(()));
+        assert_eq!(
+            Url::from_file_path(Path::new(r"\drive-relative")).unwrap(),
+            Url::parse("file:///drive-relative").unwrap()
+        );
+
+        // UNC
+        assert_eq!(
+            Url::from_file_path(Path::new(r"\\server\")).unwrap(),
+            Url::parse("file:///server").unwrap()
```
valenting, line 95: Could you explain again why relative paths would fail to parse, but drive relative paths would not?
In my mind they're not any different.

url/src/lib.rs
andrewbanchich: Yes, this is still a draft PR - I'm just getting an idea of what we expect the new behavior to be and then once I'm done I'll mark as ready for review.

url/tests/unit.rs
andrewbanchich: I agree, I just didn't update this part yet (still a draft PR).

andrewbanchich:
I'm going to close this for now and create a new PR (hopefully soon) after I organize my thoughts on this a bit more.

PR #734, opened 2021-12-10
Fix setting file hosts to `None`

### LINKED ISSUE #731: `set_host` not working
In [the example](https://docs.rs/url/2.2.2/url/struct.Url.html#method.set_host), after calling `url.set_host(None)` the host is actually not set to `None`, but is `Some("")`.

[playground demo](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd9fddebf1aee0e6ffb947eb0ff7997e)

Not sure if this is related, but I also cannot seem to set host to `None` for certain types of URLs ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9e3735eaffb59424709d0a4947cede2b)). The error just says `EmptyHost` but I can't find any explanation of what this is supposed to mean.

andrewbanchich:
Fixes #731 for file URLs.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -1813,8 +1813,10 @@ impl Url {
             return Err(ParseError::SetHostOnCannotBeABaseUrl);
         }
 
+        let scheme_type = SchemeType::from(self.scheme());
+
         if let Some(host) = host {
-            if host.is_empty() && SchemeType::from(self.scheme()).is_special() {
+            if host.is_empty() && scheme_type.is_special() && !scheme_type.is_file() {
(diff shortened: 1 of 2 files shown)


andrewbanchich:
Any thoughts on this @valenting ?

valenting:
Hi, Sorry I missed this one.
I think the fix looks good. Could you rebase so we get the CI running?
Thanks!

andrewbanchich:
all set @valenting

valenting (approved):
Thanks!

PR #735, opened 2021-12-22
Add examples to `set_path()`

annmarie-switzer:
Adds examples to the `set_path` documentation to clarify how the function handles percent-encoding.


valenting:
Thank you!

PR #738, opened 2022-01-10
Add file related methods for wasi

HKalbasi:
These methods where removed on wasm, but they are ok with wasi and should work exactly like unix if I understand correctly. So this PR adds them.


HKalbasi:
I think CI failure is not related to this PR.

PR #744, opened 2022-01-28
Appease clippy

lucacasonato:
This commit appeases clippy by fixing all of the issues it flagged.

Alternative to #740

cc @valenting


lucacasonato:
@valenting I think this is ready to land. The coverage and audit issues will be resolved by #746. I will open a follow up PR to resolve this.

valenting:
Thank you! Great work!

Issue #745, opened 2022-01-28
Updating MSRV to 1.45.0


lucacasonato:
We want to soon update the MSRV (minimum supported Rust version) for `rust-url` to 1.45.0. This is done so that these crates can use various new standard library features. The current MSRV is 1.36.0.

If you are a consumer of `rust-url` and this is a problem for you, please comment on this issue.

PR #748, opened 2022-01-28
Use `tester` instead of `rustc-test`

### LINKED ISSUE #746: Remove `rustc-test` dev dependency
This dependency is causing some trouble:

a) it has not been updated in a long time
b) it depends on old versions of `term` (which depends on an old version of `winapi`)
c) it depends on `term`: `term` is unmaintained
d) it depends on a very old version of `time`
e) id depends on `rustc-serialize` which is deprecated. `serde` should be used instead.

lucacasonato:
This commit switches the test library used by idna/tests and
data-url/tests to `tester`. Just like `rustc-test` this is a fork of
the `test` crate inside of `rustc`, but it is much more up to date.

`cargo deny` and `cargo tarpaulin` are both now not complaining anymore.

Closes #746

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -2,7 +2,7 @@ name: Coverage
 
 on:
   push:
-    branches: ['master']
+    branches: ["master"]
   pull_request:
 
 jobs:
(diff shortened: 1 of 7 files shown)


lucacasonato:
Blocked on #745

madsmtm (approved):

PR #750, opened 2022-01-28
Update Rust MSRV to 1.45

### LINKED ISSUE #745: Updating MSRV to 1.45.0

lucacasonato:
The earliest Rust version with all of the standard library features we'd like to make use of is 1.45. The only feature this does not have yet is `edition = 2021`, but other than looking nice, that would have no effect on `rust-url` anyway.

I recommend we bump the MSRV to 1.45 instead of 1.56 then, to minimize breakage across the ecosystem.

Closes #745

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2016 The rust-url developers
+Copyright (c) 2013-2022 The rust-url developers
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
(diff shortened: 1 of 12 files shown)


LICENSE-MIT
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2016 The rust-url developers
+Copyright (c) 2013-2021 The rust-url developers
```
madsmtm, line 1: *2022

madsmtm (approved):

valenting (approved):

Issue #742, opened 2022-01-28
Specification alignment tracking issue


crowlKats:
updating wpt testdata makes following nonalignment evident:

- [ ] https://github.com/web-platform-tests/wpt/pull/32539
- [ ] https://github.com/whatwg/url/pull/589
- [ ] https://github.com/web-platform-tests/wpt/pull/25987
- [ ] https://github.com/whatwg/url/pull/544 (https://github.com/servo/rust-url/issues/642)
- [ ] https://github.com/web-platform-tests/wpt/pull/28432
- [ ] https://github.com/whatwg/url/pull/550
- [ ] https://github.com/whatwg/url/pull/505 (https://github.com/servo/rust-url/issues/601)
- [ ] https://github.com/whatwg/url/issues/539 (just testdata runner failure)
- [ ] https://github.com/whatwg/url/pull/619

lucacasonato:
For some background on what this issue is:

We (the Deno project) use `rust-url` internally extensively. The crate is currently about 1.5 years out of sync with the spec. We are going to try to get it back up to 100% PASS in the next couple of weeks. The tenative plan of action is as follows:

1. Update the WPT testdata file in this repo with all of the new tests that still pass (#743).
2. Figure out which tests are blocked by which spec changes (we'll add that analysis to this issue).
3. Implement the relevant spec changes and add the corresponding tests. In some cases this may be as simple as updating the WPT test runner.

PR #753, opened 2022-01-28
Align IPv4 parsing to spec

lucacasonato:
This commit aligns the IPv4 parsing steps to spec and adds the relevant
WPT tests.

Relevant specification change: https://github.com/whatwg/url/commit/ab0e820b0b559610b30c731b7f2c1a8094181680

Towards #742

### DIFF
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -109,7 +109,8 @@ impl Host<String> {
 
         if domain.find(is_invalid_domain_char).is_some() {
             Err(ParseError::InvalidDomainCharacter)
-        } else if let Some(address) = parse_ipv4addr(&domain)? {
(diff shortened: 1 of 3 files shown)


url/src/host.rs
```diff
@@ -90,29 +91,26 @@ impl Host<String> {
...
-                    | '>'
-                    | '?'
-                    | '@'
-                    | '['
-                    | '\\'
-                    | ']'
-                    | '^'
-                    | '\u{007F}'
-            )
+            matches!(c, |'\0'..='\u{001F}'| ' '
```
lucacasonato, line 94: This is an unrelated `cargo fmt` change that slipped through a previous PR. Whoops!

PR #754, opened 2022-02-09
fix: make_relative missing filename when filenames equal in both urls

dsherret:
```rust
let from_path = url::Url:::parse("http://127.0.0.1:8080/file.txt").unwrap();
let to_path = url::Url:::parse("http://127.0.0.1:8080/test/file.txt").unwrap();
println!("{}", from_path.make_relative(&to_path).unwrap());
```

Expected: `"test/file.txt"`
Actual: `"test"`

Not sure if this expected is correct or if the solution is right, but it passes the tests.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -470,7 +470,7 @@ impl Url {
         }
 
         // Add the filename if they are not the same
-        if base_filename != url_filename {
+        if !relative.is_empty() || base_filename != url_filename {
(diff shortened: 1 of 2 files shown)


jgraettinger:
We hit this bug too. The current behavior is incorrect, and this PR LGTM for fixing it. Can it be merged?

valenting:
Thanks for the ping. This looks good to me.

PR #762, opened 2022-04-02
feat(url): InternalComponents

AaronO:


valenting (changes requested):
Could you add some documentation for the new struct and method?
Please explain why this is preferable to using the serde feature?

PR #768, opened 2022-05-13
Add a `set_scheme_unchecked` method that has less checks

expenses:
The current `set_scheme` function has the following limitation:

> Do nothing and return `Err` under the following circumstances:
> * If either the old or new scheme is `http`, `https`, `ws`,
> `wss` or `ftp` and the other is not one of these

I'm sure this is well meaning and conforms to the spec and etc. etc., but it's annoying for cases where you want to be able to do this anyway. My use case is changing the scheme of `ipfs://` urls to `http://` so I can use the [Browser Cache API](https://developer.mozilla.org/en-US/docs/Web/API/Cache) with them. This is a little devious but still something you should be able to do easily, I think.

It still has the following limitation:

> Do nothing and return `Err` under the following circumstances:
>
> * If the new scheme is not in `[a-zA-Z][a-zA-Z0-9+.-]+`

I could remove this as well but that would require creating a different version of the parser.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -2287,6 +2287,146 @@ impl Url {
         Ok(())
     }
 
+    /// Change this URL’s scheme with less checks.
+    ///
+    /// Do nothing and return `Err` under the following circumstances:
+    ///
+    /// * If the new scheme is not in `[a-zA-Z][a-zA-Z0-9+.-]+`
+    ///
+    /// See also [the URL specification's section on legal scheme state
+    /// overrides](https://url.spec.whatwg.org/#scheme-state).
+    ///
+    /// # Examples
+    ///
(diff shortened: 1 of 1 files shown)


expenses:
I'm going to close this because it's a bit of a hack and it's better to just build the url up from scratch.

sdroege:
I have a similar use-case to @ensc here. Simply being able to replace the scheme instead of having to reconstruct the URI from its components seems less error-prone.

I could open a new PR for this if this has any chance of getting merged. CC @valenting

domenkozar:
Same here, Nix uses `type+protocol://`` scheme and we need to be able to remove `type+` out of it, at the moment we have to convert url to string, do the removal and then convert it back.

PR #776, opened 2022-06-17
SFTP support

mihaigalos:
This PR contributes parsing of `sftp` URLs (port `22`).


tmccombs:
Sftp isn't listed as a special scheme in the whatwg standard is it?

This is probably another example where it would be good to have an API to tell rust_url to treat additional schemes as "special".

mihaigalos:
True. I see some URL specs do include it, see [this](https://www.ibm.com/docs/en/datapower-gateway/7.5.0?topic=handlers-url-specifications-sftp-handler).

mihaigalos:
Hi, any updates here?

valenting:
I don't think we want to merge this at this time.
This crate implements the [URL standard](https://url.spec.whatwg.org/) - which does not special case sftp.
There are a lot of other protocols which fall in the same bucket as sftp - ssh, ftps, etc - which we also don't handle specifically.
Thanks!

mihaigalos:
Ok. I think I'll start using [`rs-baseurl`](https://github.com/BradyMcd/rs-baseurl) or republish a hard forked crate. Thanks.

PR #783, opened 2022-08-03
Add Natvis visualizations for the `Url` type

ridwanabdillahi:
This change adds Natvis visualizations for the `Url` type in the `url` crate to help improve the debugging experience on Windows.

Debugging types such as the Url type under a Windows debugger, WinDbg or the Visual Studio debugger, can be a bit difficult to dissect exactly what data is contained in each type. The Rust compiler does have Natvis support for some types, but this is limited to some of the core libraries and not supported for external crates.

[RFC 3191](https://github.com/rust-lang/rfcs/pull/3191) proposes adding support for embedding debugging visualizations such as Natvis in a Rust crate. This RFC has been approved, merged and implemented.

Natvis is a framework that can be used to specify how types should be viewed under a supported debugger, such as the Windows debugger (WinDbg) and the Visual Studio debugger.

This PR adds:

* Natvis visualizations for the `Url` type in the `url` module.
* Tests for testing visualizers embedded in the url crate.
* Updates to the CI pipeline to ensure tests for visualizers are run so they do not break silently.
* Cleanups to the CI pipeline to manually select features to enable for testing, in addition to the default features. This allows unstable features to be enabled and tested separately from stable features.
* A new debugger_visualizer feature for the url crate to enable the unstable debugger_visualizer feature.

### DIFF
--- a/debug_metadata/url.natvis
+++ b/debug_metadata/url.natvis
@@ -11,23 +11,23 @@
         <!-- Add 3 to the scheme end to account for the scheme separator which is '://' -->
         <DisplayString>{(char*)(ptr()+(scheme_end + 3)),[((username_end)-(scheme_end + 3))]s8}</DisplayString>
       </Synthetic>
-      <Synthetic Name="[host]" Condition="host.discriminant != 0">
+      <Synthetic Name="[host]" Condition="host.tag != 0">
         <DisplayString>{(char*)(ptr()+host_start),[host_end-host_start]s8}</DisplayString>
       </Synthetic>
-      <Synthetic Name="[port]" Condition="port.discriminant == 1">
-        <DisplayString>{port.variant1.__0,d}</DisplayString>
+      <Synthetic Name="[port]" Condition="port.tag == 1">
+        <DisplayString>{port.variant1.value.__0,d}</DisplayString>
       </Synthetic>
       <Synthetic Name="[path]">
-        <DisplayString Condition="query_start.discriminant == 0 &amp;&amp; fragment_start.discriminant == 0">{(char*)(ptr()+path_start),[(serialization.vec.len-path_start)]s8}</DisplayString>
-        <DisplayString Condition="query_start.discriminant == 1">{(char*)(ptr()+path_start),[(query_start.variant1.__0-path_start)]s8}</DisplayString>
-        <DisplayString Condition="fragment_start.discriminant == 1">{(char*)(ptr()+path_start),[(fragment_start.variant1.__0-path_start)]s8}</DisplayString>
+        <DisplayString Condition="query_start.tag == 0 &amp;&amp; fragment_start.tag == 0">{(char*)(ptr()+path_start),[(serialization.vec.len-path_start)]s8}</DisplayString>
+        <DisplayString Condition="query_start.tag == 1">{(char*)(ptr()+path_start),[(query_start.variant1.value.__0-path_start)]s8}</DisplayString>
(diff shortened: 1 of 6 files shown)


debug_metadata/url.natvis
```diff
@@ -0,0 +1,34 @@
...
+      <Synthetic Name="[query]" Condition="query_start.discriminant == 1">
+        <DisplayString Condition="fragment_start.discriminant == 0">{(char*)(ptr()+query_start.variant1.__0+1),[((serialization.vec.len)-(query_start.variant1.__0+1))]s8}</DisplayString>
+        <DisplayString Condition="fragment_start.discriminant == 1">{(char*)(ptr()+query_start.variant1.__0+1),[((fragment_start.variant1.__0)-(query_start.variant1.__0+1))]s8}</DisplayString>
+      </Synthetic>
+      <Synthetic Name="[fragment]" Condition="fragment_start.discriminant == 1">
+        <DisplayString>{(char*)(ptr()+fragment_start.variant1.__0+1),[(serialization.vec.len-fragment_start.variant1.__0-1)]s8}</DisplayString>
+      </Synthetic>
+    </Expand>
+  </Type>
+</AutoVisualizer>
```
valenting, line 34: Add new line.

valenting (changes requested):
This is great stuff, thank you. Please fix the warnings if possible.

ridwanabdillahi:
> This is great stuff, thank you. Please fix the warnings if possible.

Great, I'll fix the warnings and update the PR. Thanks for the review!

valenting (changes requested):
Please fix the `CI / Test (ubuntu-latest, 1.45.0)` and cargo fmt lint.
Thanks!

ridwanabdillahi:
> Please fix the `CI / Test (ubuntu-latest, 1.45.0)` and cargo fmt lint.
> Thanks!

Sorry about the lint failures.

Looking at the CI/Test failure, this is actually caused by a new version of the serde_derive crate that was published and was not caused by this change. The version 1.0.143 uses a Cargo toml feature which was not stable in the version of Cargo used by the Rust 1.45.0 toolchain.

To fix this error, this crate would need to explicitly specify that only the versions before 1.0.143 of this crate are allowed to be used. Should I go ahead and make that as a separate change now?

valenting:
> One solution would be to explicitly depend on version 1.0.142 or lower of these crates but I'm not sure if that's what we want to do here. Thoughts?

How about we don't run the serde CI task on 1.45.0?

ridwanabdillahi:
> How about we don't run the serde CI task on 1.45.0?

From what I can tell there isn't a specific serde CI task. If you mean disabling the entire 1.45.0 CI job then that should probably be a separate change from this PR with a relevant issue opened and linked.

I'm not sure if the url crate has a minimum supported Rust version (MSRV) that's dependent on the 1.45.0 Rust toolchain but the crate maintainers would probably be better served to answer that question.

ridwanabdillahi:
@valenting

I opened up an issue for this CI failure https://github.com/servo/rust-url/issues/785 and a corresponding PR https://github.com/servo/rust-url/pull/786 to update the minimum rust-version to 1.51.0 which is when the `resolver` feature was stabilized

valenting:
It seems the test failed on Windows.

ridwanabdillahi:
> It seems the test failed on Windows.

The latest nightly compiler contained changes to the debuginfo type layout for enums. The Natvis definitions have been updated and the test is passing.

PR #790, opened 2022-09-07
Make IDNA dependency non-optional

crowlKats:


valenting:
This commit reverts #728 that made the IDNA dependency optional to reduce wasm file size.
However, users of the url crate, that specify `default-features = false` would suddenly stop parsing unicode domains.
Let's remove this for now, and we can try to add it back as a `#[cfg(feature = "disable-idna")]`. Not sure if the idna crate not being optional still adds to the wasm file size, or if it gets trimmed at link time.

valenting (approved):

PR #794, opened 2022-09-17
IDNA: Errors - Allow people to get specific error in basic values

Arisa-Snowbell:
When for example I wanted to figure out whats wrong with the unicode string I tried to convert to ASCII I could before this get only string from of it becuase of the implemented Debug for Errors struct but thats very inconvinient trying to parse a string where there is the possibility of accessing the basic values themselves. This commit allows people to actually get the basic values and based on that decide what to do next in their code.

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -598,18 +598,18 @@ fn is_bidi_domain(s: &str) -> bool {
 /// More details may be exposed in the future.
 #[derive(Default)]
 pub struct Errors {
-    punycode: bool,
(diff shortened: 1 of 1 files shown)


hsivonen:
> This commit allows people to actually get the basic values and based on that decide what to do next in their code.

What's the use case for this?

In `idna` 1.0, I removed all these error categorization fields, but there's an option to have each error flagged by a REPLACEMENT CHARACTER in the ToUnicode output, which can indicate to the end user where the errors were.

PR #797, opened 2022-10-02
Fix non-base64 data URLs with % characters not followed by hex digits

### LINKED ISSUE #795: Corrupted data URL output from <rect style='fill:%23000000;' width='100%' height='100%'/>
# Version: 
data-url 0.2.0 & master

# Description
The following data URL:
```
data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'><rect style='fill:%23000000;' width='100%' height='100%'/></svg>
```
produces a corrupted result with duplicate fields and unexpected values with `DataUrl::process`:
```
<svg xmlns='http://www.w3.org/2000/svg'><rect style='fill:#000000;' width='100000000;' width='100%' height='100000000;' width='100%' height='100%'/></svg>
```

Note that the data URL produces a full window black rectangle on Firefox and keeps the initial SVG intact.

# SSCCE
```
use data_url::DataUrl;    
    
static SVG: &str = r#"data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'><rect style='fill:%23000000;' width='100%' height='100%'/></svg>"#;    
    
fn main() {    
    println!("{}", SVG);    
    let url = DataUrl::process(SVG).unwrap();    
    let (body, _) = url.decode_to_vec().unwrap();    
    let utf8 = std::str::from_utf8(&body).unwrap();    
    println!("{}", utf8);                                                                                                                                                     
} 
```

SmaugPool:
When writing accumulated "non-special" characters, `slice_start` must be updated as some later conditionals/pattern matches don't update it like the case when `%` is not followed by 2 hex digits.

This fixes #795


SmaugPool:
See https://github.com/servo/rust-url/issues/795#issuecomment-1264639264 for a more detailed explanation.

All tests pass:
```
test result: ok. 65 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.72s
```

I will run more tests but some review is needed as I'm not familiar enough with the source code to be certain that does not cause regressions.

I guess an alternative could be to update `slice_start` in the empty `else` when `i > slice_start`.

tmccombs (approved):

lucacasonato:
@SmaugPool Sorry for the late reply. Could you add a test for this?

SmaugPool:
Sure @lucacasonato, I have added 3 tests.

SmaugPool:
@lucacasonato any plan to merge this? It has been a while now (PR from 2022) and it fixes a significant issue with data URLs.

valenting (approved):
Thank you for the fix!

PR #802, opened 2022-11-04
add `description` method to `ParseError`

### LINKED ISSUE #801: Add `ParseError.description() -> &'static str` method like already implemented on `SyntaxViolation`
Would be great, so we don't need to allocate a string for something which is fundamentally static.

samuelcolvin:
fix #801

### DIFF
--- a/url/src/parser.rs
+++ b/url/src/parser.rs
@@ -60,6 +60,16 @@ macro_rules! simple_enum_error {
             )+
         }
 
+        impl ParseError {
+            pub fn description(&self) -> &'static str {
+                match *self {
+                    $(
+                        ParseError::$name => $description,
+                    )+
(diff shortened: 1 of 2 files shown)


samuelcolvin:
See https://github.com/pydantic/pydantic-core/pull/324 and in particular [here](https://github.com/pydantic/pydantic-core/blob/20fe6337cc50690e629334ff9e639cbca2a72f34/src/errors/types.rs#L325-L328) where I'd like to use the static str, rather than a `String`.

samuelcolvin:
Hi, the linting failures are not related to this PR, would you like me to fix them here?

valenting:
Could you rebase your patch so we can pick up the latest changes?

PR #809, opened 2023-02-03
Update Actions

GiddyGoatGaming:
Some actions in the workflow are pretty outdated such as checkout v1 being used in some areas which has been outdated since dec 2019. This updates all actions that have available updates in the workflow folder


lucacasonato:
Thank you, but this was done in https://github.com/servo/rust-url/pull/809 now

PR #814, opened 2023-02-25
percent_encoding: faster percent_encode_byte

klensy:
Read encode table as `&[u8; 768]` instead of `&str`, so we statically know offsets and can skip bound checks, making it branchless.

https://rust.godbolt.org/z/r8bo3M9TP


klensy:
Looking at function size, it can be marked as `inline`?

valenting (approved):

PR #816, opened 2023-02-27
Trim whitespaces only while setting scheme

### LINKED ISSUE #815: set scheme with non-tab/newline C0 controls should result in no-op
currently, we're trimming C0 controls and also whitespaces; however, leading/trailing C0 controls are invalid for setting scheme.

following tests are related wpt tests:

https://github.com/web-platform-tests/wpt/blob/f72c88ab79843a46ffdf63264e833121811f08d9/url/resources/setters_tests.json#L291-L323

CYBAI:
When we use the `parser::Input::new`, it will initialize the input iterator with trimming both whitespaces and C0 controls. However, for scheme, we should only trim whitespaces. Thus, with moving from `::new` to `::trim_tab_and_newlines` could fix the issue.

This PR will fix #815.


url/tests/unit.rs
```diff
@@ -147,6 +147,17 @@ fn new_path_windows_fun() {
     }
 }
 
+#[test]
+fn set_scheme_with_leading_or_trailing_c0_controls_should_result_in_noop() {
+    let mut url: Url = "http://test".parse().unwrap();
+
+    let controls = ["\u{0000}", "\u{000C}", "\u{000E}", "\u{0020}"];
+    for control in controls.iter() {
+        assert!(url.set_scheme(&format!("{}https", control)).is_err());
```
valenting, line 156: The web-platform-test says `Non-tab/newline C0 controls result in no-op` - but this returns an error. Is that intentional?
Could you update [urltestdata.json](https://github.com/servo/rust-url/blob/master/url/tests/urltestdata.json) with the testcases you listed [here](https://github.com/web-platform-tests/wpt/blob/f72c88ab79843a46ffdf63264e833121811f08d9/url/resources/setters_tests.json#L291-L323) ?

valenting (changes requested):

url/tests/unit.rs
CYBAI: > The web-platform-test says `Non-tab/newline C0 controls result in no-op` - but this returns an error. Is that intentional?

Sorry if I misunderstood but I thought `result in no-op` means the url scheme will not be updated but the operation of `set_scheme` failed due to invalid input for scheme so it returns an error here. Or, would it be better to return `Ok` here? 👀 

> Could you update [urltestdata.json](https://github.com/servo/rust-url/blob/master/url/tests/urltestdata.json) with the testcases you listed [here](https://github.com/web-platform-tests/wpt/blob/f72c88ab79843a46ffdf63264e833121811f08d9/url/resources/setters_tests.json#L291-L323) ?

Yes! I will update the urltestdata.json file later 🙏

CYBAI: As the tests referred in the issue in under `setters_tests.json`, I just sent a separate PR to update the test cases at #820.

CYBAI: After thinking more about the test, I think I was wrong because if it should return an error, the test would test `failure: true` instead. So, I just updated the PR to make it `no-op` 🙏 Please help me review again, thanks for catching it!

url/src/lib.rs
```diff
@@ -2313,6 +2313,12 @@ impl Url {
     /// ```
     #[allow(clippy::result_unit_err, clippy::suspicious_operation_groupings)]
     pub fn set_scheme(&mut self, scheme: &str) -> Result<(), ()> {
+        // If the given scheme contains leading or trailing C0 controls,
+        // we'll ignore the set_scheme operation.
+        if scheme.trim_matches(|ch| ch <= ' ').len() != scheme.len() {
```
valenting, line 2318: There are two problems here:
1. I think we should early exit if the scheme `contains` a Non-tab/newline C0 controls - not just leading or trailing
2. This would still match tabs and newlines.

Please add a test for one of these chars in the middle of the scheme.
Also a test that setting it to `https\n` still works.

valenting (changes requested):

Issue #799, opened 2022-10-26
Unable to encode + decode URL


djc:
This code fails:

```rust
        let url = dbg!(Url::parse("         m:/          /%2E.//     \\           ").unwrap());
        let encoded = url.as_str();
        let reparsed = Url::parse(encoded).unwrap();
```

```rust
panicked at 'called `Result::unwrap()` on an `Err` value: InvalidDomainCharacter', test.rs:3:44
```

```rust
[test.rs:1] Url::parse("         m:/          /%2E.//     \\           ").unwrap() = Url {
    scheme: "m",
    cannot_be_a_base: false,
    username: "",
    password: None,
    host: None,
    port: None,
    path: "//%20%20%20%20%20\\",
    query: None,
    fragment: None,
}
```

Found as part of fuzzing the trust-dns crate (which uses `Url` in CAA records).

qsantos:
The offending URL can be reduced to `m:/.//\\`.

qsantos:
The issue comes from the [remove dot segments](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.4) step. Let's consider the URI `m:/.//`. Then, according to the RFC:

> B.  if the input buffer begins with a prefix of "/./" […], then replace that prefix with "/"

So it should be normalized to `m://`, but this has different semantics (resulting in `\` being interpreted as being part of the authority in the original example).

I have conducted a few tests with some URI normalization libraries:

```
$ node test.js
http://example.com/a/c/d
m:/.//
$ cat test.php
<?php
require_once 'URLNormalizer.php';
echo (new Normalizer('http://example.com/a/b/../c/d'))->normalize(), "\n";
echo (new Normalizer('m:/.//'))->normalize(), "\n";
?>
$ php test.php
http://example.com/a/c/d
m:/
$ cat test.pl
use feature qw(say);
use URI::Normalize qw(normalize_uri);
say normalize_uri(URI->new('http://example.com/a/b/../c/d'));
say normalize_uri(URI->new('m:/.//'));
$ perl test.pl
http://example.com/a/c/d
m://
$ cat test.rb
require 'addressable/uri'
p Addressable::URI.parse('http://example.com/a/b/../c/d').normalize.to_s
p Addressable::URI.parse('m:/.//').normalize.to_s
$ ruby test.rb
"http://example.com/a/c/d"
[...]

Issue #654, opened 2020-12-02
Slicing error found by afl.rs


StevenJiang1110:
I've used afl.rs to heavily test url2.2.0, and some slicing error may occured when specific input is provided.
To replay the panic, the test code is
```rust
let mut _local1 = url::Url::parse("arhttpsps:/.//eom/dae.com/\\\\t\\:").unwrap();
let _ = url::quirks::set_hostname(&mut _local1 ,"//eom/datcom/\\\\t\\://eom/data.cs");
```
I also placed the replay file at https://github.com/StevenJiang1110/afl_rust_crashes/tree/main/url/replay_url40
I hope you will check if this is an actually bug. Thanks a lot.
The simple bug report is like this.
<img width="1375" alt="截屏2020-12-02 下午4 24 57" src="https://user-images.githubusercontent.com/27764680/100847698-19202d80-34bb-11eb-8d2a-2f3fc1686b9d.png">

StevenJiang1110:
This is the totol report with `RUST_BACKTRACE=full`
<img width="1379" alt="截屏2020-12-02 下午4 28 04" src="https://user-images.githubusercontent.com/27764680/100847990-79af6a80-34bb-11eb-915e-1c5368aae3ae.png">

djc:
I've fixed the immediate issue in #655, but with URLs like this it looks like there are still some issues. For example, in the newly added tests, the before and after URLs don't have the same string representation:

```
thread 'fuzzed' panicked at 'assertion failed: `(left == right)`
  left: `"arhttpsps://eom/dae.com/\\\\t\\:"`,
 right: `"arhttpsps:////eom/dae.com/\\\\t\\:"`', url/tests/unit.rs:676:5
```

(It looks like `set_hostname()` ends up setting an empty hostname for the `Url`, which does... something weird.)

StevenJiang1110:
I put more inputs that may cause same crash on https://github.com/StevenJiang1110/afl_rust_crashes/tree/main/url
Your fix can successfully pass all the replay files.
The input seems to have some characteristics, but I can't summarize.
I will run afl on the patched version. 
Thanks a lot.

djc:
Thank you for doing the fuzzing, it's very valuable to have someone do the work on this. Ideally we'd commit it to the repository so that we can run it again in the future after making changes, is that something you'd be able to work on?

StevenJiang1110:
[...]

PR #817, opened 2023-02-28
Fix anarchist URL where path starts with //

qsantos:
This PR closes https://github.com/servo/rust-url/issues/799 by refusing to parse URLs with no authority whose normalized path would start with a double-slash.

---

The issue comes from the [remove dot segments](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.4) step. Let's consider the URI `m:/.//`. Then, according to the RFC:

> B.  if the input buffer begins with a prefix of "/./" […], then replace that prefix with "/"

So it should be normalized to `m://`, but this has different semantics (resulting in `\` being interpreted as being part of the authority in the original example).

I have conducted a few tests with some URI normalization libraries:

```
$ node test.js
http://example.com/a/c/d
m:/.//
$ cat test.php
<?php
require_once 'URLNormalizer.php';
echo (new Normalizer('http://example.com/a/b/../c/d'))->normalize(), "\n";
echo (new Normalizer('m:/.//'))->normalize(), "\n";
?>
$ php test.php
http://example.com/a/c/d
m:/
$ cat test.pl
use feature qw(say);
use URI::Normalize qw(normalize_uri);
say normalize_uri(URI->new('http://example.com/a/b/../c/d'));
say normalize_uri(URI->new('m:/.//'));
$ perl test.pl
http://example.com/a/c/d
m://
$ cat test.rb
require 'addressable/uri'
p Addressable::URI.parse('http://example.com/a/b/../c/d').normalize.to_s
p Addressable::URI.parse('m:/.//').normalize.to_s
$ ruby test.rb
"http://example.com/a/c/d"
/usr/share/rubygems-integration/all/gems/addressable-2.8.1/lib/addressable/uri.rb:2487:in `validate': Cannot have a path with two leading slashes without an authority set: 'm://' (Addressable::URI::InvalidURIError)
	from /usr/share/rubygems-integration/all/gems/addressable-2.8.1/lib/addressable/uri.rb:2410:in `defer_validation'
	from /usr/share/rubygems-integration/all/gems/addressable-2.8.1/lib/addressable/uri.rb:839:in `initialize'
	from /usr/share/rubygems-integration/all/gems/addressable-2.8.1/lib/addressable/uri.rb:2184:in `new'
	from /usr/share/rubygems-integration/all/gems/addressable-2.8.1/lib/addressable/uri.rb:2184:in `normalize'
	from test.rb:3:in `<main>'
```

For PHP, I am using https://github.com/glenscott/url-normalizer. In short:

- Node does not remove the dot before a double-slash; I think the logic is [somewhere in there](https://searchfox.org/mozilla-central/source/netwerk/base/nsNetUtil.cpp#1815), but I never liked browsing Firefox's source code, so who knows?
- PHP removes the dot but avoids leaving a double-slash where it would be reinterpreted as the authority; this is done by [merging consecutive slashes](https://github.com/glenscott/url-normalizer/blob/master/src/URL/Normalizer.php#L262), which is does not conform to the RFC
- Perl exhibits the same bug as `rust-url`
- Ruby straight up refuses to parse a path with two leading slashes without an authority

I feel like Ruby's is the most consistent and straightforward solution.

This does mean that the [`no_panic` test](https://github.com/servo/rust-url/blob/master/url/tests/unit.rs#L951) from https://github.com/servo/rust-url/issues/654 must be amended. However, we can cover the `m:/.//` case in a dedicated unit test.

### DIFF
--- a/url/src/parser.rs
+++ b/url/src/parser.rs
@@ -1353,7 +1353,7 @@ impl<'a> Parser<'a> {
         host_end: u32,
         host: HostInternal,
         port: Option<u16>,
-        path_start: u32,
+        mut path_start: u32,
         remaining: Input<'_>,
     ) -> ParseResult<Url> {
         // Special case for anarchist URL's with a leading empty path segment
@@ -1364,37 +1364,44 @@ impl<'a> Parser<'a> {
         // If url’s host is null, url does not have an opaque path,
         // url’s path’s size is greater than 1, and url’s path[0] is the empty string,
         // then append U+002F (/) followed by U+002E (.) to output.
-        let scheme_end = scheme_end as usize;
-        let mut path_start = path_start as usize;
-        if path_start == scheme_end + 1 {
+        let scheme_end_as_usize = scheme_end as usize;
+        let path_start_as_usize = path_start as usize;
+        if path_start_as_usize == scheme_end_as_usize + 1 {
             // Anarchist URL
-            if self.serialization[path_start..].starts_with("//") {
+            if self.serialization[path_start_as_usize..].starts_with("//") {
                 // Case 1: The base URL did not have an empty path segment, but the resulting one does
                 // Insert the "/." prefix
-                self.serialization.insert_str(path_start, "/.");
+                self.serialization.insert_str(path_start_as_usize, "/.");
                 path_start += 2;
             }
-            assert!(!self.serialization[scheme_end..].starts_with("://"));
-        } else if path_start == scheme_end + 3
-            && &self.serialization[scheme_end..path_start] == ":/."
+            assert!(!self.serialization[scheme_end_as_usize..].starts_with("://"));
+        } else if path_start_as_usize == scheme_end_as_usize + 3
+            && &self.serialization[scheme_end_as_usize..path_start_as_usize] == ":/."
         {
             // Anarchist URL with leading empty path segment
             // The base URL has a "/." between the host and the path
             assert_eq!(
-                self.serialization.as_bytes().get(path_start).copied(),
+                self.serialization
+                    .as_bytes()
+                    .get(path_start_as_usize)
+                    .copied(),
                 Some(b'/')
             );
-            if self.serialization.as_bytes().get(path_start + 1).copied() != Some(b'/') {
+            if self
+                .serialization
+                .as_bytes()
+                .get(path_start_as_usize + 1)
+                .copied()
+                != Some(b'/')
+            {
                 // Case 2: The base URL had an empty path segment, but the resulting one does not
                 // Remove the "/." prefix
                 self.serialization
-                    .replace_range(scheme_end..path_start, ":");
+                    .replace_range(scheme_end_as_usize..path_start_as_usize, ":");
                 path_start -= 2;
             }
-            assert!(!self.serialization[scheme_end..].starts_with("://"));
+            assert!(!self.serialization[scheme_end_as_usize..].starts_with("://"));
         }
-        let scheme_end = to_u32(scheme_end)?;
-        let path_start = to_u32(path_start)?;
 
         let (query_start, fragment_start) =
             self.parse_query_and_fragment(scheme_type, scheme_end, remaining)?;
--- a/url/src/parser.rs
+++ b/url/src/parser.rs
@@ -819,6 +819,37 @@ impl<'a> Parser<'a> {
                         self.parse_path(scheme_type, &mut true, base_url.path_start as usize, input)
                     }
                 };
+                // Special case for anarchist URL's with a leading empty segment
+                let scheme_end = base_url.scheme_end as usize;
+                let mut path_start = base_url.path_start as usize;
+                if !base_url.has_authority() {
+                    if path_start == scheme_end + 1 {
+                        // The path starts right after the scheme_end
+                        if self.serialization[path_start..].starts_with("//") {
+                            // Case 1: The base URL did not have an empty segment, but the resulting one does
+                            // Insert the "/." prefix
+                            self.serialization.insert_str(path_start, "/.");
+                            path_start += 2;
+                        }
+                    } else {
+                        assert_eq!(path_start, scheme_end + 3);
+                        assert_eq!(&self.serialization[scheme_end..path_start], ":/.");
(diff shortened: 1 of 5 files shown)


valenting:
You reference an RFC. This crate explicitly only implements the [URL Standard](https://url.spec.whatwg.org/).
According to the reference URL parser `m:/.//\\` should [parse successfully](https://jsdom.github.io/whatwg-url/#url=bTovLi8vXFw=&base=YWJvdXQ6Ymxhbms=)

qsantos:
My bad. Thanks for the proper reference. Then, it looks like this comes from the serialization part, more specifically, from [section 4.5](https://url.spec.whatwg.org/#url-serializing), item 3, whose note explicitly mentions the case we are discussing. I have correspondingly moved the correction to the serialization.

qsantos:
Rebased to take into account the new tests

url/tests/urltestdata.json
```diff
@@ -7668,6 +7660,7 @@
     "hash": ""
   },
   "Do not serialize /. in path",
+  "skip next",
```
valenting, line 7663: I think we need to also explicitly handle this part [here](https://github.com/servo/rust-url/blob/edeaea7f1e8bd08148bcd535f921fbc645bce53e/url/src/parser.rs#L792-L794)

qsantos:
I have moved the handling of empty fragment to `with_query_and_fragment`. That way, it should be able to normalize URLs whether they are a fresh parse, or a combination of an existing URL and a relative path.

url/src/parser.rs
```diff
@@ -1354,9 +1353,56 @@ impl<'a> Parser<'a> {
...
+        //
+        // If url’s host is null, url does not have an opaque path,
+        // url’s path’s size is greater than 1, and url’s path[0] is the empty string,
+        // then append U+002F (/) followed by U+002E (.) to output.
+        let scheme_end_as_usize = scheme_end as usize;
+        let path_start_as_usize = path_start as usize;
+        if path_start_as_usize == scheme_end_as_usize + 1 {
+            // Anarchist URL
+            if self.serialization[path_start_as_usize..].starts_with("//") {
+                // Case 1: The base URL did not have an empty fragment, but the resulting one does
```
valenting, line 1372: fragment -> something else.

url/src/parser.rs
```diff
@@ -1354,9 +1353,56 @@ impl<'a> Parser<'a> {
...
+            && &self.serialization[scheme_end_as_usize..path_start_as_usize] == ":/."
+        {
+            // Anarchist URL with leading empty fragment
+            // The base URL has a "/." between the host and the path
+            assert_eq!(
+                self.serialization
+                    .as_bytes()
+                    .get(path_start_as_usize)
+                    .copied(),
+                Some(b'/')
```
valenting, line 1388: This seems overly verbose. `self.serialization.as_bytes()[path_start_as_usize] == b'/'` should work?

url/src/parser.rs
```diff
@@ -1354,9 +1353,56 @@ impl<'a> Parser<'a> {
         host_end: u32,
         host: HostInternal,
         port: Option<u16>,
-        path_start: u32,
+        mut path_start: u32,
         remaining: Input<'_>,
     ) -> ParseResult<Url> {
+        // Special case for anarchist URL's with a leading empty fragment
```
valenting, line 1359: fragment has a special meaning in the context of a URL - it's the bit after the # sign. Let's use another word here instead.

url/src/parser.rs
```diff
@@ -1354,9 +1353,56 @@ impl<'a> Parser<'a> {
...
+                // Case 1: The base URL did not have an empty fragment, but the resulting one does
+                // Insert the "/." prefix
+                self.serialization.insert_str(path_start_as_usize, "/.");
+                path_start += 2;
+            }
+            assert!(!self.serialization[scheme_end_as_usize..].starts_with("://"));
+        } else if path_start_as_usize == scheme_end_as_usize + 3
+            && &self.serialization[scheme_end_as_usize..path_start_as_usize] == ":/."
+        {
+            // Anarchist URL with leading empty fragment
```
valenting, line 1381: fragment -> something else.

url/src/parser.rs
```diff
@@ -1354,9 +1353,56 @@ impl<'a> Parser<'a> {
...
+                Some(b'/')
+            );
+            if self
+                .serialization
+                .as_bytes()
+                .get(path_start_as_usize + 1)
+                .copied()
+                != Some(b'/')
+            {
+                // Case 2: The base URL had an empty fragment, but the resulting one does not
```
valenting, line 1397: fragment -> something else.

valenting (changes requested):
Great progress, thank for working on this. I only have a few small issues.
Also, it would be nice to add some unit tests to improve code coverage.

url/src/parser.rs
qsantos: Ah yes, thanks, I meant “path segment”. I have rebased for a less confusing history with the proper terminology.

url/src/parser.rs
qsantos: Absolutely! I edited this code quite a bit, and did not notice that the `.get()` was not necessary anymore here. Thanks!

url/src/parser.rs
qsantos: Done

url/src/parser.rs
qsantos: Done

valenting (approved):
Great stuff. Thank you!

qsantos:
Thanks for the reactivity and the detailed comments!

url/src/slicing.rs
```diff
@@ -149,7 +149,14 @@ impl Url {
                 }
             }
 
-            Position::AfterPort => self.path_start as usize,
+            Position::AfterPort => {
+                if let Some(port) = self.port {
+                    debug_assert!(self.byte_at(self.host_end) == b':');
+                    self.host_end as usize + ":".len() + port.to_string().len()
```
klensy, line 155: Oww, this can be done without allocating string.

qsantos: Done

PR #818, opened 2023-02-28
Replace Host::IPv4/IPv6 with single variant taking IpAddr

Hyxogen:
This change will make it easier to use the Host enum in a more generic way with the standard library.


valenting:
This would constitute a breaking change.
Could you outline _how_ this change makes it easier to use Host with the standard library?
Some examples or links to other crates that are having this issue would be appreciated.

PR #819, opened 2023-02-28
Update urltestdata.json WPT test cases

valenting:


lucacasonato (approved):
LGTM - maybe would be good to break out expectations into a separate file in a follow up to make updating the test data easier. But seems low prio.

valenting:
Thanks!

PR #820, opened 2023-03-01
Update setters_tests WPT test cases

CYBAI:
@valenting as I found you've sent a PR to update `urltestdata.json`, this PR will only update `setters_tests.json`. 🙏


CYBAI:
I just moved the test import of C0 control for protocol into #816. I will look into other failures so please let me close this one now.

PR #806, opened 2022-12-27
set empty password tests

raccmonteiro:
Two tests regarding empty passwords and output of `Url` `as_str()`

Issue #796, opened 2022-09-26
URL with username/password removed creates invalid URL with .to_string()


varjolintu:
When setting username/password to empty string value, `:` and `@` chars should be removed from the URL when using `to_string()`.

Steps to reproduce:
- Have a `String` URL with username and password: `https://username:password@example.com`.
- Create a new URL using `Url::parse()` from the String.
- Use `set_username()` and `set_password()` to reset the values as empty strings.
- `.as_str()` or `.to_string()` for the `Url` shows: `https://:@example.com/`.

varjolintu:
Same applies to `query`. If URL has a query and it is set to empty string using `set_query()`, printing the URL still shows `?` at the end of the URL.

raccmonteiro:
There's a difference between `set_password(Some(""))` and `set_password(None)`, the first has an empty string as password and output of `.as_str()` is `https://:@example.com/` as you mentioned, the later does not have a password and the output is `https://example.com/`.
Check the new tests in this PR https://github.com/servo/rust-url/pull/806

qsantos:
From the [URL Standard](https://url.spec.whatwg.org/), in [URL serializing](https://url.spec.whatwg.org/#url-serializing):

> 2. If url [includes credentials](https://url.spec.whatwg.org/#include-credentials), then:
>     1. Append url’s [username](https://url.spec.whatwg.org/#concept-url-username) to output.
>     2. If url’s [password](https://url.spec.whatwg.org/#concept-url-password) is not the empty string, then append U+003A (:), followed by url’s [password](https://url.spec.whatwg.org/#concept-url-password), to output.
>     3. Append U+0040 (@) to output.

The first link of the quote leads to:

> A [URL](https://url.spec.whatwg.org/#concept-url) includes credentials if its [username](https://url.spec.whatwg.org/#concept-url-username) or [password](https://url.spec.whatwg.org/#concept-url-password) is not the empty string.

So I think the `set_password(Some(""))` should have the same effect on the URL as `set_password(None)`, and we should not leave an empty `:@`.

[...]

PR #825, opened 2023-03-14
No colon when setting empty password

qsantos:
This addresses https://github.com/servo/rust-url/issues/796 by removing the colon (`:`) when setting the password to `Some("")`. According to the [URL Standard](https://url.spec.whatwg.org/#url-serializing):

> If url’s [password](https://url.spec.whatwg.org/#concept-url-password) is not the empty string, then append U+003A (:), followed by url’s [password](https://url.spec.whatwg.org/#concept-url-password), to output.

This cherry-picks https://github.com/servo/rust-url/pull/806 to add some tests for this.

### DIFF
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -269,7 +269,7 @@ fn ends_in_a_number(input: &str) -> bool {
     } else {
         last
     };
-    if !last.is_empty() && last.chars().all(|c| ('0'..='9').contains(&c)) {
+    if !last.is_empty() && last.chars().all(|c| c.is_ascii_digit()) {
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


url/src/host.rs
```diff
@@ -269,7 +269,7 @@ fn ends_in_a_number(input: &str) -> bool {
     } else {
         last
     };
-    if !last.is_empty() && last.chars().all(|c| ('0'..='9').contains(&c)) {
+    if !last.is_empty() && last.chars().all(|c| c.is_ascii_digit()) {
```
klensy, line 272: You can replace `chars()` with `as_bytes()` here and lower, as it expects for ASCII range only, which is probably faster, than walking over utf8 code points.

qsantos: Ah yes, very good point!

valenting (approved):
Thanks!

PR #826, opened 2023-03-16
Url is special

### LINKED ISSUE #800: Please could you export `SchemeType` or add `is_special()` method to `Url`
Currently there's no way to check if a url/scheme is "special" without duplicating logic from 

https://github.com/servo/rust-url/blob/1c1e406874b3d2aa6f36c5d2f3a5c2ea74af9efb/url/src/parser.rs#L161-L167

Would be great if this could be made public, or an `is_special` method added to `Url`.

Also, I guess `SchemeType::from` should probably be implemented via the `Into` trait?

qsantos:
This closes #800 by exposing `Url::is_special`

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -819,6 +819,27 @@ impl Url {
         self.slice(..self.scheme_end)
     }
 
+    /// Return whether the URL is special (has a special scheme)
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use url::Url;
+    /// # use url::ParseError;
+    ///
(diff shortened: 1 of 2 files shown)


samuelcolvin (commented):
apart from tests, LGTM.

qsantos:
I have addressed the new linting rules in https://github.com/servo/rust-url/pull/825. I'll avoid moving the fix to this PR since there is already a discussion there. I'll rebase once #825 is merged.

valenting (approved):

PR #827, opened 2023-03-18
Update msrv to 1.56 to keep up with serde-derive

valenting:
The latest update of serde-derive increased its msvr to 1.56 causing our 1.51 CI to [fail](https://github.com/servo/rust-url/actions/runs/4454647785/jobs/7827117227#step:4:117).


lucacasonato (approved):
LGTM

PR #829, opened 2023-03-30
Explicitly depend on serde_derive

Veykril:
The `derive` feature of `serde` puts `serde` after `serde_derive` in the build graph which can hurt compile times. Depending on `serde_derive` explicitly avoids this, and as the change is rather non-invasive I figured this might be okay to submit.

see https://github.com/rust-lang/rust-analyzer/pull/14450


Veykril:
Apparently it is recommended to use the `derive` feature so I'll go ahead and close this, sorry for the noise

PR #836, opened 2023-05-02
Add feature "disable_idna"

parasyte:
This has a fairly colorful history.

The `idna` dependency was first made optional in #728, and then reverted in #790. This PR introduces a feature flag to opt-in to ASCII-only domain support, as suggested in https://github.com/servo/rust-url/pull/790#issuecomment-1239186515. It partially reverts #790.

Unfortunately, this opt-in method is less robust than using a default-enabled feature and making `idna` optional (#728) because `idna` is always built. It is trivial for future changes to accidentally link `idna` unconditionally, making the `disable_idna` feature a no-op.

In my testing, using this feature flag saves around 235 KiB in the produced binary in release mode on Windows x86_64 builds. (Side note: Windows builds with VS are "always stripped", and debug symbols are provided in a separate pdb file. For everything else, the binary `Cargo.toml` can include `strip = "symbols"` in the release profile.) The feature was originally requested for WASM, and binary size deltas should be in the same ballpark with that arch.

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -35,6 +35,8 @@ serde = {version = "1.0", optional = true, features = ["derive"]}
 
 [features]
 default = []
+# Produce smaller binaries by omitting Unicode domain support
+disable_idna = []
 # UNSTABLE FEATURES (requires Rust nightly)
 # Enable to use the #[debugger_visualizer] attribute.
 debugger_visualizer = []
--- a/url/src/host.rs
+++ b/url/src/host.rs
@@ -162,8 +162,21 @@ impl Host<String> {
     }
 
     /// convert domain with idna
+    #[cfg(not(feature = "disable_idna"))]
     fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
         idna::domain_to_ascii(domain).map_err(Into::into)
+    }
+
+    /// checks domain is ascii
+    #[cfg(feature = "disable_idna")]
+    fn domain_to_ascii(domain: &str) -> Result<String, ParseError> {
+        // without idna feature, we can't verify that xn-- domains correctness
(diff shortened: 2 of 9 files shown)


valenting:
So, I'm less and less convinced that this is a direction we want to pursue.
If a crate using URL wants to opt out of the IDNA, they can replace the idna dependency in Cargo.toml with a dummy crate that implements `domain_to_unicode`, `domain_to_ascii`, `domain_to_ascii_strict`, and contains `struct Errors`- see instructions [here](https://doc.bccnsoft.com/docs/rust-1.36.0-docs-html/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html)

valenting:
Rather than adding all this to the url crate, instead I suggest patching out idna as suggested earlier.
I have an example [here](https://github.com/valenting/no-idna/blob/main/example/Cargo.toml)
Pretty much the only requirement is adding the following to your Cargo.toml
```
[patch.crates-io]
idna = {git = 'https://github.com/valenting/no-idna.git'}
```

This avoids adding any spec incompatible flag to this crate.
If this works for you I propose we close this issue.

parasyte:
I don't think patch works with published crates. If I can't get rid of the heavy IDNA dependency in my published project, that doesn't seems like a useful workaround.

micolous:
I've written an alternative to this for `wasm32-unknown-unknown` in #887.

For Windows, there are platform IDN bindings which `idna` could use instead, which should be significantly smaller for projects which already link against `windows-rs`.

hsivonen:
With `url` 2.5.3 and `idna` 1.0.3, it's now possible to disable idna data and support for non-ASCII domain input. See https://docs.rs/crate/idna_adapter/latest for instructions.

PR #842, opened 2023-06-14
Only strip spaces from opaque path when both query and fragment are null

ohno418:
In the current behavior, when `url.set_query(None)` is called to remove
the query from a opaque path, spaces are stripped even if a fragment
remains.

As the spec [1] says, strip spaces only when both the query and fragment
are null. For example:

```
let mut url = URL::parse("data:space   ?query#hash");
url.set_query(None);
assert_eq!(url.as_str(), "data:space   #hash")
```

[1]: https://url.spec.whatwg.org/#potentially-strip-trailing-spaces-from-an-opaque-path

---

You can find the related test here:
https://github.com/web-platform-tests/wpt/pull/37556/files#diff-316fd17e018fdca0e399146c072ea3a95a08fe1ccd3f8fe940d73ca2b9d26c9eR57-R63

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -1552,12 +1552,12 @@ impl Url {
                     parser::Input::trim_tab_and_newlines(input, vfn),
                 )
             });
+            self.restore_already_parsed_fragment(fragment);
         } else {
             self.query_start = None;
+            self.restore_already_parsed_fragment(fragment);
             self.strip_trailing_spaces_from_opaque_path();
         }
-
-        self.restore_already_parsed_fragment(fragment);
     }
 
     /// Manipulate this URL’s query string, viewed as a sequence of name/value pairs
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -1552,12 +1552,12 @@ impl Url {
                     parser::Input::trim_tab_and_newlines(input, vfn),
                 )
             });
+            self.restore_already_parsed_fragment(fragment);
         } else {
             self.query_start = None;
+            self.restore_already_parsed_fragment(fragment);
             self.strip_trailing_spaces_from_opaque_path();
```
ohno418, line 1559: Stripping spaces should only happen when the fragment is null, so the fragment should be restored first.

valenting:
I think this was fixed in #879

PR #843, opened 2023-06-14
Fix no_std Support for idna

domenukk:
This fixes `no_std` builds for the `idna` crate.


domenukk:
The CI error doesn't seem related to this PR:
`error: package `log v0.4.19` cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.56.0
`

lucacasonato (commented):
It's concerning we didn't catch this in CI. We should perform `no_std` builds for all crates in CI to catch stuff like this.

domenukk:
There's a CI check in place in PR #831

lucacasonato (approved):
LGTM, thanks

PR #845, opened 2023-07-02
Resolves #844 by returning `Ok(())` if username or password is empty

### LINKED ISSUE #844: `set_username` & `set_password` should check for a value before returning `Err(())`
The methods `set_username` and `set_password` both attempt to fail early if called when the url cannot have a username/password:

```rust
    pub fn set_username(&mut self, username: &str) -> Result<(), ()> {
        // has_host implies !cannot_be_a_base
        if !self.has_host() || self.host() == Some(Host::Domain("")) || self.scheme() == "file" {
            return Err(());
        }
       // ...
```

```rust
    pub fn set_password(&mut self, password: Option<&str>) -> Result<(), ()> {
        // has_host implies !cannot_be_a_base
        if !self.has_host() || self.host() == Some(Host::Domain("")) || self.scheme() == "file" {
            return Err(());
        }
        // ...
```

I believe they should first check for the existence of a value (or empty string, in the case of `set_username`) before returning an error.

```diff
    pub fn set_username(&mut self, username: &str) -> Result<(), ()> {
        // has_host implies !cannot_be_a_base
        if !self.has_host() || self.host() == Some(Host::Domain("")) || self.scheme() == "file" {
+           if(username.is_empty()) {
+                return Ok(());
+           }
            return Err(());
        }
```

```diff
    pub fn set_password(&mut self, password: Option<&str>) -> Result<(), ()> {
        // has_host implies !cannot_be_a_base
        if !self.has_host() || self.host() == Some(Host::Domain("")) || self.scheme() == "file" {
+           if password.is_none() {
+                return Ok(());
+           }
            return Err(());
        }
```

chanced:
resolves #844 by checking for a value before returning `Err(())` in `set_username` and `set_password`.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -2088,6 +2088,9 @@ impl Url {
     pub fn set_password(&mut self, password: Option<&str>) -> Result<(), ()> {
         // has_host implies !cannot_be_a_base
         if !self.has_host() || self.host() == Some(Host::Domain("")) || self.scheme() == "file" {
+            if password.is_none() || password == Some("") {
+                return Ok(());
+            }
             return Err(());
         }
         let password = password.unwrap_or_default();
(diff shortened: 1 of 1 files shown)


chanced:
This pull request is lacking in tests at the moment. I'll add them if ya'll are interested in it.

PR #851, opened 2023-07-11
Enable the GitHub merge queue

mrobinson:


lucacasonato (approved):
LGTM - you'll need to land https://github.com/servo/rust-url/pull/849 to unblock CI if you'd like to land this on a green main :)

mrobinson:
Huh. Something is wrong as the final step passed even with an intermediate failure.

.github/workflows/main.yml
```diff
@@ -75,3 +77,20 @@ jobs:
...
+    if: always()
+    needs:
+      - "Test"
+      - "WASM"
+      - "Lint"
+      - "Audit"
+    steps:
+      - name: Mark the job as successful
+        run: exit 0
+        if: success()
```
lucacasonato, line 93:
```suggestion
        if: needs.Test.result == 'success' && needs.WASM.result == 'success' && needs.Lint.result == 'success' && needs.Audit.result == 'success'
```

And then reverse for the below `if`?

lucacasonato: Or you can invert this if, change it to `exit 1`, and remove the `exit 0` job (it's not needed as exit 0 === no step)

mrobinson: Looks like the issue was `if: always()` which can be removed in this case.

lucacasonato (approved):
LGTM

mukilan (approved):

mrobinson:
@valenting @lucacasonato Thanks for looking at this. In order to land changes now, simply click the "Merge when ready" button on pull requests. The merge style is set to "squash" right now, but let me know if you'd like it to be "rebase" instead which will preserve all commits in the PR instead of squashing them.

mrobinson:
You need a job to control whether the CI passes or fails. This Result job tries to do that by failing if any dependent job fails. I'm not sure if it's working on this case or not.

PR #853, opened 2023-07-12
Move debugger_visualizer tests to separate crate

lucacasonato:
To prevent having to raise MSRV again, we move the debugger_visualizer
to a separate crate. This crate is only compiled when the feature
`debugger_visualizer` is enabled, because only then the
`debugger_visualizer` test is added as a compile target.

Implements @valenting's suggestion from #849


valenting (approved):

valenting:
Sweet. Thank you for implementing this.

PR #857, opened 2023-07-13
Rewrite WPT runner

lucacasonato:
This PR rewrites the WPT test data runner for `url` to use a custom harness.
This has the following benefits:

- Tests can now be filtered, for example
  `cargo test --test url_wpt "set username"` runs only tests with "set username"
  in the name.

- There is now an "expected_failures.txt" file that lists tests that are
  expected to fail, by name. The test harness will ensure that these tests fail,
  and will report a failure if they pass.


valenting (approved):
Great stuff!

PR #862, opened 2023-07-23
Fix #861 serde::Deserialize impl for Url prints confusing error message

### LINKED ISSUE #861: Fix `serde::Deserialize` impl for `Url` prints confusing error message
When failing to deserialize a `str` into a `Url` the current implementation of `Deserialize` gives a confusing message.

For example:
```
use serde_json::{json, Result};
use url::Url;

let json = json!("bad_url");
let result: Result<Url> = serde_json::from_value(json);
println!("{}", result.err().unwrap());
```
Actual:  `invalid value: string "bad_url", expected relative URL without a base`
Expected: `invalid value: string "bad_url", expected a string representing a URL`

justy777:
Fixes #861


denschub:
I think a better implementation would be

```rs
Error::invalid_value(Unexpected::Other(&err_s), &self)
```

which then would keep the original "why it's broken" info intact:

```
    TOML parse error at line 2, column 12
      |
    2 | root_url = "http://[::1]:123456"
      |            ^^^^^^^^^^^^^^^^^^^^^
    invalid value: invalid port number, expected a string representing an URL
```

justy777:
@denschub 
Thanks for your attention on my pull request.

I respectfully disagree with your implementation as it does not pass the parameter intended in serde's API and causes a different problem.

https://docs.rs/serde/latest/serde/de/trait.Error.html#method.invalid_value

"The unexp argument provides information about what value was received. This is the value that was present in the input file or other source data of the Deserializer."

"err_s" is not the invalid value, and the resulting message will not print the invalid value.

In a real world you could have several fields of type Url, if you don't print the invalid value, you'd have no idea which of the fields is wrong.

I understand that the reason it's invalid is relevant information, but it doesn't fit into serde's API.
serde is meant to serialize and deserialize, not to validate. validation should be done separately, and can be accomplished with the [validator](https://crates.io/crates/validator) crate instead.

denschub:
The documentation for [`Unxpected::Other`](https://docs.rs/serde/latest/serde/de/enum.Unexpected.html#variant.Other) describe the parameter to it as

> A message stating what uncategorized thing the input contained that was not expected.

an invalid port number is "a thing", it was not expected, and the input contained it. Your PR changes the implementation in a way that's even less helpful. Seeing a

> invalid value: string "http://localhost:804211/", expected a string representing an URL

does not help anyone figure out why the parsing failed. "serde is meant to parse, not to validate" is a valid statement, but you need to parse before you can validate, and nobody is helped if parsing fails for reasons not clear to anyone, and for reasons that are not communicated.

Ultimately, though, I'm not a maintainer of this create. And I don't really mind either way, because I have ways to work around both solutions.

justy777:
Those are good points.
serde could use an API that allows you to better communicates why it refuses to parse something.

Though in my opinion, if you want to provide a feature to provide an implementation for a particular trait it should do so in a way that's predictable and consistent with standard/core implementations for that trait.

At the end of the day, I ignore the "serde" feature and just use the NewType pattern over Url to implement serde::Serialize and serde::Deserialize or whatever functionality I happen to need. After almost a year with no comment or response from anyone, I don't see that changing anytime soon.

Thanks again for being the first to comment.

PR #866, opened 2023-08-30
Compile serde_derive separately from serde

xfix:
When `serde`'s `derive` feature is used, `serde_derive` must be compiled before `serde` can be, as `serde` with that feature has a `serde_derive` dependency.

As of serde 1.0.186, this issue can be avoided by adding a separate `serde_derive` dependency due to the fact that serde 1.0.186 has a never-applicable dependency on `serde_derive`, which ensures that there is no incompatible version of `serde_derive` in a program (https://github.com/serde-rs/serde/pull/2588).

Because MSRV being set to 1.56, it's not possible to use the `dep:` syntax in features, so `serde` crate needs to be renamed. `dep:` syntax was added in Rust 1.60 if we want to bump MSRV.

This should improve compilation times of programs that use `url` with its `serde` feature, provided it doesn't have other crates that use `serde` with its `derive` feature.

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -17,7 +17,7 @@ edition = "2018"
 rust-version = "1.56"
 
 [dev-dependencies]
-serde = { version = "1.0", features = ["derive"] }
+serde_crate = { package = "serde", version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 bencher = "0.1"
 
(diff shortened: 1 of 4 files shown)


xfix:
Albeit, it seems like according to https://github.com/serde-rs/serde/issues/2584, the `serde` may be split into `serde_core` which doesn't have `serde_derive` which would close this issue, closing for now.

PR #873, opened 2023-10-19
Reexport percent_encoding

lewiszlw:


valenting:
> `percent-encoding` is already an externally usable crate.

Exactly.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef54d980432a386e372b175f491e7124
```rust
use percent_encoding;
fn main() {
    println!("{}",percent_encoding::utf8_percent_encode("foo bar?", percent_encoding::NON_ALPHANUMERIC));
}
```

PR #876, opened 2023-11-02
Added #[must_use] Attributes for Configuration Options

Redfire75369:
Resolves issues similar to https://github.com/Redfire75369/spiderfire/pull/35, where the updated `ParseOptions` isn't used since it is `Copy`.

### DIFF
--- a/idna/src/uts46.rs
+++ b/idna/src/uts46.rs
@@ -509,6 +509,7 @@ impl Idna {
 }
 
 #[derive(Clone, Copy)]
+#[must_use]
 pub struct Config {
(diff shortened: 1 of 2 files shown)


valenting:
Thank you for the fix @Redfire75369 
@lucacasonato I was thinking this might be a breaking change, but I think it would only warn/break incorrect uses of the struct, so that would be OK?

lucacasonato (approved):
LGTM

Usually I'd ask for a test, but this seems difficult to test without `compiletest`

lucacasonato:
Yeah I agree, would only break invalid uses

Issue #870, opened 2023-10-15
Integer overflow when parsing long URL


Byron:
When parsing the attached URL ([long.url.zip](https://github.com/servo/rust-url/files/12909032/long.url.zip)) with `url::Url::parse()` when compiled with overflow checks, there will be an overflow in `idna` with the panic.

```
thread '<unnamed>' panicked at /rust/registry/src/index.crates.io-6f17d22bba15001f/idna-0.4.0/src/punycode.rs:272:17:
attempt to add with overflow
```

Without overflow checks, the same will cause an `IdnaError` to be created, but won't panic.

Since the URL is very long, I suppose that this issue can be prevented by avoiding to parse unrealistically long host and domain names, which makes this issue related to #868 .

Byron:
I have another URL of this kind, along with a reproducer directly in [the Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1481a041499cbad4cd541c4a755d206a).

Byron:
Here is another [Rust playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=52b98cabd5435407e3009f7e38c59fed) of the same overflow, but using a smaller URL.

Byron:
And another new test-case, which fails in the same spot.

[clusterfuzz-testcase-minimized-gix-url-parse-5849655294164992.zip](https://github.com/servo/rust-url/files/13243590/clusterfuzz-testcase-minimized-gix-url-parse-5849655294164992.zip)

It's sad that the fuzzer keeps creating new cases for the same issue, and I will stop reporting these here now.

valenting:
Fixed by https://github.com/servo/rust-url/pull/877

PR #877, opened 2023-11-03
use checked addition to not panic in debug build

Skgland:
used checked addition so this does not panic in debug builds

related to issue #870 

~~not sure if #870 is completely resolved as there are other comments indicating a possible overflow condition that I did not understand and as such did not fix~~


Skgland:
Its also probably not ideal tha the fixed code and the test are in different crates, but I didn't know how to separate the punycode from the url and how to add a test in the punycode crate as that appears to use a custom testing framework

Skgland:
> not sure if #870 is completely resolved as there are other comments indicating a possible overflow condition that I did not understand and as such did not fix

I think the line I was refering to is fine: 
`min_code_point - code_point > (u32::MAX - delta) / (processed + 1)`

`min_code_point` should be lager than `code_point` due to the iterator filter `c >= code_point` as such this can't underflow

`u32::MAX - delta` cannot underflow either as `delta` can't be larger than `u32:MAX`

`process + 1` can't overflow as `process < input_length` per the while condition

`process + 1` is not 0 as process is unsigned, so no division by zero

Skgland:
Restructured changes and moved test to idna where it belongs in my opinion.
- The first commit fixes the line endings of `debug_metadata/README.md` as it was always considered dirty and prevented me from rebasing otherwise.
- The second commit adds the test, fails without the third commit.
- The third and last commit fixed the panic in debug builds due to overflow.

Skgland:
Rebase again as I accidentally hardcoded the string in the Bad Punycode test instead of using the value(s) from `bad_punycode_tests.json`.

Skgland:
This should now be ready for review again.

Skgland:
Fixed added test to not increase msrv beyond 1.56 and fixed clippy lint.
Ready for review again.

Skgland:
Now with fixed formatting.
Ready for review again, sorry for the noise.

Skgland:
I messed up my rebase, one moment.

Skgland:
Ok, now the rebase should be fixed.

valenting (approved):
Thank you for the fix. I really appreciate it.

PR #880, opened 2023-11-13
Fix another overflow in punycode encode_into

Skgland:
It was possible to panic due to an overflow in build with panic on overflow (i.e. debug builds), if the input iterator contained more than `u32::MAX` elements.
Which can happen an systems with `target_pointer_width > 32` if a hugh string or char slice is passed to `encode` or `encode_str` respectively.

Commit 1. adds a failing test
Commit 2. fixes the implementation so that the test passes
Commit 3. adds some early detection to `encode` and `encode_str` to not waisted work allocating the String and iterating the input

### DIFF
--- a/idna/src/punycode.rs
+++ b/idna/src/punycode.rs
@@ -235,9 +235,9 @@ where
     I: Iterator<Item = char> + Clone,
 {
     // Handle "basic" (ASCII) code points. They are encoded as-is.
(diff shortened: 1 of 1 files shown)


valenting (approved):
Thanks for the fix.
I'll land this as is. I think we actually want to limit the input size even more, as IDNA labels that exceed 63 characters are not allowed in DNS anyway.

PR #885, opened 2023-11-22
Update url 2.5.0

valenting:


edgul (approved):
LGTM

idna/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "idna"
-version = "0.4.0"
+version = "0.5.0"
```
djc, line 3: It's pretty unclear to me why this got a semver-incompatible bump? I don't see any changes that require it -- causing me to have duplicate dependencies where none should be needed.

PR #886, opened 2023-11-28
Fix multiple issues on wasm32, and runs url tests in CI

micolous:
This fixes multiple issues on `wasm32-unknown-unknown`:

* Remove `Url::socket_addrs()` on `wasm32-unknown-unknown`. That function will never work – those platform API calls are not supported.
* Disable unit tests which won't work on `wasm32-unknown-unknown`.
* Run tests using `wasm_bindgen_test` on `wasm32-unknown-unknown`.
* Remove `panic::catch_unwind` from wpt tests, as that conflicts with wasm-bindgen-test's panic handler.
* Run `url` tests on `wasm32-unknown-unknown` in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,11 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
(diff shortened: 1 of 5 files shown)


micolous:
Any updates?

Issue #881, opened 2023-11-16
Mark Origin::unicode_serialization as deprecated?


evilpie:
The method descriptions links to https://html.spec.whatwg.org/multipage/browsers.html#unicode-serialisation-of-an-origin. Which just says:
> There used to also be a Unicode serialization of an origin. However, it was never widely adopted.

https://github.com/whatwg/html/pull/2689

PR #887, opened 2023-11-28
Use web_sys::Url for IDN conversion on wasm32-unknown-unknown

micolous:
**This pull request is based on #886, and that will need to be merged first. This has been marked as a draft until that's unblocked.**

This removes the dependency on `idna` on `wasm32-unknown-unknown`, replacing it with a small shim which calls out to `web_sys::Url`. This saves about 220 KB binary size on a release build, so that should go a lot of the way towards addressing #557, while *retaining IDN support*.

This removes anything which depends on `domain_to_unicode`, because I couldn't see an obvious way to implement it with web APIs. One usage is `Origin::unicode_serialization()` – which I've also marked as deprecated (#881). My 5¢ is that `domain_to_unicode` should be part of `idna` only, as nothing else in the WHATWG `Url` type really uses it (it's always either a punycode or percent-encoded string).

## Alternatives considered

* Downstream libraries could adopt `web_sys::Url` on `wasm32-unknown-unknown`.

  This is API-breaking, and complicated to handle, for reasons I've covered in #557 and https://github.com/seanmonstar/reqwest/issues/2050.

  There are similar pains discussed in https://github.com/sagebind/isahc/issues/382 for a `http::Uri` to `url::Url` migration.

* Shimming out *every* single piece of functionality to JavaScript's `Url` type (via `web_sys::Url`) on `wasm32-unknown-unknown`.

  This would probably make things smaller again, but there are enough subtle API differences to make this a problem. For example, `url::Url::host()` returns a `url::Host`, but `web_sys::Url::hostname()` returns a `String`.

* Shimming in `idna`.

  Not all the functionality required is supported in web APIs (`domain_to_unicode`), and so it needs changes to `Url` anyway.

  This also allows one to continue using the `idna` crate directly on `wasm32-unknown-unknown` with its normalisation database.

## Known issues

* There are a few extra test failures (see `expected_failures_{chromium,firefox,safari}.txt`) – I'm not certain which of these are browser bugs or `url::Host` bugs.

  In any case, I don't think I can fix them, and I don't know how much one should care.

* I haven't tried this on `redox` or `wasi` targets – they don't appear to be covered by CI, and so I've left them as-is.

  It may be possible to get similar binary size savings on those targets, but I'm keeping them out of scope for now.


micolous:
While this is based on #886, I've marked it as "ready for review", as this isn't really a work in progress – it's just blocked.

PR #891, opened 2023-12-28
Add UniFFI support impl FfiConverter for Url - gated behind `uniffi` feature flag.

Sajjon:
As [UniFFI](https://github.com/mozilla/uniffi-rs) gathers support, it would be convenient for Url to be able to be used with UniFFI!

Inspired by @novacrazy's work in https://github.com/Lantern-chat/iso8601-timestamp/pull/8 credits goes to Nova!

However, as pointed out by @mhammond at UniFFI users can instead of using this feature flag (part of Url) do it "manually" in their own project like so:

https://github.com/mozilla/uniffi-rs/blob/main/examples/custom-types/src/lib.rs#L36-L47

So might not be worth merging.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -3056,10 +3056,7 @@ mod uniffi_core_impl {
 
     use super::*;
 
-    use uniffi_core::{
-        deps::bytes::{Buf, BufMut},
-        metadata, FfiConverter, MetadataBuffer,
-    };
+    use uniffi_core::{metadata, FfiConverter, MetadataBuffer};
 
     uniffi_core::derive_ffi_traits!(blanket Url);
 
@@ -3077,7 +3074,10 @@ mod uniffi_core_impl {
         }
 
         const TYPE_ID_META: MetadataBuffer =
-            MetadataBuffer::from_code(metadata::codes::TYPE_STRING);
+            MetadataBuffer::from_code(metadata::codes::TYPE_CUSTOM)
+                .concat_str("url::Url")
+                .concat_str("Url")
+                .concat(<String as uniffi_core::Lower<String>>::TYPE_ID_META);
     }
 
     #[cfg(test)]
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -77,6 +77,9 @@ assert!(data_url.fragment() == Some(""));
 
 Enable the `serde` feature to include `Deserialize` and `Serialize` implementations for `url::Url`.
 
+## Uniffi
+- Implements `uniffi_core::FfiConverter` and friends for `Url`.
+
 # Base URL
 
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -3043,3 +3046,89 @@ impl<'a> Drop for UrlQuery<'a> {
...
+
+    use uniffi_core::{
+        deps::bytes::{Buf, BufMut},
+        metadata, FfiConverter, MetadataBuffer,
+    };
+
+    uniffi_core::derive_ffi_traits!(blanket Url);
+
+    unsafe impl<UT> FfiConverter<UT> for Url {
+        uniffi_core::ffi_converter_rust_buffer_lift_and_lower!(UT);
```
Sajjon, line 3067: This `impl` "piggybacks" on UniFFI Core's impl of String, see https://github.com/mozilla/uniffi-rs/blob/main/uniffi_core/src/ffi_converter_impls.rs#L148-L169

For sake of simplicity, but this requires us to roundtrip via `to_string` on Url and via `Url::from_str`, there might be faster implementations, but this was the most convenient to get something started.

url/src/lib.rs
```diff
@@ -3043,3 +3046,89 @@ impl<'a> Drop for UrlQuery<'a> {
...
+            <String as FfiConverter<UT>>::write(string, buf)
+        }
+
+        fn try_read(buf: &mut &[u8]) -> uniffi_core::Result<Self> {
+            <String as FfiConverter<UT>>::try_read(buf)
+                .and_then(|s| Url::from_str(s.as_str()).map_err(|e| e.into()))
+        }
+
+        const TYPE_ID_META: MetadataBuffer =
+            MetadataBuffer::from_code(metadata::codes::TYPE_STRING);
```
Sajjon, line 3080: I'm also a bit unsure about the implications of reusing `TYPE_STRING` as `TYPE_ID_META`, possibly more correct to use `::TYPE_CUSTOM`.

valenting:
> However, as pointed out by @mhammond at UniFFI users can instead of using this feature flag (part of Url) do it "manually" in their own project like so:
> https://github.com/mozilla/uniffi-rs/blob/main/examples/custom-types/src/lib.rs#L36-L47
> 
> So might not be worth merging.

Thank you for the PR. I'm not very familiar with UniFFI, so I'm not totally convinced the Url crate needs to support it just yet. If it becomes something that a large share of the crates need, then it would make sense to add the support in the Url crate.
As you mention, users can already do this manually, and the FfiConverter you added round-trips through `to_string` and `parse` anyway, so it has no inherent advantage to a user provided implementation.

I'll close this for now, but I'm open to reopening if the user needs change in the future.

PR #894, opened 2024-01-08
perf(punycode): avoid double allocation in decode_to_string

bishopcheckmate:
Hey,
while indeed `decode_to_string` is a nice ergonomics addition, the fact of being a wrapper over the `decode` results in an intermediate `Vec`.
Since the implementation of it can be the same as `decode` and it's quite straightforward, maybe we could get rid of double alloc cheaply

### DIFF
--- a/idna/src/punycode.rs
+++ b/idna/src/punycode.rs
@@ -41,10 +41,12 @@ fn adapt(mut delta: u32, num_points: u32, first_time: bool) -> u32 {
 
 /// Convert Punycode to an Unicode `String`.
 ///
-/// This is a convenience wrapper around `decode`.
+/// Return None on malformed input or overflow.
+/// Overflow can only happen on inputs that take more than
+/// 63 encoded bytes, the DNS limit on domain name labels.
 #[inline]
 pub fn decode_to_string(input: &str) -> Option<String> {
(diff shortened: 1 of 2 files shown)


valenting (approved):

idna/src/punycode.rs
```diff
@@ -41,10 +41,12 @@ fn adapt(mut delta: u32, num_points: u32, first_time: bool) -> u32 {
...
 /// Convert Punycode to an Unicode `String`.
 ///
-/// This is a convenience wrapper around `decode`.
+/// Return None on malformed input or overflow.
+/// Overflow can only happen on inputs that take more than
+/// 63 encoded bytes, the DNS limit on domain name labels.
 #[inline]
 pub fn decode_to_string(input: &str) -> Option<String> {
-    decode(input).map(|chars| chars.into_iter().collect())
+    Some(Decoder::default().decode(input).ok()?.collect())
```
valenting, line 49: Could you maybe add a unit test that exercises this method?

bishopcheckmate: sure, will do that at the evening

bishopcheckmate: it is now being tested together with encode_str and decode

PR #911, opened 2024-03-04
Fix lint

valenting:


valenting:
addr_eq was stabilized in 1.76 - Let's close this until we're ready to increase the MSRV

Issue #930, opened 2024-05-18
perf: Use `NonZeroU16` for port numbers


DonIsaac:
Changing URL's port field from `Option<u16>` to `Option<NonZeroU16>` would enable some memory layout optimizations, allowing `port` to have the same size as `u16`. Trying this out locally results in a 10% smaller memory footprint for `Url` (88 bytes -> 80 bytes).

Scripter17:
Is it possible to store the port as a plain `u16` and compare `Url::host_end` and `Url::path_start` to know when `Url::port` should return `None`?

Edit: Theoretically you can remove `Url::host_end` and calculate it from `Url::path_start` and `Url::port`. You shouldn't. The perf would be awful. But you trade saving 2 bytes for saving 4 bytes.

PR #931, opened 2024-05-18
perf: change ports type to `Option<NonZeroU16>`

DonIsaac:
Proposal implementation for #930

I've left the public API unchanged; all public port getter/setters still deal with `u16`s.


valenting:
Doesn't this make it impossible to parse `http://example.org:0/` ?
That's still a valid URL according to the URL standard.

DonIsaac:
Yes, it is a valid port. It appears to be [used for some use cases in the wild](https://ieeexplore.ieee.org/document/8816853). I'm wondering if/how we could get this memory layout improvement without impacting users that need port 0. Maybe a feature flag?

valenting:
I don't think this is possible for port, but I encourage you to explore options for
```
    query_start: Option<u32>,    // Before '?', unlike Position::QueryStart
    fragment_start: Option<u32>, // Before '#', unlike Position::FragmentStart
```

If I'm not mistaken, it should be impossible for them to start at 0.

PR #942, opened 2024-06-17
Revert "Reimplement idna on top of ICU4X (#923)"

valenting:
This reverts commit 3d6dbbb1dfc64c597745d5d6b97f2a8dd543c42b.

See #937 for reasons behind this backout.

### DIFF
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "idna"
-version = "1.0.0"
+version = "1.0.1"
 authors = ["The rust-url developers"]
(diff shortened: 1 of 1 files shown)


djc:
Might be easier to only revert the url -> idna dependency to 0.5 for now?

valenting:
> Might be easier to only revert the url -> idna dependency to 0.5 for now?

https://github.com/servo/rust-url/blob/3d6dbbb1dfc64c597745d5d6b97f2a8dd543c42b/url/Cargo.toml#L29
```toml
idna = { version = "1.0.0", path = "../idna" }
```

That might be a better idea, but we'd have to stop referring dependencies by path. :thinking:

Manishearth (approved):

PR #943, opened 2024-06-17
Normalize URL paths: convert /.//p, /..//p, and //p to p

theskim:
https://bugzilla.mozilla.org/show_bug.cgi?id=1874119

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -1709,6 +1709,39 @@ impl Url {
         let old_after_path_pos = to_u32(self.serialization.len()).unwrap();
         let cannot_be_a_base = self.cannot_be_a_base();
         let scheme_type = SchemeType::from(self.scheme());
+        let mut path_empty = false;
+
+        // Check ':' and then see if the next character is '/'
+        let mut has_host = if let Some(index) = self.serialization.find(":") {
+            if self.serialization.len() > index + 1
+                && self.serialization.as_bytes().get(index + 1) == Some(&b'/')
+            {
+                let rest = &self.serialization[(index + ":/".len())..];
+                let host_part = rest.split('/').next().unwrap_or("");
+                path_empty = rest.is_empty();
+                !host_part.is_empty() && !host_part.contains('@')
+            } else {
+                false
+            }
+        } else {
+            false
+        };
+
+        // Ensure the path length is greater than 1 to account
+        // for cases where "/." is already appended from serialization
+        // If we set path, then we already checked the other two conditions:
+        // https://url.spec.whatwg.org/#url-serializing
+        // 1. The host is null
+        // 2. the first segment of the URL's path is an empty string
+        if path.len() > 1 {
+            if let Some(index) = self.serialization.find(":") {
+                let removal_start = index + ":".len();
+                if self.serialization[removal_start..].starts_with("/.") {
+                    self.path_start -= "/.".len() as u32;
+                }
+            }
+        }
+
         self.serialization.truncate(self.path_start as usize);
         self.mutate(|parser| {
             if cannot_be_a_base {
@@ -1718,7 +1751,6 @@ impl Url {
                 }
                 parser.parse_cannot_be_a_base_path(parser::Input::new_no_trim(path));
             } else {
-                let mut has_host = true; // FIXME
                 parser.parse_path_start(
                     scheme_type,
                     &mut has_host,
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -1723,6 +1723,32 @@ impl Url {
...
+        // To handle cases like <non-spec:/> set pathname to </..//p>
+        // For instance, //p should be converted to /..//p here
+        // At this point, we would get "non-spec://p" for serialization
+        // and "/..//p" for path
+        if let Some(path_pos) = path.rfind("//") {
+            if let Some(serialization_pos) = self.serialization.rfind("//") {
+                const PATH_INCREMENT: usize = 2; // length of "/."
+
+                if path_pos + PATH_INCREMENT <= path.len()
+                    && serialization_pos + PATH_INCREMENT <= self.serialization.len()
```
valenting, line 1736: I'm not sure about these conditions.
Wouldn't they always be true?
Moreover,  I don't know if `path.rfind("//")` is the right check.
Step 3 of the serialization says:
> If url’s [host](https://url.spec.whatwg.org/#concept-url-host) is null, url does not have an [opaque path](https://url.spec.whatwg.org/#url-opaque-path), url’s [path](https://url.spec.whatwg.org/#concept-url-path)’s [size](https://infra.spec.whatwg.org/#list-size) is greater than 1, and url’s [path](https://url.spec.whatwg.org/#concept-url-path)[0] is the empty string, then append U+002F (/) followed by U+002E (.) to output.

Url's path size is greater that 1 and path[0] is empty string basically means path begins with `//`.
I don't think a rfind should be needed - if you think it is make sure to document it really well.

valenting (changes requested):
Thank you for the fix.
I think you're on the right track, but I'd like the fix to be more rooted in the spec.

url/src/lib.rs
theskim: This bug might need to be fixed after `has_host` is implemented, considering the case of:
```
"postgres://postgres@localhost"
"postgres:/.//postgres@localhost"
```

url/src/lib.rs
```diff
@@ -1709,6 +1709,33 @@ impl Url {
...
+        // 1. The host is null
+        // 2. the first segment of the URL's path is an empty string
+        // https://url.spec.whatwg.org/#url-serializing
+        if path.len() > 1 {
+            if let Some(index) = self.serialization.find(":") {
+                let removal_start = index + ":".len();
+                if self.serialization[removal_start..].starts_with("/.") {
+                    self.serialization
+                        .replace_range(removal_start..removal_start + "/.".len(), "");
+                    self.path_start -= "/.".len() as u32;
```
valenting, line 1734: Since we truncate at path_start just below, I think we don't necessarily need to replace_range, right?

url/src/lib.rs
```diff
@@ -1709,6 +1709,33 @@ impl Url {
         let old_after_path_pos = to_u32(self.serialization.len()).unwrap();
         let cannot_be_a_base = self.cannot_be_a_base();
         let scheme_type = SchemeType::from(self.scheme());
+        let mut path_empty = false;
+        let mut has_host = if let Some(index) = self.serialization.find(":/") {
```
valenting, line 1713: I'm a bit worried that this would also match URL's such as `scheme:hello:/fakehost`.
Can you instead find `:` and then check if the next character is `/` ?

valenting (changes requested):

url/src/lib.rs
theskim: Good point, thank you!

url/src/lib.rs
```diff
@@ -1751,14 +1783,28 @@ impl Url {
...
+
+        // For cases where normalization is applied across both the serialization and the path.
+        // Append "/." immediately after the scheme (up to ":")
+        // This is done if three conditions are met.
+        // https://url.spec.whatwg.org/#url-serializing
+        // 1. The host is null
+        // 2. The url's path length is greater than 1
+        // 3. the first segment of the URL's path is an empty string
+        if !has_host && path.len() > 1 && path_empty {
+            if let Some(index) = self.serialization.find("://") {
```
valenting, line 1802: I think we have a similar issue here.
`://` can certainly appear in the path part of a URL:
`scheme:/example.com/hello://bla`
`scheme:hello://bla`

url/src/lib.rs
```diff
@@ -1751,14 +1783,28 @@ impl Url {
...
         });
+
+        // For cases where normalization is applied across both the serialization and the path.
+        // Append "/." immediately after the scheme (up to ":")
+        // This is done if three conditions are met.
+        // https://url.spec.whatwg.org/#url-serializing
+        // 1. The host is null
+        // 2. The url's path length is greater than 1
+        // 3. the first segment of the URL's path is an empty string
+        if !has_host && path.len() > 1 && path_empty {
```
valenting, line 1801: Shouldn't the path_empty check apply to the new path that we just parsed in the block above?

valenting: Nevermind, The path empty bit applies to the first one, meaning if we have had a `/.` before setting the path, we needed to add back the `/.` (empty path) back afterwards.

valenting (approved):

url/src/lib.rs
```diff
@@ -1742,6 +1742,37 @@ impl Url {
         let old_after_path_pos = to_u32(self.serialization.len()).unwrap();
         let cannot_be_a_base = self.cannot_be_a_base();
         let scheme_type = SchemeType::from(self.scheme());
+        let mut path_empty = false;
+
+        // Check ':' and then see if the next character is '/'
+        let mut has_host = if let Some(index) = self.serialization.find(":/") {
```
valenting, line 1748: Actually, can we change this to - self.serialization.find(":") then check if the next char is `/` ?
That way we don't have to call find twice.

valenting (changes requested):

valenting (approved):

PR #949, opened 2024-07-01
Adjust Punycode overflow checks

hsivonen:
* The change made in 1.0.0 incorrectly assumed that the input length limit removed the need to do overflow check when decoding. Now the internal-caller length limit is taken as a permission to skip overflow checks only when encoding.
* The RFC gives overflow checking pre-flight math for languages like that don't have checked math. Since Rust does, the code now uses checked_add and checked_mul instead of pre-flight when overflow checks are performed.

### DIFF
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "idna"
-version = "1.0.1"
+version = "1.0.2"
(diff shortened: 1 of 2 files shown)


Manishearth (approved):
Would be nice to also have tests, but that's ok in a followup

PR #951, opened 2024-07-09
Support url parsing at compile-time via proc-macro

NobodyXu:
Added new crate `url_macro`, usage:

```rust
const URL: url::Url = url_macro::parse!("https://www.github.com");
assert_eq!(URL, url::Url::parse("https://www.github.com").unwrap());
```

# Motivation

It is not uncommon for users to want to parse `Url` at compile time, instead of having to use `OnceCell`/`OnceLock`, it would be more ergonomic/efficient to parse it at compile-time.

### DIFF
--- a/url-macro/src/lib.rs
+++ b/url-macro/src/lib.rs
@@ -1,0 +1,14 @@
+pub fn add(left: usize, right: usize) -> usize {
+    left + right
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn it_works() {
+        let result = add(2, 2);
+        assert_eq!(result, 4);
+    }
+}
--- a/url-macro/src/lib.rs
+++ b/url-macro/src/lib.rs
@@ -1,14 +1,74 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
+use quote::{quote, ToTokens};
+use syn::{parse_macro_input, LitStr};
+use url::{
+    quirks::{internal_components, InternalComponents},
+    Host, Url,
+};
+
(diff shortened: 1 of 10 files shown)


url/src/lib.rs
```diff
@@ -194,7 +194,7 @@ pub struct Url {
     ///   authority = "//" userinfo? host [ ":" port ]?
     ///   userinfo = username [ ":" password ]? "@"
     ///   hierarchical-path = [ "/" path-segment ]+
-    serialization: String,
+    serialization: Cow<'static, str>,
```
NobodyXu, line 197: Could replace this with [`beef::lean::Cow`](https://docs.rs/beef/latest/beef/lean/type.Cow.html) to reduce size of `Url`, but might be a pre-mature optimization

url-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+/// Parse url at compile-time
+///
+/// ```
+/// const URL: url::Url = url_macro::parse!("https://www.github.com");
+/// # fn main() {
+/// assert_eq!(URL, url::Url::parse("https://www.github.com").unwrap());
+/// # }
+/// ```
+#[proc_macro]
+pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
NobodyXu, line 25: TODO: Add test for `parse!`

Manishearth (changes requested):
I'm not convinced this is worth the addition of the Cow, or worth it for _this_ project to maintain

NobodyXu:
Thanks for the quick feedback, in that case I will close this.

PR #954, opened 2024-07-30
feat: Derive clone

dev-ardi:
Derive clone on some structs


valenting:
This is lacking in detail.
Could you explain why this is necessary & useful?

dev-ardi:
- There is no reason not to allow cloning.
- I use this crate and I wanted to clone the `Serializer` but I couldn't, so I opened this PR

wilfredjonathanjames:
Building the success/cancel urls for stripe checkout. I want to set up a base url with shared query params and clone once for cancelation specific params. No clone makes it necessary to define shared params twice, or reparse a modified url twice. Messy.

PR #963, opened 2024-09-01
Fix idna tests with no_std

cjwatson:
This failed since 464b1f7d8fe9de117af184c1acb52b4f821a0cdc:

  (cd idna && cargo test --all-targets --no-default-features --features alloc -- --ignored)


hsivonen (approved):
The change itself LGTM, but I don't know what to conclude from the failing CI runs.

hsivonen:
Not merging to give @valenting a chance to take a look at the CI status. To me, it looks like this change isn't supposed to introduce those failures.

PR #964, opened 2024-09-13
Merge main to the idna-v1x branch

hsivonen:
The point of this merge is to have commit 54346fa288e16b25b71c45149d7067c752b450e0 in this branch but in a way where the merge node makes it pretty much go away so that merging from idna-v1x to main in the future becomes feasible without the old revert blowing away the idna 1.0.x code.


hsivonen:
The same CI runs also failed in https://github.com/servo/rust-url/pull/963

valenting (approved):

PR #965, opened 2024-09-16
Enable choice from multiple Unicode back ends

hsivonen:
(This PR is intentionally to the `idna-v1x` branch as an intermediate step and not directly to `main`.)

This PR is a draft due to the dependency declaration using `path`.

[Build time charts](https://hsivonen.com/test/moz/reqwest-idna-m3-pro-adapter/) of `reqwest` without application code

Logically, the review here involves reviewing also:

* [The `main` branch of `idna_mapping`](https://github.com/hsivonen/idna_mapping/tree/main).
  - TODO before release:
    * Change version to 1.0.0
    * Optional: Move the repo from github.com/hsivonen/ to github.com/servo/
  - The `unicode-joining-type` dependency is behind this crate to hide its upcoming semver break from `idna_adapter` and to keep open the option to provide the data directly in `idna_mapping` if needed.
* `idna_adapter` branches:
  - Optional TODO before release: Move the repo from github.com/hsivonen/ to github.com/servo/
  - [`main` branch](https://github.com/hsivonen/idna_adapter/tree/main) (ICU4X 1.x from `icu_normalizer` 1.4.3 and `icu_properties` 1.4.2 onwards but not 2.x. Currently in practice 1.5.)
    * TODO before release: Change version to 1.2.0
  - [`unicode-rs` branch](https://github.com/hsivonen/idna_adapter/tree/unicode-rs) (`idna_mapping`, `unicode-normalization`, `unicode-bidi`)
    * TODO before release: Change version to 1.1.0
  - [`no-unicode` branch](https://github.com/hsivonen/idna_adapter/tree/no-unicode) (No Unicode back end; rejects Unicode-form input and does not check Punycode input for rules that require Unicode data)
    * [Tests are in a different crate](https://github.com/hsivonen/test_no_idna)
    * Obviously cannot pass all the tests from `url` and `idna`.
    * TODO before release: Change version to 1.0.0

`idna_adapter` also has a [branch called `icu4x-trunk`](https://github.com/hsivonen/idna_adapter/tree/icu4x-trunk). It has a `path` dependency to an adjacent `icu4x` checkout. Once ICU4X 2.0 has been released, the expectation is to merge this code to the `main` branch of `idna_adapter` as version 1.2.1.

I have omitted previously-prototyped ICU4X-1.x-relevant performance hacks from `idna` when the performance issues will be remedied by ICU4X 2.0.

On AMD Ryzen Threadripper PRO 5975WX, the fastest path that does not call into `idna_adapter` at all apart from its constructor is a nanosecond slower with ICU4X 1.x or 2.x compared with `unicode-rs` or `no-unicode`. In the ICU4X 2.0 case, it would be possible to put the adapter in a `static`, which logically is supposed to avoid running the constructor in the fastest-path case, but that makes it another nanosecond slower, which generally makes no sense. I didn't investigate this further, but from this, it's reasonable to guess that there are likely some issues with the code layout interacting with the instruction cache or somesuch in an unlucky way, and it's not worthwhile to try to optimize for such effects. (When there's any Unicode involved at all, ICU4X performs better than `unicode-rs`.)

Technically, `idna_adapter` and `idna_mapping` could be published to crates.io with the repos being under my GitHub account. My understanding of Servo policies is that moving the repos to under github.com/servo/ requires an OK from the TSC as a matter of formality, since the licenses are not MPL2. I have posted a [request](https://github.com/servo/project/issues/101#issuecomment-2346616613) to get such an OK when the TSC meets on September 23rd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ URL library for Rust, based on the [URL Standard](https://url.spec.whatwg.org/).
 [Documentation](https://docs.rs/url)
 
 Please see [UPGRADING.md](https://github.com/servo/rust-url/blob/main/UPGRADING.md) if you are upgrading from a previous version.
+
+## Alternative Unicode back ends
+
+`url` depends on the `idna` crate. By default, `idna` uses [ICU4X](https://github.com/unicode-org/icu4x/) as its Unicode back end. If you wish to opt for different tradeoffs between correctness, run-time performance, binary size, compile time, and MSRV, please see the [README of the latest version of the `idna_adapter` crate](https://docs.rs/crate/idna_adapter/latest) for how to opt into a different Unicode back end.
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/servo/rust-url/"
 license = "MIT OR Apache-2.0"
 autotests = false
 edition = "2018"
-rust-version = "1.67"
+rust-version = "1.57" # For panic in const context
 
 [lib]
 doctest = false
--- a/idna/Cargo.toml
+++ b/idna/Cargo.toml
@@ -17,7 +17,7 @@ doctest = false
 default = ["std", "compiled_data"]
 std = ["alloc"]
 alloc = []
-compiled_data = ["icu_normalizer/compiled_data", "icu_properties/compiled_data"]
+compiled_data = ["idna_adapter/compiled_data"]
 
 [[test]]
 name = "tests"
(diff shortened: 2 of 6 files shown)


hsivonen:
The CI failures are due to the `path` dependencies, which can go away once `idna_mapping` and `idna_adapter` are on crates.io.

hsivonen:
And the intent, of course, is to make `idna` depend on `idna_adapter` version `1` once replacing the `path` dependency.

valenting:
Looks great so far!

hsivonen:
Per Zulip regarding Servo TSC discussion about the Unicode-3.0 part here: "everyone was fine, and it uses an open source license so no issues with that"

hsivonen:
Per today's Servo TSC meeting, I'll publish idna_adapter and idna_mapping to crates.io with the repos remaining under my GitHub account at least for now.

hsivonen:
Looks like the coverage check is bad at dealing with expressions formatted over multiple lines.

hsivonen:
> Looks great so far!

Merged per out-of-bad check to interpret this comment as review.

PR #967, opened 2024-09-16
Fix debugger_visualizer test failures.

valenting:
The debugger_visualizer tests pass locally, but my cdb version is slightly newer.
Let's see if this works.


valenting:
I installed the specific version locally, and it still passes.
I marked the test as `continue-on-error` instead.

hsivonen (approved):
rs+

PR #969, opened 2024-09-19
Add AsciiSet::EMPTY and boolean operators

joshka:
In RFCs, the sets of characters to percent-encode are often defined as
the union of multiple sets. This change adds an `EMPTY` constant to
`AsciiSet` and implements the `Add` trait for `AsciiSet` so that sets
can be combined with the `+` operator. The ! negation operator is also
defined, as well as equivalent constant functions for these (`union()`,
`complement()`).

AsciiSet now derives `Debug`, `PartialEq`, and `Eq` so that it can be
used in tests.

---

Example: https://www.rfc-editor.org/rfc/rfc3986#section-3.4 defines
```
   reserved      = gen-delims / sub-delims
   gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

Using this new method, this can be easily represented as:
```rust
const SUB_DELIMS: AsciiSet = AsciiSet::EMPTY.add(b'!').add(b'$') ...
const GEN_DELIMS: AsciiSet = AsciiSet::EMPTY.add(b':').add(b'/') ...
const RESERVED: AsciiSet = GEN_DELIMS.union(SUB_DELIMS);
```

Similarly the set of characters that must be encoded is defined as the set of characters that are not in the allowed characters

https://www.rfc-editor.org/rfc/rfc3986#section-2.2

>    URI producing applications should percent-encode data octets that
   correspond to characters in the reserved set **unless these characters
   are specifically allowed by the URI scheme to represent data in that
   component.**  If a reserved character is found in a URI component and
   no delimiting role is known for that character, then it must be
   interpreted as representing the data octet corresponding to that
   character's encoding in US-ASCII.

So a part like query is defined in https://www.rfc-editor.org/rfc/rfc3986#appendix-A as:
```
   query         = *( pchar / "/" / "?" )
   pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
   unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
   pct-encoded   = "%" HEXDIG HEXDIG
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```
which can be translated to:
```rust
const QUERY: AsciiSet = PCHAR.add(b'/').add(b'?');
const PCHAR: AsciiSet = UNRESERVED.union(SUB_DELIMS).add(b':').add(b'@');
const UNRESERVED: AsciiSet = ...

// which can be then used like:
let encoded_query = utf8_percent_encode("foo?:@/=bar", !QUERY);
```

### DIFF
--- a/percent_encoding/src/lib.rs
+++ b/percent_encoding/src/lib.rs
@@ -120,6 +120,15 @@ impl ops::Add for AsciiSet {
     }
 }
 
+impl ops::Not for AsciiSet {
+    type Output = Self;
+
+    fn not(self) -> Self {
+        let mask = self.mask.map(|chunk| !chunk);
+        AsciiSet { mask }
+    }
+}
+
 /// The set of 0x00 to 0x1F (C0 controls), and 0x7F (DEL).
 ///
 /// Note that this includes the newline and tab characters, but not the space 0x20.
@@ -507,5 +516,13 @@ mod tests {
         let right = AsciiSet::EMPTY.add(b'B');
         let expected = AsciiSet::EMPTY.add(b'A').add(b'B');
         assert_eq!(left + right, expected);
+    }
+
+    #[test]
+    fn not() {
+        let set = AsciiSet::EMPTY.add(b'A').add(b'B');
+        let not_set = !set;
+        assert!(!not_set.contains(b'A'));
+        assert!(not_set.contains(b'C'));
     }
 }
--- a/percent_encoding/src/lib.rs
+++ b/percent_encoding/src/lib.rs
@@ -106,17 +106,30 @@ impl AsciiSet {
         mask[byte as usize / BITS_PER_CHUNK] &= !(1 << (byte as usize % BITS_PER_CHUNK));
         AsciiSet { mask }
     }
+
+    /// Return the union of two sets.
+    pub const fn union(&self, other: Self) -> Self {
+        let mask = [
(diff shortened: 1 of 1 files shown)


valenting (approved):

joshka:
Thanks!

percent_encoding/src/lib.rs
```diff
@@ -77,6 +78,11 @@ const ASCII_RANGE_LEN: usize = 0x80;
 const BITS_PER_CHUNK: usize = 8 * mem::size_of::<Chunk>();
 
 impl AsciiSet {
+    /// An empty set.
+    pub const EMPTY: AsciiSet = AsciiSet {
```
ForsakenHarmony, line 82: This seems like it's now inconsistent with the existing consts being and fucntion taking `&'static AsciiSet`.

joshka: Yeah, I wasn't 100% sure about that. I went with EMPTY being a constant on the AsciiSet as empty seems like an inherent property of a type, but the other constants seem like usages of AsciiSet. The rationale for making the constants references rather than just values all seemed odd to me. What was that necessary for?

PR #837, opened 2023-05-07
Additional percent-encode sets

cryptoquick:
I've added some additional constants that should be useful for others.

I'd also recommend adding #721.

PR #971, opened 2024-09-19
feat(percent-encoding): add RFC2396 ascii sets

ForsakenHarmony:
I know the doc comment says that you intentionally don't have a lot of sets, but I feel like common ones should probably be supported if anyone using the library has to add them.

### DIFF
--- a/percent_encoding/src/lib.rs
+++ b/percent_encoding/src/lib.rs
@@ -66,7 +66,7 @@ use core::{fmt, mem, ops, slice, str};
 /// /// https://url.spec.whatwg.org/#fragment-percent-encode-set
 /// const FRAGMENT: &AsciiSet = &CONTROLS.add(b' ').add(b'"').add(b'<').add(b'>').add(b'`');
 /// ```
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct AsciiSet {
     mask: [Chunk; ASCII_RANGE_LEN / BITS_PER_CHUNK],
 }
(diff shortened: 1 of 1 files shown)


percent_encoding/src/lib.rs
```diff
@@ -101,14 +101,26 @@ impl AsciiSet {
         AsciiSet { mask }
     }
 
+    pub const fn add_range(&self, start: u8, end: u8) -> Self {
```
joshka, line 104: If this took [Range](https://doc.rust-lang.org/std/ops/struct.Range.html) instead of two vars, then it would be more obvious that this was the half-open range (it's actually the inclusive range, but that's not obvious from the method name at all)

percent_encoding/src/lib.rs
```diff
@@ -176,40 +204,100 @@ static_assert! {
...
-    .add(b'*')
-    .add(b'+')
-    .add(b',')
-    .add(b'-')
-    .add(b'.')
-    .add(b'/')
-    .add(b':')
+pub const NON_ALPHANUMERIC: &AsciiSet = &ALPHA_NUM.complement();
+
+/// [RFC2396](https://datatracker.ietf.org/doc/html/rfc2396).
```
joshka, line 209: non-blocking nit:

```suggestion
/// Lower case alpha numeric characters defined in [RFC2396](https://datatracker.ietf.org/doc/html/rfc2396#section-1.6).
```
A good description like this helps users searching for things (e.g. with a search engine), and the direct link to the section that these are defined in the RFC is useful when you need to dig into various things (obviously it's less useful for this particular case, but it's generally useful to have this as a general thing I think).

joshka (commented):
I'd suggest putting the add_range method and the transitions of the existing code to use it into its own PR, which can likely be merged fairly non-controversially as a good simplification.

I think it would be worth it to pull these out into a specific module (or possibly a crate) named after the RFC. There's another PR that's open to add the whatwg url definitions which are subtly different from the RFC definitions. https://github.com/servo/rust-url/pull/837. It would be good to have both these available for different circumstances. The RFC defined values are good for things defined by RFCs (backend / services etc.), while the whatwg versions are better for things more frontend / http focused.

BTW 2396 is updated by https://www.rfc-editor.org/rfc/rfc3986. The reason I came to look at this crate in the first place is that I wanted a spec compliant way to write a percent encoded query string for implementing another RFC which doesn't currently have a crate.

PR #986, opened 2024-10-29
Test MSRV with idna unicode-rs back end and test ICU4X back end with 1.67

hsivonen:
Add idna_adapter awareness to CI. (The no-unicode branch of idna_adapter does not make sense to test here, since it obviously fails normal tests.)


hsivonen:
Is it excessive to run the MSRV (1.57.0) on Mac and Windows? Should those be excluded to make the MSRV test Ubuntu-only?

Manishearth (approved):

PR #987, opened 2024-10-29
Prepare url crate for publication with idna 1.0.3

hsivonen:
Prepare the idna-v1x branch so that `url` 2.5.3 depending on `idna` 1.0.3 could be published from it.

However, instead of actually publishing `url` 2.5.3 and `idna` 1.0.3 from the idna-v1x branch, my expectation is that after this PR has landed, we'd merge idna-v1x into main, at which point `url` 2.5.3 and `idna` 1.0.3 could be published from main and the idna-v1x branch would become obsolete.

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "2.5.2"
+version = "2.5.3"
 authors = ["The rust-url developers"]
 
 description = "URL library for Rust, based on the WHATWG URL Standard"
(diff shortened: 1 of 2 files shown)


url/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 
 name = "url"
 # When updating version, also modify html_root_url in the lib.rs
-version = "2.5.2"
+version = "2.5.3"
```
valenting, line 5: Make sure to also update the version [here](https://github.com/servo/rust-url/blob/08a326820a4a54d6a9730ecabd590942fc6b66dc/url/src/lib.rs#L146)

valenting (approved):

url/Cargo.toml
hsivonen: Thanks. Done.

PR #991, opened 2024-11-04
Change no_std to no-std in Cargo.toml

valenting:
Otherwise I get the following error when attempting to publish the crate:
```
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 400 Bad Request): The following category slugs are not currently supported on crates.io: no_std

  See https://crates.io/category_slugs for a list of supported slugs.
```


hsivonen (approved):
LGTM. Thanks!

Issue #984, opened 2024-10-21
Mutating path of URL without authority (idempotence, empty path segments)


x11x:
I have found a couple more cases related to this (https://github.com/whatwg/url/pull/505; number 3. under [4.5. URL Serializing](https://url.spec.whatwg.org/#url-serializing)).

It occurs when mutating the path of a non-"_cannot-be-a-base_" URL without an authority.

(note that `"web+demo:/"` is *not* "_cannot-be-a-base_" because path is `"/"`).

```rust
use url::Url;

#[test]
fn test_can_be_a_base_with_set_path() {
    let mut url = Url::parse("web+demo:/").unwrap();
    assert!(!url.cannot_be_a_base());

    // Set path to "//not-a-host" using `set_path`
    url.set_path("//not-a-host");

    // PASSES (path is correctly set):
    assert_eq!(url.path(), "//not-a-host");

    // PASSES (has segments):
    let segments: Vec<_> = url.path_segments().expect("should have path segments").collect();
    // PASSES:
    assert_eq!(segments, vec!["", "not-a-host"]);

    // **FAILS**
    // EXPECTED: "web+demo:/.//not-a-host"
    // ACTUAL:   "web+demo://not-a-host"
    assert_eq!(url.as_str(), "web+demo:/.//not-a-host");
}

#[test]
fn test_can_be_a_base_with_path_segments_mut() {
    let mut url = Url::parse("web+demo:/").unwrap();
    assert!(!url.cannot_be_a_base());

    // Set path to "//not-a-host" using `path_segments_mut`
    url.path_segments_mut()
        .expect("should have path segments")
        .push("")  /* NOTE: any number of push("") here appears to make no difference (all ignored) */
        .push("not-a-host");

    // **FAILS**
    // EXPECTED: "web+demo:/.//not-a-host"
    // ACTUAL:   "web+demo:/not-a-host"
    assert_eq!(url.as_str(), "web+demo:/.//not-a-host");

    // **FAILS**
    // EXPECTED: "//not-a-host"
    // ACTUAL:   "/not-a-host"
    assert_eq!(url.path(), "//not-a-host");

    // PASSES (has segments):
    let segments: Vec<_> = url.path_segments().expect("should have path segments").collect();

    // **FAILS**
    // EXPECTED: ["", "not-a-host"]
    // ACTUAL:   ["not-a-host"]
    assert_eq!(segments, vec!["", "not-a-host"]);
}
```
[...]

PR #996, opened 2024-11-08
Fix mutating path of URL without authority (idempotency, empty path segments)

theskim:
From Issue #984


url/tests/unit.rs
```diff
@@ -1379,3 +1379,46 @@ fn serde_error_message() {
...
+        .expect("should have path segments")
+        .collect();
+
+    assert_eq!(segments, vec!["", "not-a-host"]);
+
+    assert_eq!(url.as_str(), "web+demo:/.//not-a-host");
+    quirks::set_hostname(&mut url, "test").unwrap();
+    assert_eq!(url.as_str(), "web+demo://test//not-a-host");
+    quirks::set_hostname(&mut url, "").unwrap();
+    assert_eq!(url.as_str(), "web+demo:/.//not-a-host");
```
valenting, line 1403: Actually, I was wrong about this. We are setting the host to "", not null (in fact you can't set it to null again).
So the result should be `web+demo:////not-a-host`

I've checked this with the reference URL parser
```
// Go to https://jsdom.github.io/whatwg-url/
// Open the devtools console and type in all the » lines
» import('./whatwg-url.mjs').then(m => whatwg = m.default)
Promise { <state>: "fulfilled", <value>: {…} }

» u = new whatwg.URL("web+demo:/")
Object { … }

» u.href
"web+demo:/"
» u.pathname = "//not-a-host"
"//not-a-host"
» u.href
"web+demo:/.//not-a-host"
» u.host = "host"
"host"
» u.href
"web+demo://host//not-a-host"
» u.host = ""
""
» u.href
"web+demo:////not-a-host" 
```

url/src/lib.rs
```diff
@@ -2143,6 +2143,23 @@ impl Url {
         self.host_end = to_u32(self.serialization.len()).unwrap();
         self.host = host.into();
 
+        // Adjust serialization to switch between host and empty segment
+        if suffix.starts_with("/.//") {
+            suffix = suffix["/.".len()..].to_string();
```
valenting, line 2148:
```suggestion
            suffix.drain(.."/.".len());
```

I think this should avoid an unnecessary allocation.

valenting (changes requested):
Sorry for the delayed review. See inline:

url/src/lib.rs
```diff
@@ -2143,6 +2143,13 @@ impl Url {
         self.host_end = to_u32(self.serialization.len()).unwrap();
         self.host = host.into();
 
+        // Adjust serialization to switch from host to empty segment
+        if suffix.starts_with("/.//") {
+            suffix.drain(.."/.".len());
+            // pathname should be "//p" not "p" given that the first segment was empty
+            self.path_start -= "//".len() as u32;
+        }
```
theskim, line 2151: `else if` part is no longer necessary since we just change the host to empty.

theskim:
@valenting, should I be worried about this failure?

valenting (approved):

valenting:
Note: I think we need to add https://github.com/servo/rust-url/pull/943 to this PR considering we backed it out in #999

theskim:
Need to add tests to check  `scheme_end`, `username_end`, `host_start`, etc.

theskim:
@valenting, thoughts on this patch? As you can see, now we actually check the position of host, ref, etc.

url/src/path_segments.rs
```diff
@@ -253,7 +253,44 @@ impl PathSegmentsMut<'_> {
...
+                };
+
+                // For cases where normalization is applied across both the serialization and the path.
+                // Append "/." immediately after the scheme (up to ":")
+                // This is done if three conditions are met.
+                // https://url.spec.whatwg.org/#url-serializing
+                // 1. The host is null
+                // 2. The url's path length is greater than 1
+                // 3. the first segment of the URL's path is an empty string
+                if !has_host && segment.len() > 1 && path_empty {
```
valenting, line 282:
```suggestion
                if !has_host && self.path().starts_with("//") {
```

I think the entire If can be simplified to this?

PR #1012, opened 2024-12-23
simplify Iterator::next for PercentEncode

hkBst:
This reduces duplication including of an unsafe block...


hkBst:
I'm a bit puzzled by the report of a leaked allocation. I'm pretty sure I did not introduce that, but I could be wrong.

PR #1014, opened 2024-12-31
faster byte_serialized_unchanged

dev-ardi:
This uses a precomputed table.
This abuses the neat trick that ascii fits inside a u128.


Manishearth (approved):

dev-ardi:
Oh, this requires bumping the msrv

hsivonen:
It should be easy to write this as using a `const` function, which should make this work at least all the way back to Rust 1.57.

See https://github.com/servo/rust-url/blob/8a683ff4ab10498d619fd8206b64402de19474be/idna/src/uts46.rs#L72 and https://github.com/servo/rust-url/blob/8a683ff4ab10498d619fd8206b64402de19474be/idna/src/uts46.rs#L85

dev-ardi:
oops i forgor :skull:

form_urlencoded/src/lib.rs
```diff
@@ -128,8 +128,26 @@ pub struct ByteSerialize<'a> {
...
+            as u128)
+            << c;
+        c += 1;
+    }
+    magic
+}
+const GLYPHLESS_MASK: u128 = glyphless_mask();
+
+#[inline]
+pub fn byte_serialized_unchanged(byte: u8) -> bool {
```
dsherret, line 146: Should this remain private?

lucacasonato (approved):
LGTM

PR #1018, opened 2025-01-27
perf: improve to_file_path()

dsherret:
Reduces some heap allocations and pre-allocates the vector for building the string.

### DIFF
--- a/url/src/lib.rs
+++ b/url/src/lib.rs
@@ -2722,7 +2722,7 @@ impl Url {
                 _ => return Err(()),
             };
 
-            return file_url_segments_to_pathbuf(host, segments);
+            return file_url_segments_to_pathbuf(self.as_str().len(), host, segments);
         }
         Err(())
     }
@@ -3032,6 +3032,7 @@ fn path_to_file_url_segments_windows(
     any(unix, target_os = "redox", target_os = "wasi", target_os = "hermit")
 ))]
 fn file_url_segments_to_pathbuf(
+    estimated_capacity: usize,
     host: Option<&str>,
     segments: str::Split<'_, char>,
 ) -> Result<PathBuf, ()> {
@@ -3049,11 +3050,10 @@ fn file_url_segments_to_pathbuf(
         return Err(());
     }
 
-    let mut bytes = if cfg!(target_os = "redox") {
-        b"file:".to_vec()
-    } else {
-        Vec::new()
-    };
+    let mut bytes = Vec::with_capacity(estimated_capacity);
+    if cfg!(target_os = "redox") {
+        bytes.extend(b"file:");
+    }
 
     for segment in segments {
         bytes.push(b'/');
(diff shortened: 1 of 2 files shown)


url/src/lib.rs
```diff
@@ -3049,11 +3050,10 @@ fn file_url_segments_to_pathbuf(
         return Err(());
     }
 
-    let mut bytes = if cfg!(target_os = "redox") {
-        b"file:".to_vec()
-    } else {
-        Vec::new()
-    };
+    let mut bytes = Vec::with_capacity(estimated_capacity);
```
dsherret, line 3053: Before 1,000 iterations:

> test url_to_file_path ... bench:     202,125 ns/iter (+/- 12,207)

After:

> test url_to_file_path ... bench:     127,257 ns/iter (+/- 4,782)

url/src/lib.rs
```diff
@@ -3085,22 +3085,26 @@ fn file_url_segments_to_pathbuf(
...
     segments: str::Split<char>,
 ) -> Result<PathBuf, ()> {
-    file_url_segments_to_pathbuf_windows(host, segments)
+    file_url_segments_to_pathbuf_windows(estimated_capacity, host, segments)
 }
 
 // Build this unconditionally to alleviate https://github.com/servo/rust-url/issues/102
 #[cfg(feature = "std")]
 #[cfg_attr(not(windows), allow(dead_code))]
 fn file_url_segments_to_pathbuf_windows(
```
dsherret, line 3098: Before 1000 iterations (within an iteration):

> test url_to_file_path ... bench:     437,555 ns/iter (+/- 11,519)

After:

> test url_to_file_path ... bench:     119,461 ns/iter (+/- 5,927)

url/src/lib.rs
```diff
@@ -3085,22 +3085,26 @@ fn file_url_segments_to_pathbuf(
...
     host: Option<&str>,
     mut segments: str::Split<'_, char>,
 ) -> Result<PathBuf, ()> {
-    use percent_encoding::percent_decode;
-    let mut string = if let Some(host) = host {
-        r"\\".to_owned() + host
+    use percent_encoding::percent_decode_str;
+    let mut string = String::with_capacity(estimated_capacity);
+    if let Some(host) = host {
+        string.push('\\');
```
dsherret, line 3106: oops. bug

dsherret:
Opened https://github.com/servo/rust-url/pull/1019 for the CI.

url/src/lib.rs
```diff
@@ -2720,7 +2720,7 @@ impl Url {
                 _ => return Err(()),
             };
 
-            return file_url_segments_to_pathbuf(host, segments);
+            return file_url_segments_to_pathbuf(self.as_str().len(), host, segments);
```
lucacasonato, line 2723: Please split the `estimated_len` into its own temp variable so it's cleaner to read what this is.

I think this over-alloc 7 bytes in 99% of cases on unix (`file://`). On redox, just two (`//`). On Windows, even 8 (`file:///`). Windows also has `hostname` support though, so to ensure no-realloc in most cases, you should probably do -6 there too (`file:/` - the extra `/` is for the leading `\` in hostname cases).

The cost of doing this calculation may be too much perf cost though - so please check if this minor memory improvement wouldn't regress perf too much.

lucacasonato (approved):
LGTM

url/src/lib.rs
dsherret: It adds a few nanoseconds. Worth it to use less memory IMO.

lucacasonato: Agreed.

url/src/lib.rs
```diff
@@ -2720,7 +2720,18 @@ impl Url {
                 _ => return Err(()),
             };
 
-            return file_url_segments_to_pathbuf(host, segments);
+            let estimated_capacity = self.as_str().len()
+                - if cfg!(target_os = "redox") {
```
lucacasonato, line 2724: Can you make this a clamped subtract so that it doesn't wrap around in release when someone calls this with something other than `file` scheme URLs?

dsherret: Good catch. I think the logic needs to be a bit smarter here because urls could be like `a:/path`? I updated the code to at least be a fast path for `file://`-like urls and maybe in some weird cases like `a:/path` it might under allocate. Probalby not worth the effort to make it smarter than it is now.

url/src/lib.rs
```diff
@@ -2720,17 +2720,25 @@ impl Url {
...
-                } else if cfg!(windows) {
-                    // remove file: - has posssible \\ for hostname
-                    5
+            let str_len = self.as_str().len();
+            let estimated_capacity = if cfg!(target_os = "redox") {
+                let scheme_len = self.scheme().len();
+                let file_scheme_len = "file".len();
+                // remove only // because it still has file:
+                if scheme_len < file_scheme_len {
+                    let scheme_diff = file_scheme_len - scheme_len;
```
lucacasonato, line 2729: This still wraps around when `scheme` is longer than 4.

dsherret: There's a check above for this:

```
if scheme_len < file_scheme_len { // if 5 < 4 {
```

lucacasonato (approved):
LGTM, very nice improvement.

PR #1020, opened 2025-02-06
perf: remove heap allocation in parse_query

dsherret:
Also updates parse_fragment to work similarly to how parse_query works now where it operates on longer string slices for percent decoding, which helps to not have to do that for every character.

Before:

```
test long           ... bench:         302 ns/iter (+/- 42) = 142 MB/s
test fragment       ... bench:         370 ns/iter (+/- 9) = 140 MB/s
```

After:

```
test long           ... bench:         243 ns/iter (+/- 7) = 176 MB/s
test fragment       ... bench:         275 ns/iter (+/- 9) = 189 MB/s
```

### DIFF
--- a/url/src/parser.rs
+++ b/url/src/parser.rs
@@ -73,6 +73,12 @@ macro_rules! simple_enum_error {
     }
 }
 
+macro_rules! ascii_tab_or_new_line_pattern {
+    () => {
+        '\t' | '\n' | '\r'
+    };
+}
+
 #[cfg(feature = "std")]
 impl std::error::Error for ParseError {}
 
@@ -207,7 +213,7 @@ impl<'i> Input<'i> {
             if input.len() < original_input.len() {
                 vfn(SyntaxViolation::C0SpaceIgnored)
             }
-            if input.chars().any(|c| matches!(c, '\t' | '\n' | '\r')) {
+            if input.chars().any(ascii_tab_or_new_line) {
                 vfn(SyntaxViolation::TabOrNewlineIgnored)
             }
         }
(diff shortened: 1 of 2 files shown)


url/src/parser.rs
```diff
@@ -1473,37 +1473,81 @@ impl<'a> Parser<'a> {
...
+        impl<'i> Iterator for QueryPartIter<'i, '_> {
+            type Item = (&'i str, bool);
+
+            fn next(&mut self) -> Option<Self::Item> {
+                let start = self.input.chars.as_str();
+                // bypass self.input.next() in order to get string slices
+                // which are faster to operate on
+                while let Some(c) = self.input.chars.next() {
+                    match c {
+                        '\t' | '\n' | '\r' => {
```
lucacasonato, line 1493: Can you extract this pattern to be shared with `self.input.next()` so that it is clear if this ever gets updated?

url/src/parser.rs
```diff
@@ -1526,31 +1570,75 @@ impl<'a> Parser<'a> {
...
+        impl<'i> Iterator for FragmentPartIter<'i, '_> {
+            type Item = &'i str;
+
+            fn next(&mut self) -> Option<Self::Item> {
+                let start = self.input.chars.as_str();
+                // bypass self.input.next() in order to get string slices
+                // which are faster to operate on
+                while let Some(c) = self.input.chars.next() {
+                    match c {
+                        '\t' | '\n' | '\r' => {
```
lucacasonato, line 1588: Ditto

lucacasonato (commented):
Great!

url/src/parser.rs
dsherret: Makes it 10ns slower even with `#[inline]` for some reason (my guess is that causes it to lose out on some optimization in the match). It seems this pattern is used in a lot of places.

lucacasonato: You can do
```
macro_rules! some_name {
  () => '\t' | '\n' | '\r'
}
```

and then use `some_name!()` instead of the pattern here. That should have no runtime cost.

dsherret: Thanks! That worked.

lucacasonato (approved):
LGTM, thanks!

PR #1021, opened 2025-02-06
perf: remove heap allocation in parse_host

dsherret:
Before:

```
test long           ... bench:         302 ns/iter (+/- 42) = 142 MB/s
```

After:

```
test long           ... bench:         271 ns/iter (+/- 7) = 158 MB/s
```

### DIFF
--- a/idna/src/lib.rs
+++ b/idna/src/lib.rs
@@ -99,7 +99,7 @@ impl core::fmt::Display for Errors {
 ///
 /// This process may fail.
 ///
-/// If you have a `&str` instead of `&[u8]`, just call `.to_bytes()` on it before
+/// If you have a `&str` instead of `&[u8]`, just call `.as_bytes()` on it before
 /// passing it to this function. It's still preferable to use this function over
 /// the sibling functions that take `&str`.
 pub fn domain_to_ascii_cow(
--- a/idna/src/lib.rs
+++ b/idna/src/lib.rs
@@ -129,7 +129,10 @@ pub fn domain_to_ascii_cow(
 /// and using Punycode as necessary.
 ///
 /// This process may fail.
-pub fn domain_to_ascii_from_cow(domain: Cow<'_, [u8]>, ascii_deny_list: AsciiDenyList) -> Result<Cow<'_, str>, Errors> {
+pub fn domain_to_ascii_from_cow(
+    domain: Cow<'_, [u8]>,
+    ascii_deny_list: AsciiDenyList,
+) -> Result<Cow<'_, str>, Errors> {
     Uts46::new().to_ascii_from_cow(
         domain,
         ascii_deny_list,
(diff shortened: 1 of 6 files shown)


url/src/host.rs
```diff
@@ -80,15 +79,34 @@ impl Host<String> {
     ///
     /// <https://url.spec.whatwg.org/#host-parsing>
     pub fn parse(input: &str) -> Result<Self, ParseError> {
```
dsherret, line 81: I didn't  add, change, or remove from the public API, but please verify this as well in case I missed something.

url/src/host.rs
```diff
@@ -137,17 +154,39 @@ impl Host<String> {
...
 }
 
+/// convert domain with idna
+fn domain_to_ascii(domain: Cow<'_, [u8]>) -> Result<Cow<'_, str>, ParseError> {
+    let value = idna::domain_to_ascii_cow(&domain, idna::AsciiDenyList::URL)?;
+    // TODO: would be better to move this into the idna crate
+    Ok(match value {
+        Cow::Owned(value) => Cow::Owned(value),
+        // SAFETY: If borrowed, then the original string is ascii and we can return the
+        // original Cow in order to save an allocation
```
lucacasonato, line 182: My reading of the code suggests this is correct right now, but:

There is no guarantee that `idna::domain_to_ascii_cow` could not return a `Cow::Borrowed(x)` where `x.len() < domain.len()` in the future (where this code would break or possibly be unsafe). To avoid that refactoring hazard, we should hoist this up into `idna`.

dsherret: Hoisted. Not to sure about the new function name.

idna/src/lib.rs
```diff
@@ -114,6 +114,33 @@ pub fn domain_to_ascii_cow(
...
+/// perform the [forbidden domain code point](https://url.spec.whatwg.org/#forbidden-domain-code-point)
+/// check in addition to the [domain to ASCII](https://url.spec.whatwg.org/#concept-domain-to-ascii)
+/// algorithm.
+///
+/// Return the ASCII representation a domain name,
+/// normalizing characters (upper-case to lower-case and other kinds of equivalence)
+/// and using Punycode as necessary.
+///
+/// This process may fail.
+pub fn domain_to_ascii_from_cow(
```
lucacasonato, line 132: @hsivonen Please check if you find this new public API of `idna` to be acceptable.

lucacasonato: gentle ping @hsivonen

dsherret: Since it's been several months, maybe this can be merged and then deprecated later if it's not the right approach?

lucacasonato (approved):
LGTM

Considering the code looks good, the perf gain is meaningful, and no-one has objected to the new API surface in 3 months, I am going to land this now.

PR #1022, opened 2025-02-06
perf: slightly improve parsing a port

dsherret:
Before:

```
test port           ... bench:         222 ns/iter (+/- 5) = 108 MB/s
```

After:

```
test port           ... bench:         211 ns/iter (+/- 7) = 113 MB/s
```


lucacasonato (approved):
LGTM. The safety comment on the `unsafe` is sound.

PR #1026, opened 2025-02-07
perf: url encode path segments in longer string slices

dsherret:
Before:

```
test long           ... bench:         250 ns/iter (+/- 11) = 172 MB/s
```

After:

```
test long           ... bench:         221 ns/iter (+/- 24) = 248 MB/s
```

### DIFF
--- a/url/benches/parse_url.rs
+++ b/url/benches/parse_url.rs
@@ -13,7 +13,7 @@ fn short(bench: &mut Bencher) {
 }
 
 fn long(bench: &mut Bencher) {
-    let url = "https://example.com/parkbench?tre=es&st=uff";
+    let url = "https://example.com/parkbench?tre=es&st=ufflongerlonger";
 
     bench.bytes = url.len() as u64;
     bench.iter(|| black_box(url).parse::<Url>().unwrap());
--- a/url/src/parser.rs
+++ b/url/src/parser.rs
@@ -1191,19 +1191,67 @@ impl<'a> Parser<'a> {
         path_start: usize,
         mut input: Input<'i>,
     ) -> Input<'i> {
+        // it's much faster to call utf8_percent_encode in bulk
+        fn push_pending(
+            serialization: &mut String,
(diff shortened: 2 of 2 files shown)


url/src/parser.rs
```diff
@@ -1305,7 +1367,7 @@ impl<'a> Parser<'a> {
             }
         }
         if scheme_type.is_file() {
-            // while url’s path’s size is greater than 1
+            // while url's path’s size is greater than 1
             // and url’s path[0] is the empty string,
```
Scripter17, line 1371: Missed a `

Scripter17:
It fully occurred to me how annoying it must be to get a "you missed a smudge" review from a rando on your genuinely very impressive work so I want to redeem myself: You're awesome

URL parsing has been a bit of a bottleneck in my current main project for a while and all your work does a pretty good dent in it

Seeing all your pull requests was a very nice thing to wake up and look forward to

url/src/parser.rs
dsherret: Oops. I missed reverting that after I noticed that kind of quote is used everywhere in the codebase (habit I have because in Deno, we avoid using non-ascii characters in JS because it causes the source code to be stored less efficiently. Example: https://github.com/microsoft/TypeScript/pull/53608). Thanks!

url/benches/parse_url.rs
```diff
@@ -13,7 +13,7 @@ fn short(bench: &mut Bencher) {
 }
 
 fn long(bench: &mut Bencher) {
-    let url = "https://example.com/parkbench?tre=es&st=uff";
+    let url = "https://example.com/parkbench?tre=es&st=ufflongerlonger";
```
dsherret, line 16:
```suggestion
    let url = "https://example.com/parkbench?tre=es&st=uff";
```

dsherret:
Not annoying. Thanks!

lucacasonato (approved):
LGTM

PR #1060, opened 2025-08-06
set_hostname should error when encountering colon ':'

edgul:
When setting via hostname api, the spec notes [in the basic url parser](https://url.spec.whatwg.org/#hostname-state) that when we encounter a colon ":":

> If state override is given and state override is [hostname state](https://url.spec.whatwg.org/#hostname-state), then return failure.

### DIFF
--- a/url/src/quirks.rs
+++ b/url/src/quirks.rs
@@ -212,7 +212,10 @@ pub fn set_hostname(url: &mut Url, new_hostname: &str) -> Result<(), ()> {
         return Ok(());
     }
 
-    if let Ok((host, _remaining)) = Parser::parse_host(input, scheme_type) {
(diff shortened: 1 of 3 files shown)


url/src/quirks.rs
```diff
@@ -212,7 +212,10 @@ pub fn set_hostname(url: &mut Url, new_hostname: &str) -> Result<(), ()> {
         return Ok(());
     }
 
-    if let Ok((host, _remaining)) = Parser::parse_host(input, scheme_type) {
+    if let Ok((host, remaining)) = Parser::parse_host(input, scheme_type) {
+        if remaining.split_prefix(':').is_some() {
```
valenting, line 216:
```suggestion
        if remaining.starts_with(':') {
```

I think it's much clearer to use starts_with instead of split_prefix.

valenting (approved):

PR #1064, opened 2025-08-21
Use no-std instead of no_std in Cargo.toml

valenting:
Otherwise cargo publish complains the category isn't allowed.


hsivonen (approved):
LGTM

Manishearth (approved):

PR #1065, opened 2025-08-21
version bump to 2.5.6

edgul:

PR #1067, opened 2025-08-21
Fix: bump idna dependency

BenjaminBrienen:


BenjaminBrienen:
already fixed in #1065

PR #1075, opened 2025-09-10
Improve Compile Times with `serde_derive`

CryZe:
This massively improves compile times by not forcing `serde` to depend on `serde_derive`. This allows many crates to start compiling much sooner, which ends up resulting in often quite massive compile time wins for real world projects.

In my case this PR reduced the build times of my project (clean, release) from 91.4s to 76.0s.

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -17,7 +17,8 @@ edition = "2018"
 rust-version = "1.63" # From libc
 
 [dev-dependencies]
-serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


CryZe:
Arguably the entire derive dependency could be removed by replacing it with manual implementations, but it probably almost never would improve compile times.

CryZe:
Seems like #1071 is related, as that bug is the only reason why serde (and transitively serde_derive) are even being pulled in.

Manishearth (approved):

PR #1120, opened 2026-05-05
make idna an optional dependency

BenjaminBrienen:
For use cases where domain names are already ascii, the compile-time hit of the idna dependency and the unnecessary processing can be avoided by disabling the optional feature

### DIFF
--- a/url/Cargo.toml
+++ b/url/Cargo.toml
@@ -27,14 +27,14 @@ wasm-bindgen-test = "0.3"
 
 [dependencies]
 form_urlencoded = { version = "1.2.2", path = "../form_urlencoded", default-features = false, features = ["alloc"] }
-idna = { version = "1.1.0", path = "../idna", default-features = false, features = ["alloc", "compiled_data"] }
+idna = { version = "1.1.0", optional = true, path = "../idna", default-features = false, features = ["alloc", "compiled_data"] }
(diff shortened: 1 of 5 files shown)


hsivonen:
Surely the compile time impact of `idna` itself is minimal if you turn off IDNA using the [documented mechanism](https://docs.rs/crate/idna_adapter/latest) liked from the README of this crate.

BenjaminBrienen:
My bad

BenjaminBrienen:
This seems cursed. Now I have to remember to run `cargo update -p idna_adapter --precise 1.0.0` every time I want to `cargo update`. Adding `idna_adapter = { version = "=1.0.0" }` to my `workspace.dependencies` does not pin the version. Cargo is built around semantic versioning and features.

edit: If I add `idna_adapter` as a dependency, it looks like an unused dependency to tools such as `cargo-shear`. I guess this could be considered a false positive.

hsivonen:
> Cargo is built around semantic versioning and features.

And that is insufficient for decisions that should be made at the top level of compilation. Like it says in the docs, the current situation is a workaround for https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618 not being available, yet.

As for the ergonomics, you are asking for "please do not comply with the spec for the sake of build times". It's somewhat reasonable for "please do not comply with the spec" not to have great ergonomics. OTOH, it would not great if `default-feature =false` resulted in not complying with the spec.

PR #1128, opened 2026-06-05
Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, since `self` is not used later, the clone can be safely removed.

CC @nunoplopes


FranciscoTGouveia:
Looking at the logs, I believe this was a spurious failure in the CI.

valenting (approved):

PR #1131, opened 2026-06-22
Fix rustc 1.82.0 CI

valenting:


Manishearth (approved):
sigh. once we're on a newer MSRV cargo will automatically handle this for us

PR #1142, opened 2026-07-12
perf: speed up URL parsing and IDNA ToASCII

geeknoid:
Avoid redundant work on the URL parsing hot path:

- host: only percent-decode when the input contains a `%`, and append domain hosts to the serialization with `push_str` instead of `write!`.
- parser: skip the userinfo scan when the authority has no `@`, and advance the input iterator by the known byte length after scanning a host rather than re-iterating it character by character.
- idna: pass ASCII labels with a leading digit through the fast path when the domain cannot be a bidi domain name.

Add Callgrind (gungraun) instruction-count benchmarks alongside the existing wall-clock ones.

| benchmark       | instr orig | instr new |   Δ%    |  ns orig |   ns new |   ns Δ% |
|-----------------|-----------:|----------:|--------:|---------:|---------:|--------:|
| plain           |       2403 |      1823 |  -24.1% |      209 |      168 |  -19.6% |
| short           |       2653 |      2074 |  -21.8% |      243 |      186 |  -23.5% |
| long            |       3517 |      2935 |  -16.5% |      306 |      253 |  -17.3% |
| fragment        |       3887 |      3306 |  -14.9% |      344 |      285 |  -17.2% |
| port            |       2965 |      2318 |  -21.8% |      321 |      264 |  -17.8% |
| hyphen          |       3573 |      2607 |  -27.0% |      307 |      235 |  -23.5% |
| leading_digit   |       3674 |      2315 |  -37.0% |      315 |      211 |  -33.0% |
| unicode_mixed   |       6791 |      6128 |   -9.8% |      729 |      678 |   -7.0% |
| punycode_mixed  |       6105 |      5289 |  -13.4% |      607 |      498 |  -18.0% |
| unicode_ltr     |       8379 |      7609 |   -9.2% |      847 |      788 |   -7.0% |
| punycode_ltr    |       8265 |      7249 |  -12.3% |      793 |      692 |  -12.7% |
| unicode_rtl     |       8547 |      7853 |   -8.1% |      834 |      766 |   -8.2% |
| punycode_rtl    |       8336 |      7306 |  -12.4% |      780 |      660 |  -15.4% |
| url_to_file_path|       2736 |      2734 |   -0.1% |      285 |      280 |   -1.8% |

Issue #1135, opened 2026-06-24
Implement URL Standard change to pass through bogus xn-- labels if the whole domain is ASCII


hsivonen:
`idna` will need major changes to implement https://github.com/whatwg/url/pull/914 . I haven't thought through how badly it breaks the current design, but I very much want the accept/reject results to be consistent between `to_ascii` and `to_unicode` and have commented to that effect on the spec PR.

(I won't have time to work on this before August.)

PR #1145, opened 2026-08-01
perf: fast path for absolute http(s) URLs, and better join buffer sizing

bartlomieju:
Two changes to speed up parsing.

First, `Url::join` sized its parse buffer from the relative reference, but the
result is roughly the base plus that reference, so every join grew the buffer
through three reallocations. It now sizes for both.

Second, and this is where most of the win is, a fast path for plain absolute
http(s) URLs. Most URLs are already canonical, so there is nothing to normalise,
but the parser still walks them character by character, calls IDNA, and checks
every path byte against a percent-encode set. The fast path spots that case up
front and builds the `Url` directly. It only takes a lowercase ASCII domain with
a path and query that need no encoding, and bails on anything else (uppercase,
userinfo, ports, fragments, percent-encoding, dot segments, IP literals,
punycode, non-ASCII). On a 100k corpus of real URLs, 97% take it. `Url::join`
uses it too when the specifier is already absolute, since the base is ignored
there anyway. A differential test checks that everything the fast path accepts
comes out identical to the general parser, field for field. WPT is green.

Benchmarks on an Apple M5, criterion with paired baselines:

| | before | after | |
|---|---|---|---|
| short absolute URL | 97.5 ns | 34.6 ns | -64.5% |
| bulk parse, 100k URLs | 21.82 ms | 11.42 ms | -47.3% |
| bulk parse plus getters | 21.89 ms | 11.26 ms | -48.1% |
| `join`, absolute specifier | 328.1 ns | 113.0 ns | -65.6% |
| `join`, relative, base reparsed | 218 ns | 119.9 ns | -45.0% |
| `join`, relative, base pre-parsed | 109 ns | 88.3 ns | -19.0% |
| five module specifiers | 519 ns | 347.2 ns | -33.1% |
| IDN host (fast path bails) | 300 ns | 302.9 ns | +0.8% |

That last row is the cost of trying the fast path and bailing, which is within
noise here. Happy to share the benchmark harness if it would help review.

I tested all of these changes against `ada` and on my machine in most of 
the benchmarks `rust-url` is now faster. Both parsers still give identical result
in the ada test corpus, the only difference is performance.

I made sure this change doesn't clash with https://github.com/servo/rust-url/pull/1142
or potential redesign in https://github.com/servo/rust-url/issues/1135.

_Disclaimer: I used AI to research, implement and test this change._


Manishearth:
Hi! Current Servo AI policy does not allow AI contributions

https://book.servo.org/contributing/getting-started#ai-contributions

Thank you for working on this.

PR #1146, opened 2026-08-02
chore: add explicit package include lists to all crates.

dreyfus92:
as part of an [e18e](https://e18e.dev) effort to slim down published crates across the Rust ecosystem, this PR adds explicit `include` lists to all published crates so the `.crate` files no longer ship dev-only assets, and future additions to the crate directories (CI configs, fuzz corpora, bench data) can't silently land in the tarballs.

we left tests due to distro packagers run these crates' tests from the crates.io tarball Debian's debcargo configs for `url`, `idna`, and `data-url` all execute them in autopkgtest. Stripping tests wouldn't fail their builds (cargo strips the test targets from the published manifest), it would silently drop their test coverage.

`idna/tests/IdnaTestV2-Unicode16.txt` (760 KB) is never read by any test. CI copies it over `IdnaTestV2.txt` to test against the latest Unicode, and CI runs from the repo checkout, not the package. every other fixture is `include_str!`'d by a live test. idna's bench harness is also no longer shipped.

**Total: 381.0 KB → 276.0 KB (−28%)**  and since `idna` is a dependency of `url`, the −42% is paid by nearly every consumer of `url`.

| Crate | Before | After | Saved |
|---|---:|---:|---:|
| idna | 249.8 KB | 144.7 KB | −42.1% |
| data-url | 23.4 KB | 23.5 KB | ~0 |
| percent_encoding | 11.3 KB | 11.6 KB | ~0 |
| form_urlencoded | 9.1 KB | 9.2 KB | ~0 |

> [!NOTE]
> sizes are compressed `.crate` tarballs from `cargo package`. all crates pass `cargo package` verify builds, and the full workspace test suite (including the 12k-case UTS-46 suite that reads `IdnaTestV2.txt`) passes unchanged.


Manishearth:
What are the files that were included that no longer are?

dreyfus92:
just two files, both in idna:

 - `tests/IdnaTestV2-Unicode16.txt` nothing in the package reads it. it's only there for CI, which copies it over `IdnaTestV2.txt` to test against the latest Unicode, and CI runs from the repo checkout anyway.
- `benches/all.rs` the benchmarks file.

everything else packages the exact same file list as before. all the test fixtures still ship, so running the tests from the tarball still.

Manishearth:
Seems fine.

If this is a broader effort I would highly recommend including the list of removed files in the PR body: crates get asked a lot to reinclude files (for e.g. Debian), so it's not a straightforward call each time, there are tradeoffs. Listing them lets the crate author make a quick decision.

PR #1147, opened 2026-08-09
Fix Windows drive letter normalization in hostless file: URLs

### LINKED ISSUE #889: Incorrect parsing of Windows drive letter quirk
- [X] Note that this crate implements the [URL Standard](https://url.spec.whatwg.org/) not RFC 1738 or RFC 3986

**Description**
The following test case fails

```rust
let u = Url::parse("file:///C|/hello/world").unwrap();
assert_eq!(u.as_str(), "file:///C:/hello/world");
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f89f4eaa50a9ae2c747d1f2e26e02105)

According to the specification, it appears to me that this test case should pass. Currently the result given by the parser will be `"file:///C|/hello/world"` (without the `'|'` to `':'` conversion). Looking at the implementation `Parser::parse_file_host` ignores the `bool` returned by `Parser::file_host`, which indicates the branch specified in [file host state](https://url.spec.whatwg.org/#file-host-state).1. This results in the error described.

Please let me know if this is indeed an issue with the implementation or if I've missed something.

lenamonj:
Fixes #889.

`Url::parse("file:///C|/hello/world")` returns `file:///C|/hello/world`. Per [path state](https://url.spec.whatwg.org/#path-state) the `|` has to be rewritten: "If url's scheme is 'file', url's path is empty, and buffer is a Windows drive letter, then replace the second code point in buffer with U+003A (:)". The expected output is `file:///C:/hello/world`.

The repo already records the same defect from the other direction: `<file:///w|/m>` sits in `url/tests/expected_failures.txt`, and the WPT entry for it expects `file:///w:/m`.

## Cause

`parse_path` implements the spec's "url's path is empty" test as `segment_start == path_start + 1`, assuming `path[0]` begins one byte after `path_start`. That holds when the serialization carries a single slash there, but a `file:` URL can end up with a run of them: for `file:///w|/m`, `parse_path_start` finds `file://` already ending in `/` and returns without adding one, so `parse_path` appends its own.

Instrumenting the check on `main` shows it directly:

```
input file:///w|/m
  serialization="file:////"    path_start=7 segment_start=8 segment=""   -> test passes
  serialization="file:////w|/" path_start=7 segment_start=9 segment="w|" -> test fails
```

The only segment the test accepts is the empty one; the drive letter arrives a slash later and is never rewritten. The spellings that do work today (`file:/C|/a`, `file://host/C|/a`) carry exactly one slash at `path_start`, so the test holds there; the failing shapes are the ones that put a run of slashes at `path_start`, with the hostless `file:///` form the common case.

## Fix

Derive where `path[0]` actually starts instead of assuming `path_start + 1`.

## Verification

Removing `<file:///w|/m>` from `url/tests/expected_failures.txt` is the regression test: the harness grades both directions, so the line fails the suite without the fix and fails it again as an unexpected success if left in place with the fix.

On `main` with only that line removed, `cargo test -p url --test url_wpt` fails:

```
<file:///w|/m> ... ❌
  err:  expected href "file:///w:/m", but got "file:///w|/m"
```

With this change it passes, and the case from #889 returns `file:///C:/hello/world`. Also green on x86_64-pc-windows-msvc, rustc 1.90.0: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test`, `cargo test --features "url/serde,url/expose_internals"`, `cargo test --no-default-features --features=alloc`.

Behaviour before and after over a grid of 410 `file:` inputs and `join` pairs: the only differences are drive letters spelled `|` normalized to `:` when they form the first path segment. The neighbouring WPT cases `file:///w|m` and `file:///w||m`, which must keep the `|`, are unchanged, as are `file:/C|/a` and `file://host/C|/a`.


lenamonj:
Friendly ping. Small spec-mandated fix for #889, still clean against master; happy to adjust anything.

